PR #91, opened 2026-04-19
fix: auto-create AnimationPlayer when player_path missing (#86)

### LINKED ISSUE #86: animation_create_simple: auto-create AnimationPlayer when path doesn't resolve (parity with library auto-create)
### What happened

Calling `animation_create_simple` with `player_path` pointing at a `Node3D` (or a non-existent path) errors out:

```
INVALID_PARAMS: Node at /Root/Pivot is not an AnimationPlayer (got Node3D)
INVALID_PARAMS: Node not found: /Root/Pivot/AnimationPlayer
```

…which forced a manual `node_create type=AnimationPlayer` before the animation tool would accept the call.

### Why this is friction

The same tool *does* auto-create the `AnimationLibrary` if it's missing — `animation_create` returns `library_created: true`, and `animation_create_simple` does the same in [animation_handler.gd:681-683](plugin/addons/godot_ai/handlers/animation_handler.gd). It's internally inconsistent to auto-provision the library but not the player it lives on, especially when the player is a trivial `AnimationPlayer.new()` with no required config.

The docstring — "creates an AnimationPlayer + library if missing" (in the smoke template / tool docs callers are copying from) — reinforces the expectation that the tool is a one-shot animation authoring tool.

### Where the rejection lives

[animation_handler.gd:1252-1254](plugin/addons/godot_ai/handlers/animation_handler.gd):

```gdscript
if not node is AnimationPlayer:
    return McpErrorCodes.make(McpErrorCodes.INVALID_PARAMS,
        "Node at %s is not an AnimationPlayer (got %s)" % [player_path, node.get_class()])
```

### Proposed fix

If `player_path` doesn't resolve, create an `AnimationPlayer` at that path (its parent must exist, same rule as `node_create`), and include `animation_player_created: true` in the response — same shape as the existing `library_created: true` affordance. Bundle the creation into the same undo action so one Ctrl-Z rolls back the player, library, and animation together (see the pattern under "Auto-create missing dependencies in the same undo action" in `CLAUDE.md`).

If `player_path` resolves to a non-AnimationPlayer node, the current error is fine — that's a genuinely different node and the caller should pick another path.

### Repro

Freshly-installed plugin 1.1.0, Godot 4.6.2:

```
node_create type=Node3D name=Pivot parent_path=/Root

animation_create_simple player_path=/Root/Pivot name=bob   tweens='[{"target":".","property":"position","from":{"x":0,"y":0,"z":0},"to":{"x":0,"y":2,"z":0},"duration":1}]'
→ INVALID_PARAMS: Node at /Root/Pivot is not an AnimationPlayer (got Node3D)

animation_create_simple player_path=/Root/Pivot/AnimationPlayer name=bob tweens=[...]
→ INVALID_PARAMS: Node not found: /Root/Pivot/AnimationPlayer
```

Found during v1.1.0 install smoke — see friction_log.md entry `2026-04-19 — v1.1.0 install smoke`.

dsarno:
## Summary

- `animation_create` and `animation_create_simple` now auto-instantiate an `AnimationPlayer` (with an empty default `AnimationLibrary`) when `player_path` doesn't resolve, instead of erroring `"Node not found"`.
- Creation is bundled into the same `EditorUndoRedoManager` action that adds the animation — one Ctrl-Z rolls back the player **and** the clip.
- Response grows `animation_player_created: bool` so callers can tell if they got a fresh player vs. hit an existing one.
- The value coercer takes an `override_root_node` parameter so `{x, y, z}` dicts still land as `Vector3` (and Color dicts as `Color`, etc.) even when the just-created player isn't in the tree yet.
- The parent must already exist — we don't silently materialize arbitrary ancestor chains. If it doesn't, we return a helpful `INVALID_PARAMS` message with `ScenePath.format_parent_error`.
- Wrong-type collision (path resolves but to a non-AnimationPlayer node) still returns the original error — no clobbering.

Closes #86

## Test plan

- [x] `pytest -v` — 519/519 Python tests pass (including new `test_create_animation_player_created_flag_passes_through`)
- [x] `test_run` in live Godot editor — 673/673 GDScript tests pass, including 5 new tests in `test_animation.gd`:
  - `test_create_animation_auto_creates_player_when_missing`
  - `test_create_animation_auto_create_is_undoable` (undo removes player, redo restores)
  - `test_create_animation_errors_when_path_is_wrong_node_type`
  - `test_create_animation_errors_when_parent_missing`
  - `test_create_simple_auto_creates_player_and_coerces_vector_values` (the coercion override-root path)
- [x] Live-smoke in Godot: happy path returns `animation_player_created: true`, wrong-type still errors, missing-parent error message is useful


dsarno:
## Code Review — PR #91

### Overview

Adds auto-create of `AnimationPlayer` to `animation_create` and `animation_create_simple` when `player_path` doesn't resolve. Creation is bundled into the same undo action as the new clip. Adds `animation_player_created` to the response and an `override_root_node` parameter on the value coercer so vector/color dicts still land as typed Variants when the player isn't in the tree yet.

Diff: ~299 lines across 4 files (handler, tool docstrings, GDScript tests, Python integration test).

### Code quality

- **Matches the auto-create-dependency pattern** from CLAUDE.md (see `material_handler::assign_material`, `particle_handler::create_particle`, `animation_handler::create_animation`'s existing library auto-create). Single `create_action` rolls back all pieces.
- **Eager library attachment** in `_instantiate_player` (`new_player.add_animation_library("", lib)`) mirrors the existing `create_player` pattern — avoids a second `add_do_method(add_animation_library)` entry and lets the library ride the `add_do_reference(player)` across undo→redo.
- **`_resolve_player` grew two return fields** (`player_created`, `player_parent`) rather than a separate helper. Existing callers (delete_animation, add_property_track, etc.) just ignore the extra keys. Reasonable given the number of call sites.
- **`_commit_animation_add` grew two trailing optional params.** The 9-param signature is long but each param is load-bearing. Not worth extracting a builder class for a two-caller helper.
- **Name check in `_instantiate_player`**: `rfind("/")` + slicing handles both bare leaf names ("AP") and nested paths (`/Main/AP`). Empty-leaf-name check guards against trailing slashes.

### Correctness

- **Node leak fix applied**: `AnimationPlayer` is a `Node` (manually memory-managed). If `create_simple` bails out after `_resolve_player(true)` but before `_commit_animation_add`, the detached player would leak. Length validation is now hoisted before the resolve, and the remaining post-resolve error paths (`overwrite` collision, coercion errors × 2) call `player.queue_free()` when `created_player` is true. `create_animation` isn't affected — a freshly auto-created library can't have the requested animation, so the overwrite path is unreachable.
- **Coercion with detached player**: `_resolve_track_prop_context` can now accept an `override_root_node` to stand in for `player.root_node` resolution. Without this, the coercer would hit `is_inside_tree() == false`, fall through to pass_through, and silently store `{"x":…,"y":…}` dicts as keyframe values. Covered by `test_create_simple_auto_creates_player_and_coerces_vector_values`.
- **Index-at-undo-time rule** from CLAUDE.md: not applicable — we use `remove_child(player)` and `remove_animation_library("")`, both keyed by stable identifiers, not indices.

### Tests

- 5 GDScript tests: happy path, undo/redo, wrong-type error still fires, parent-missing error, coercion-with-override-root. All 673 GDScript tests pass.
- 1 Python integration test confirms `animation_player_created` flows through the tool layer. All 519 Python tests pass.
- The test that checks coercion via `track_get_key_value(0, 1) is Vector3` follows the CLAUDE.md guidance about asserting on the stored Variant, not on counts.

### Risks

- **Parent auto-creation is explicitly not supported** — `"/Main/MissingParent/AP"` errors with `ScenePath.format_parent_error`. That's the right call (avoid silent ancestor materialization), and the error message is useful.
- **Scene dirty without save**: auto-creating a player dirties the scene. Callers who `project_run` afterward will autosave the new player into the scene file (see CLAUDE.md "Live-smoke scene hygiene"). Not a bug, but worth noting for agents bulk-authoring animations.
- **Live-smoke risk from the current worktree**: this worktree is `animplayer-autocreate`. If the user merges and deletes it, the `test_project/main.tscn` changes from smoke runs would vanish with it. Main-repo tests already cover the behavior via `test_run`; no cleanup needed.

### Verdict

Ship it. The changes follow existing MCP conventions (undo bundling, eager deps, response flags for auto-creation), the Node leak is addressed, and coverage is thorough on both sides of the wire.

Issue #81, opened 2026-04-19
project_run silently persists MCP-driven scene mutations via try_autosave()


dsarno:
## Problem

`project_run` (any mode) internally calls `try_autosave()` → `_save_scene_with_preview()`, which persists any in-memory scene mutations to disk. For MCP-driven workflows this is surprising: a sequence like

```
script_attach /Main res://smoke.gd     # in-memory only
project_run mode=current               # persists the attachment to main.tscn on disk
```

silently commits the smoke-test attachment to `test_project/main.tscn`. During live smoke testing this leaks experimental state into scene files that other tests (and humans) depend on.

## Reproduction

1. Open `test_project/main.tscn` in Godot.
2. Via MCP: `script_create res://smoke.gd ...`, `script_attach /Main res://smoke.gd`.
3. Via MCP: `project_run mode=current`.
4. Stop the game. `git status` → `main.tscn` is modified with the attachment baked in.

Hit this while smoke-testing PR #78. Caught it before staging, but the trap is easy to miss, especially in a worktree where the pollution can leak into a PR. Workaround is now documented in CLAUDE.md via PR #79 (`git checkout -- test_project/<scene>` after smoking), but that is papering over the surprise.

## Why the autosave is there

`try_autosave()` is Godot's own behavior — the plugin can't opt out at the engine level without reimplementing the play path. But the MCP tool can give callers a choice.

## Possible fixes

Not prescriptive — flagging the shape of the fix rather than prescribing it:

- **`project_run(autosave=false, ...)` parameter**: plugin-side, snapshot the scene state before `play_*_scene()` and restore it after the play call returns. Behavior: transient in-memory mutations stay in memory, the file on disk stays untouched. Default stays `true` to preserve current behavior.
- **Dirty-scene warning**: return a non-error `warning` field in the `project_run` response when the scene was dirty at play time, listing which nodes/resources were mutated via MCP since the last explicit save. Surfaces the risk without blocking the action.
- **Transactional play mode**: a new `project_run_sandboxed` (or flag) that internally duplicates the scene to a temp `.tscn`, plays that, and discards — never touches the original file. Useful for smoke tests specifically.

(1) is probably the right trade-off — smallest API surface, preserves default behavior, gives agents a clean opt-out for smoke tests.

## Related

- CLAUDE.md "Live-smoke scene hygiene" subsection (PR #79) documents the workaround.
[...]

PR #95, opened 2026-04-19
fix: distinguish retryable EDITOR_NOT_READY states (#85)

### LINKED ISSUE #85: Distinguish retryable 'importing' state from other EDITOR_NOT_READY errors
## Problem

Right after a `filesystem_write_text` (or any write that triggers a Godot reimport), the session's readiness briefly flips to `importing` for ~1-2s. During that window, any write tool fails with:

```
EDITOR_NOT_READY: Editor is importing resources — try again shortly
```

The error code `EDITOR_NOT_READY` is also used for `playing` (game is running), `no_scene` (no scene open), and other non-retryable states. Callers can't distinguish "this will clear up in a moment, retry" from "this is a configuration problem, stop."

In practice agents end up string-matching the message to decide whether to back off, which is brittle.

## Suggested fix

Either:

1. **New error code** — add `EDITOR_IMPORTING` (and arguably `EDITOR_PLAYING` too) in `protocol/errors.py` + `utils/error_codes.gd`. Callers can now handle it distinctly without matching strings.

2. **`retryable` field on the error data** — cheaper change, same ergonomic outcome:
   ```python
   raise GodotCommandError(
       code=ErrorCode.EDITOR_NOT_READY,
       message="...",
       data={"retryable": True, "state": "importing"},
   )
   ```
   Clients check `error.data.retryable` before giving up.

Option 2 is lighter-touch but option 1 is more discoverable and the existing error-code pattern already anticipates it.

## Secondary: readiness flip isn't visible via `editor_state`

During the transient `importing` window, `editor_state` still reports the *previous* readiness (e.g. `no_scene`), so polling `editor_state` to decide when to retry doesn't work either. `session_list` does show the current readiness correctly, so the plugin knows — the two read paths just aren't synced.

## Context

Found while smoke-testing PR #78. See friction log entry on `main`.

dsarno:
## Summary

- `GodotCommandError` now accepts an optional `data: dict` and, when populated, embeds it as a `[k=v, ...]` suffix in `str(exc)` so MCP clients that only see the serialized exception can extract structured hints without string-matching the human-readable message.
- `require_writable` attaches `data={"retryable": bool, "state": str}` to the `EDITOR_NOT_READY` error it raises — `importing` is retryable, `playing` is not.
- Backward compatible: callers that don't pass `data=` get the existing `"CODE: message"` format unchanged.

Closes #85.

## Why this change

The same `EDITOR_NOT_READY` code was raised for both the ~1-2s `importing` window right after a `filesystem_write_text` (genuinely transient — retry with backoff works) and for the `playing` state (caller needs to stop the game). Agents had no programmatic way to tell them apart, so they resorted to string-matching the message — brittle.

This takes option 2 from the issue (structured `retryable` hint) and additionally folds it into `str(exc)` so it survives FastMCP's exception-to-error serialization, which is all the MCP client currently sees.

## Scope

Python-only. No plugin/GDScript changes. `ErrorDetail` wire protocol is unchanged — if the plugin ever needs to emit structured error data itself, that's a follow-up that plumbs `data` through `envelope.py` and the raising site in `client.py`.

## Test plan

- [x] `pytest` — all 519 tests pass
- [x] New assertions cover both the `.data` field and the serialized-form contract for importing + playing
- [x] New `test_godot_command_error_without_data_preserves_legacy_format` pins backward compat for the no-data path
- [x] `ruff check` + `ruff format --check` clean on touched files
- [ ] Live smoke against a running Godot editor — **not done locally** (see note below)

## Note on live smoke

This was run by the `godot-ai-issue-fixer` scheduled task in auto mode without a live Godot session pointed at this worktree's `test_project/`. The change is pure-Python in the readiness-gating path with full unit coverage of the happy path, both rejection branches, and the backward-compat path. Reviewers who want to exercise it live: drive `filesystem_write_text` on a large asset to hit the `importing` window, then immediately call any write tool and inspect the resulting error for the `[retryable=True, state=importing]` suffix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@
 
 <p align="center">
   <img src="docs/images/huddemo.gif" alt="Cyberpunk HUD demo" width="800"><br>
-  <em>UI demo built in ~2 hours with zero coding, zero image gen, all programmatically drawn by Godot AI</em>
+  <em>UI demo built in ~2 hours with zero coding, zero image gen, all programmatically drawn by Godot AI — <a href="https://github.com/hi-godot/cyberpunk-hud-demo">source</a></em>
 </p>
 
 ---
--- a/src/godot_ai/godot_client/client.py
+++ b/src/godot_ai/godot_client/client.py
@@ -15,10 +15,20 @@ logger = logging.getLogger(__name__)
 class GodotCommandError(Exception):
     """Raised when a Godot plugin command returns an error response."""
 
-    def __init__(self, code: str, message: str):
+    def __init__(
+        self,
+        code: str,
+        message: str,
+        data: dict[str, Any] | None = None,
+    ):
         self.code = code
         self.message = message
-        super().__init__(f"{code}: {message}")
+        self.data = data or {}
+        if self.data:
+            suffix = " [" + ", ".join(f"{k}={v}" for k, v in self.data.items()) + "]"
+            super().__init__(f"{code}: {message}{suffix}")
+        else:
+            super().__init__(f"{code}: {message}")
 
 
 class GodotClient:
--- a/src/godot_ai/handlers/_readiness.py
+++ b/src/godot_ai/handlers/_readiness.py
@@ -6,9 +6,12 @@ from godot_ai.godot_client.client import GodotCommandError
 from godot_ai.protocol.errors import ErrorCode
 from godot_ai.runtime.interface import Runtime
 
-_READINESS_MESSAGES = {
-    "importing": "Editor is importing resources — try again shortly",
-    "playing": "Editor is in play mode — stop the game first",
+# (message, retryable). Retryable means the condition clears on its own
+# (Godot finishes reimporting); non-retryable requires the caller to change
+# state (stop the game).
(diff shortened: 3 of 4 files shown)


dsarno:
## Live smoke verified

Ran against a live Godot 4.6.2 editor pointed at `test_project/` on this branch, with the dev server restarted from the branch's `.venv` and `src/` (so the new code was live):

```
editor_state           → readiness=ready
project_run mode=main  → readiness=playing
node_create Node3D     → Error calling tool 'node_create':
                         EDITOR_NOT_READY: Editor is in play mode —
                         stop the game first [retryable=False, state=playing]
```

Confirms the `[retryable=False, state=playing]` suffix survives FastMCP's exception-to-error serialization all the way to the MCP client — the whole point of embedding it in `str(exc)`.

I did not catch the `importing` window live (it's 1-2s and loses the race against sequential MCP calls), but the `retryable=True, state=importing` branch is covered symmetrically by the unit test and uses the same serialization path that just verified working above.

Cleanup: `git checkout -- test_project/snowman.tscn` to revert the autosave pollution from `project_run` (issue #81 in action), and `rm` on the two smoke scripts.

PR #102, opened 2026-04-20
fix: awaitable command dispatcher + readiness-aware project_stop (#29)

### LINKED ISSUE #29: Upgrade plugin dispatcher to support awaitable handlers + replace project_stop polling
## Motivation

PR #28 landed a nine-item cleanup from a retroactive review of PRs #17 and #21, but deliberately deferred one architectural change because it touches the hot code path every MCP command flows through. This issue captures that follow-up work so it can land as its own small, bisect-friendly PR.

The deferred change: **upgrade the plugin's command dispatcher to support `await`-able handlers, and use that to replace PR #28's server-side readiness polling in `project_stop` with an authoritative plugin response.**

## Why it matters

Today's dispatcher at [plugin/addons/godot_ai/dispatcher.gd:72](../blob/main/plugin/addons/godot_ai/dispatcher.gd#L72) calls handlers synchronously:

```gdscript
result = _handlers[command].call(params)
```

This has two consequences:

1. **No handler can `await` frame processing.** `EditorInterface.stop_playing_scene()` is async — `is_playing_scene()` only flips to `false` on the next frame. The plugin handler can't wait for that flip before returning the response, so the handler can't report authoritative readiness in the response.

2. **PR #28's `project_stop` workaround polls `session.readiness` from the server side** ([src/godot_ai/handlers/project.py:34-50](../blob/main/src/godot_ai/handlers/project.py#L34)) with a 20ms interval and 1s timeout:
   ```python
   while session.readiness == "playing" and loop.time() < deadline:
       await asyncio.sleep(0.02)
   ```
   This works but depends on `readiness_changed` events arriving from the plugin's `_process` polling. Under sustained load or CI slow-down, the event may take longer than 1s and the server falls through without authoritative truth.

### Phase 1 — Dispatcher coroutine support

Change [plugin/addons/godot_ai/dispatcher.gd:72](../blob/main/plugin/addons/godot_ai/dispatcher.gd#L72) from:
```gdscript
result = _handlers[command].call(params)
```
to:
```gdscript
result = await _handlers[command].call(params)
```

`await` on a non-coroutine return value resolves immediately, so existing synchronous handlers work unchanged. The change propagates one level up: `_dispatch` becomes a coroutine, so its caller (the command-loop frame in `Connection._process`) must also use `await` or accept the coroutine and handle it.

**Risk surface:** every command in the MCP server flows through this path. The refactor is small but the blast radius is "every tool." This is exactly why it was split out of PR #28 — an isolated change lets `git bisect` localize any regression to this single commit.

**Tasks:**
- [ ] Update `_dispatch` in `dispatcher.gd` to `await` handler return values
- [ ] Verify `Connection._process` (or whatever calls `_dispatch`) correctly handles the new coroutine return path
- [ ] Confirm no handler accidentally returns a `Signal` or `GDScriptFunctionState` that `await` would misinterpret
- [ ] Add a dispatcher-level test that a coroutine handler (one that does `await get_tree().process_frame`) returns its final value correctly
- [ ] Add a dispatcher-level test that synchronous handlers (the existing ~80 handlers) still work unchanged
- [ ] Run the full 388 Python + 355 GDScript test suite

### Phase 2 — Readiness Option C

Rewrite [plugin/addons/godot_ai/handlers/project_handler.gd](../blob/main/plugin/addons/godot_ai/handlers/project_handler.gd)'s `stop_project`:

```gdscript
func stop_project(_params: Dictionary) -> Dictionary:
    if not EditorInterface.is_playing_scene():
        return McpErrorCodes.make(McpErrorCodes.INVALID_PARAMS, "Project is not running")

    EditorInterface.stop_playing_scene()
    # Yield twice so Godot can tick the stop-play state change. After this,
    # is_playing_scene() reflects truth and get_readiness() is authoritative.
    await get_tree().process_frame
    await get_tree().process_frame

    return {
        "data": {
            "stopped": true,
            "undoable": false,
            "reason": "Play/stop is a runtime action",
            "readiness_after": ConnectionBase.get_readiness(),  # authoritative
        }
    }
```

Then simplify [src/godot_ai/handlers/project.py](../blob/main/src/godot_ai/handlers/project.py):

```python
async def project_stop(runtime: Runtime) -> dict:
    result = await runtime.send_command("stop_project")
    session = runtime.get_active_session()
    if session is not None:
        readiness_after = result.get("readiness_after")
        if readiness_after is not None:
            # Plugin already waited for the frame tick and gave us truth.
            session.readiness = readiness_after
        else:
            # Version skew fallback — old plugin without readiness_after.
            loop = asyncio.get_running_loop()
            deadline = loop.time() + 1.0
            while session.readiness == "playing" and loop.time() < deadline:
                await asyncio.sleep(0.02)
    return result
```

No more polling loop in the happy path. The plugin-side handler waits exactly as long as needed (typically 2 frames ≈ 34ms at 60fps), never more.

**Tasks:**
- [ ] Rewrite `stop_project` in `project_handler.gd` to `await get_tree().process_frame` twice and include `readiness_after` in the response
- [ ] Update `project.py::project_stop` to consume `readiness_after` directly, keeping the polling loop as a version-skew fallback
- [ ] Add a unit test verifying the fallback path fires when `readiness_after` is absent
- [ ] Add an integration test verifying the new path is exercised when the plugin includes `readiness_after`
- [ ] Live smoke: `project_run` → immediate `project_stop` → immediate `node_create` — confirm no `EDITOR_NOT_READY` race

### Phase 3 — Follow-on opportunities

Once the dispatcher supports coroutines, other async patterns become possible (list, don't implement in this PR):

- `scene_open` could `await` until `readiness_changed` from `importing` back to `ready` before returning
- `filesystem_reimport` could wait for the scan to complete
- Any future tool that needs to wait on an editor-side frame or event becomes straightforward

These are NOT in scope for this PR — keep the change focused on dispatcher + `project_stop`.

## Verification

1. `ruff check src/ tests/` — lint passes
2. `pytest -v` — all Python tests pass (today: 388)
3. `test_run` via MCP — all GDScript tests pass (today: 355)
4. Live smoke against the editor:
   - `project_run` then `project_stop` then `node_create` in quick succession — succeeds with no readiness race
   - Existing handlers (any synchronous tool like `node_create`, `theme_set_color`) still return results correctly — the `await` on synchronous return values must be a no-op
   - Plugin reload via `editor_reload_plugin` still works (it goes through the dispatcher too)

## Out of scope

- Any broader refactor of `Connection._process` beyond what's needed to support an awaitable `_dispatch`
- Converting existing synchronous handlers to coroutines "because we can"
- Any new MCP tools

Keep this PR small. The dispatcher change is plumbing; `project_stop` is the motivating single caller. Everything else is follow-on work that can land piecemeal once the plumbing is stable.

## Context

- PR #28 (this cleanup): https://github.com/hi-godot/godot-ai/pull/28
- Source review that surfaced this: retroactive review of PRs #17 (animation tools) and #21 (friction fixes)
- Plan file from PR #28 planning session: `.claude/plans/federated-frolicking-phoenix.md` (see "Deferred to follow-up PR" section)

dsarno:
## Summary

Closes #29.

- **Dispatcher**: `tick()` is now a fire-and-forget drain that awaits handler coroutines serially and pushes responses through a `response_sink` callback. Handlers may be synchronous *or* use `await` internally; both paths round-trip uniformly through `dispatch_direct`. `batch_execute` and the test runner propagate `await` through sub-dispatches so coroutine handlers compose inside batches and suites.
- **project_stop**: plugin awaits two `process_frame`s so `is_playing_scene()` has time to flip, then returns `readiness_after` in the payload. Python `project_stop` consumes it directly, replacing the prior asyncio polling loop. Retains a polling fallback for plugin/server version skew.

## Why this matters

Several handlers need to observe engine state that only settles on the next idle tick (play/stop transitions, scene open/close, reimports). Without `await`, the Python layer had to poll — slower, racy, and the polling loops accumulated across handlers. This change is the infra for #29 and unblocks follow-ups that rely on post-op readiness being authoritative.

## Test plan

- [x] GDScript dispatcher suite (10/10) — includes new coroutine-handler tests: `test_dispatch_direct_awaits_synchronous_handler`, `test_dispatch_direct_awaits_coroutine_handler`, `test_tick_delivers_coroutine_response_via_sink`, `test_tick_delivers_sync_response_via_sink`.
- [x] pytest full suite — 524 passed.
- [x] Live smoke test via MCP: `project_run` → `project_stop` returned `readiness_after:"ready"`; immediate `node_create` after stop had no `EDITOR_NOT_READY` race. `batch_execute` with multi-step commands worked through the new await path.
- [x] `/simplify` review pass — pruned narrative comments, validated no duplication with existing utilities.

## Not in this PR (flagged for follow-up)

- `_drain` budget_ms now only caps *burst-starts*, not frame-time-spent (once a handler awaits, the drain sees it through). Consider renaming or resetting the budget across `await` boundaries.
- `readiness_after` is a stringly-typed cross-boundary key; no central constant. Worth a follow-up to consolidate readiness literals.
- `plugin.gd` wires the sink to `_connection._send_json` — reaches into a private-named method. A public `send_response` on `Connection` would be cleaner.

### DIFF
--- a/plugin/addons/godot_ai/connection.gd
+++ b/plugin/addons/godot_ai/connection.gd
@@ -52,8 +52,7 @@ func _process(delta: float) -> void:
 			_check_state_changes()
 
 			if dispatcher:
-				for response in dispatcher.tick():
-					_send_json(response)
+				dispatcher.tick()
 
 		WebSocketPeer.STATE_CLOSED:
 			if _connected:
--- a/plugin/addons/godot_ai/dispatcher.gd
+++ b/plugin/addons/godot_ai/dispatcher.gd
@@ -2,12 +2,18 @@
 class_name McpDispatcher
 extends RefCounted
 
-## Routes incoming commands to handlers and manages the command queue
-## with a per-frame time budget.
+## Routes incoming commands to handlers and manages the command queue.
+##
+## Handlers may be synchronous or awaitable. Because a single awaitable
+## handler can span multiple frames, tick() kicks off a fire-and-forget
+## drain that delivers responses through `_response_sink`. Only one drain
+## runs at a time so ordering is preserved across `await` boundaries.
 
 var _command_queue: Array[Dictionary] = []
 var _handlers: Dictionary = {}  # command_name -> Callable
 var _log_buffer: McpLogBuffer
+var _response_sink: Callable = Callable()
+var _draining := false
 var mcp_logging := true
 
 
@@ -15,7 +21,12 @@ func _init(log_buffer: McpLogBuffer) -> void:
 	_log_buffer = log_buffer
 
 
-## Register a command handler. The callable receives (params: Dictionary) -> Dictionary.
+func set_response_sink(sink: Callable) -> void:
+	_response_sink = sink
+
+
+## Register a command handler. The callable receives (params: Dictionary) -> Dictionary
+## or an awaitable that eventually returns one.
 func register(command_name: String, handler: Callable) -> void:
 	_handlers[command_name] = handler
 
@@ -26,7 +37,7 @@ func register(command_name: String, handler: Callable) -> void:
 func dispatch_direct(command: String, params: Dictionary) -> Dictionary:
 	if not _handlers.has(command):
 		return McpErrorCodes.make(McpErrorCodes.UNKNOWN_COMMAND, "Unknown command: %s" % command)
-	return _call_handler(command, params)
+	return await _call_handler(command, params)
 
 
 ## Whether a command is registered.
@@ -66,24 +77,26 @@ func enqueue(cmd: Dictionary) -> void:
 const DEFERRED_RESPONSE := {"_deferred": true}
 
 
-## Process queued commands within a frame budget (milliseconds).
-## Returns an array of response dictionaries to send back.
-func tick(budget_ms: float = 4.0) -> Array[Dictionary]:
-	var responses: Array[Dictionary] = []
+## Start draining the queue. No-op if a drain is already in flight.
+func tick(budget_ms: float = 4.0) -> void:
+	if _draining or _command_queue.is_empty():
+		return
+	_drain(budget_ms)
+
+
+## Budget only bounds how many new commands we *start* in one window;
+## once a handler awaits we always see it through to completion.
+func _drain(budget_ms: float) -> void:
+	_draining = true
 	var start := Time.get_ticks_msec()
-	var idx := 0
-
-	while idx < _command_queue.size() and (Time.get_ticks_msec() - start) < budget_ms:
-		var cmd: Dictionary = _command_queue[idx]
-		var response := _dispatch(cmd)
-		if not response.get("_deferred", false):
-			responses.append(response)
-		idx += 1
-
-	if idx > 0:
-		_command_queue = _command_queue.slice(idx)
-
-	return responses
+	while not _command_queue.is_empty() and (Time.get_ticks_msec() - start) < budget_ms:
+		var cmd: Dictionary = _command_queue.pop_front()
+		var response: Dictionary = await _dispatch(cmd)
+		if response.get("_deferred", false):
+			continue
+		if _response_sink.is_valid():
+			_response_sink.call(response)
+	_draining = false
 
 
 func _dispatch(cmd: Dictionary) -> Dictionary:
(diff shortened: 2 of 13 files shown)


dsarno:
Superseded by #106.

On closer read, the "Phase 1" dispatcher-await refactor this PR implements duplicates infrastructure that already shipped in #76: `DEFERRED_RESPONSE` + `Connection.send_deferred_response()` already let handlers decouple their reply from their return. CLAUDE.md documents it under "Deferred responses".

#106 reuses that existing pattern for `project_stop` — 3 files, 185 lines, one handler — instead of adding a second parallel mechanism (13 files, 652 lines, touching every batch/test call site).

Phase 1 can be revisited later behind a concrete second caller, but it isn't needed as plumbing for Phase 2.

PR #168, opened 2026-04-22
feat(dock): client-config drift detector for stale URLs

### LINKED ISSUE #166: Client-config drift detector: surface stale client URLs after port change
## Background

PR [#164](https://github.com/hi-godot/godot-ai/pull/164) makes HTTP/WS ports configurable and gives users an Apply + Reload escape hatch when the default port is busy. After that Apply, every existing MCP client config (\`~/.claude.json\`, \`~/.codex/config.toml\`, Cursor's settings, …) still points at the old URL until the user manually re-runs \"Configure Clients\".

The picker used to carry a line of fine print saying so. That was dropped because forecasts-masquerading-as-detection is a smell — it fired whether or not any configured client actually existed, and it didn't refresh itself when the user re-ran Configure Clients successfully.

## Proposal

Event-driven drift detector (no per-frame cost). Triggers:

1. Plugin \`_enter_tree\` (once per session)
2. After \`_on_apply_new_port\` completes (the one action that causes drift)
3. When the Configure Clients window opens (already happens)
4. On editor focus-in (\`NOTIFICATION_APPLICATION_FOCUS_IN\`) — catches external edits

Each trigger fires **one** sweep of \`McpClientConfigurator.check_status()\` per registered client (filesystem read per client, ~18 clients, negligible). The dock caches the result until the next trigger.

## Implementation sketch

- \`McpClient.Status\` currently has \`CONFIGURED\` / \`NOT_CONFIGURED\` / \`ERROR\`. Add \`CONFIGURED_MISMATCH\`.
- The strategies (\`_json_strategy.gd\`, \`_toml_strategy.gd\`, \`_cli_strategy.gd\`) already read the stored entry and compare the URL against the current \`http_url()\`. Today they return \`NOT_CONFIGURED\` when the URL doesn't match — split that into \`NOT_CONFIGURED\` (no entry) vs \`CONFIGURED_MISMATCH\` (entry exists, wrong URL).
- The dock renders a banner near the existing Clients section (NOT wedged into the spawn-failure panel) listing mismatched clients, with a \"Reconfigure mismatched\" action. Banner hides when no drift.
- \`_apply_row_status\` already handles per-row dot colors; add an amber dot + \"Reconfigure\" button for the new status.

## Out of scope

- Per-frame polling. Triggers are sufficient.
- Automatic reconfigure. Show the drift, let the user press one button.
- Detecting drift the other direction (client config newer than plugin default). Unlikely in practice and hard to distinguish from legitimate manual edits.

## Reference

See the \`TODO:\` comment at the end of \`_build_port_picker_section\` in \`plugin/addons/godot_ai/mcp_dock.gd\` and the discussion starting [here](https://github.com/hi-godot/godot-ai/pull/164).

dsarno:
## Summary

- Adds a `CONFIGURED_MISMATCH` client status distinguishing "entry exists but URL is stale" from "no entry at all." After the user changes `http_port` via the picker, previously-configured clients silently kept pointing at the old URL and still looked green in the dock.
- New amber banner above the Clients row cites the drifted count + current port, with a single **Reconfigure** button that rewrites every mismatched entry.
- JSON / TOML / CLI strategies updated to detect mismatch; handler emits the new `"configured_mismatch"` wire string.
- Dock refactor: `_client_statuses` dict is the single source of truth (no more color-as-state), FOCUS_IN sweeps are debounced (2s cooldown) so rapid alt-tab does not re-shell-out CLI clients, and banner refresh short-circuits when the count is unchanged.
- Event-driven refresh only — no per-frame polling.

Closes #166.

## Test plan

- [x] `pytest` — 547 passed
- [x] `ruff check` — clean
- [x] GDScript `test_run` — 774 passed / 14 skipped / 0 failed across 33 suites
- [x] New tests: 6 strategy-level (`test_clients.gd`: JSON drift vs absent, TOML enabled=false vs URL mismatch, custom `verify_entry`, handler wire-string emission) + 3 dock-level (`test_dock.gd`: banner hidden when no drift, shows count+port, singular vs plural)
- [x] Live smoke via MCP `client_status` — all 18 clients report correct statuses
- [ ] Manual in-editor smoke of the amber banner after an Apply+Reload port change (reviewer)

### DIFF
--- a/plugin/addons/godot_ai/clients/_base.gd
+++ b/plugin/addons/godot_ai/clients/_base.gd
@@ -6,7 +6,11 @@ extends RefCounted
 ## Subclasses set fields in _init(); they should not contain control flow.
 ## Strategies (json/toml/cli) consume these fields.
 
-enum Status { NOT_CONFIGURED, CONFIGURED, ERROR }
+## CONFIGURED_MISMATCH distinguishes "entry exists but URL is stale" from
+## "no entry at all" (NOT_CONFIGURED). The dock uses this to offer a
+## "Reconfigure mismatched" escape hatch after a port change without making
+## the user remember which clients they configured. See issue #166.
+enum Status { NOT_CONFIGURED, CONFIGURED, CONFIGURED_MISMATCH, ERROR }
 
 var id: String = ""                              ## stable key, e.g. "cursor"
 var display_name: String = ""                    ## "Cursor"
--- a/plugin/addons/godot_ai/clients/_json_strategy.gd
+++ b/plugin/addons/godot_ai/clients/_json_strategy.gd
@@ -31,9 +31,12 @@ static func check_status(client: McpClient, server_name: String, server_url: Str
 	var entry = holder[server_name]
 	if not (entry is Dictionary):
 		return McpClient.Status.NOT_CONFIGURED
+	# Entry exists — if URL doesn't match our current server, it's drift, not absence.
+	# Lets the dock surface a "Reconfigure" banner after a port change instead of
+	# letting a stale URL silently disguise itself as "not configured".
 	if client.verify_entry.is_valid():
-		return McpClient.Status.CONFIGURED if client.verify_entry.call(entry, server_url) else McpClient.Status.NOT_CONFIGURED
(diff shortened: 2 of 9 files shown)


dsarno:
Superseded by #167, which shipped the same drift detector and closed #166.

PR #200, opened 2026-04-25
fix(node): reject uncoercible values to Vector/Color slots instead of silent zero (#191)

### LINKED ISSUE #191: node_set_property silently coerces Vector3/Vector2 values to zero
## Summary

`node_set_property` reports success and returns `undoable: true` when setting `position`, `scale`, or any Vector2/Vector3/Color property, but the value is silently coerced to the zero-vector regardless of input. The mutation actually lands on the node — `position` becomes `(0, 0, 0)` — so this is data-corrupting, not a no-op.

## Reproduction

```
node_set_property(path="/Main/SomeNode", property="position", value={"x": -8, "y": 0, "z": 0})
```

Response:

```json
{
  "old_value": {"x": 1.0, "y": 0.0, "z": 0.0},
  "path": "/Main/SomeNode",
  "property": "position",
  "undoable": true,
  "value": {"x": 0.0, "y": 0.0, "z": 0.0}
}
```

`old_value` is correct, so the call reaches the editor and the property exists. The applied `value` is zero. A follow-up `node_get_properties` confirms the node's position on disk is `(0, 0, 0)`.

Reported by a user who tried multiple input shapes — dict (`{"x":5,"y":5,"z":5}`), list (`[5,5,5]`), JSON string (`"{\"x\":5,...}"`), GDScript-style (`Vector3(5,5,5)`) — all produced the same zero-vector result. The fact that *every* shape collapses to zero (not just one or two) points at the value arriving on the GDScript side as the wrong Variant type entirely.

## Diagnosis

The Python tool signature ([`src/godot_ai/tools/node.py:175-216`](https://github.com/hi-godot/godot-ai/blob/main/src/godot_ai/tools/node.py#L175)) is correct: `value: str | int | float | bool | dict | list | None`. But on the GDScript side, [`node_handler.gd::_coerce_value`](https://github.com/hi-godot/godot-ai/blob/main/plugin/addons/godot_ai/handlers/node_handler.gd#L570) only handles the dict-with-x/y/z case for `TYPE_VECTOR3`. If the value arrives as a String (e.g. JSON-RPC serialization round-trip stringifying the dict), `_coerce_value` returns the input unchanged. `_check_dict_coerce_failed` then doesn't fire (a string isn't a wrong-shape dict), and `UndoRedo.add_do_property(node, "position", "<string>")` causes Godot to silently default-construct `Vector3.ZERO` rather than raise.

## Suggested fixes

1. **Loud failure**: when `_coerce_value` can't coerce to the target type, return an `INVALID_PARAMS` error with the actual received type — never let the mutation proceed with an unconvertible value. Worth verifying for primitive scalars too (int property + string input) — same silent-zero failure mode is plausible.
2. **Tolerate JSON-string compound values**: if the value arrives as a String and the target is Vector2/Vector3/Color/Array/Dictionary, try `JSON.parse_string` first, then re-coerce.
3. **Find and fix the stringification**: trace the Python→GDScript path for the `value` parameter and confirm whether the dict is being stringified somewhere in the JSON-RPC bridge. If so, fix at the source so dicts round-trip cleanly.
4. **Regression test**: round-trip `node_set_property("position", {"x":1,"y":2,"z":3})` → `node_get_properties` → assert equality. Cover Vector3, Vector2, Color, and Array/Dictionary properties.

## What does work in the same session

`material_apply_to_node`, `node_create`, `node_duplicate`, `node_delete`, `script_detach`, `resource_create` (assigning a `BoxMesh` to `mesh`) — so the bug is specific to `node_set_property` with compound-type properties, not a general session/connection issue.

dsarno:
## Summary

Closes #191. `node_set_property` was returning `status=ok, undoable=true` for Vector3/Vector2/Color properties given uncoercible inputs (Array, scalar int, JSON-string-encoded dict, wrong-shape dict), but `add_do_property` silently default-constructed `Vector3.ZERO` so the actual mutation was data-corrupting (e.g. `position=[1,2,3]` became `(0,0,0)` on disk). The existing `_check_dict_coerce_failed` only fired for Dictionary inputs, so non-dict shapes sailed through to the silent-zero path.

## Fix

Three pieces, all in `plugin/addons/godot_ai/handlers/node_handler.gd` plus a mirror at `resource_handler.gd::_apply_resource_properties`:

1. **Broaden `_check_coerced`** to be safe-to-call for any `target_type` (early-return null when the target isn't `Vector2`/`Vector3`/`Color`), then use it from `set_property` and `_apply_resource_properties` instead of the narrower dict-only check. Now Array/scalar/JSON-string inputs to a Vector slot return `INVALID_PARAMS` with a `"Property 'X': must coerce to Vector3, got Array"` message instead of silently writing zero.

2. **Teach `_coerce_value` to JSON-parse String inputs** that begin with `{` or `[` for compound targets (`Vector2`/`3`/`Color`/`Array`/`Dictionary`). Handles middleware/clients that stringify nested-dict tool args (suggestion #2 from the issue). Hex Color strings still bypass to `Color(value)` since they don't begin with `{`/`[`.

3. **Extract `VECTORLIKE_TYPES` and `JSON_PARSE_STRING_TARGETS` as module consts** so the type-gate arrays aren't allocated on every `set_property` call (hot in `batch_execute`). GDScript doesn't intern array literals.

## Tests

- **GDScript** (`test_project/tests/test_node.gd`): rejects `[1,2,3]` and `5` to `position`, accepts `'{"x":-8,"y":4,"z":1}'` JSON-string and verifies `node.position == Vector3(-8,4,1)` (stored Variant check, not just response counts), preserves `"#ff0000"` hex Color, and verifies garbage JSON like `"{not valid json}"` passes through to the loud-error path.
- **Python integration** (`tests/integration/test_mcp_tools.py`): verifies a Vector3 dict round-trips through the FastMCP pipeline as a real `dict` (not stringified), guarding against pydantic union resolution or middleware flattening it before it reaches the plugin.

## Test plan
- [x] `pytest -q` — 564 passed (was 563; +1 new integration test)
- [x] GDScript tests added in `test_project/tests/test_node.gd` (7 new tests covering the previously-silent failure modes)
- [ ] Live smoke against running Godot editor — **not run**: this sandbox has no Godot binary. Reviewer running locally should `test_run suite=node` to confirm the new tests pass and that the existing `test_set_property_*` tests still pass.

## Out of scope

`Transform3D`/`Quaternion`/`Basis`/`Plane` are also compound types not currently in `_coerce_value` — they'd suffer the same silent-zero bug if an agent tried to set them. The doc comment on `_check_coerced` already says "extend the match alongside `_coerce_value` if you add a new coerce target" — leaving as a follow-up since #191 specifically calls out Vector2/3/Color.

https://claude.ai/code/session_0124t36fmiJtwQJeiER23uRd

---
_Generated by [Claude Code](https://claude.ai/code/session_0124t36fmiJtwQJeiER23uRd)_

### DIFF
--- a/plugin/addons/godot_ai/handlers/node_handler.gd
+++ b/plugin/addons/godot_ai/handlers/node_handler.gd
@@ -237,8 +237,10 @@ func set_property(params: Dictionary) -> Dictionary:
 		instantiated_resource = true
 	else:
 		value = _coerce_value(value, target_type)
-		## Refuse wrong-shape dicts that _coerce_value passed through (#123).
-		var coerce_err := _check_dict_coerce_failed(value, target_type)
+		## Refuse uncoercible values that would silently zero-default at
+		## `add_do_property` (#123 — wrong-shape dicts; #191 — non-dict
+		## values like Array/String/int landing on a Vector slot).
+		var coerce_err := _check_coerced(value, target_type, "Property '%s'" % property)
 		if coerce_err != null:
 			return coerce_err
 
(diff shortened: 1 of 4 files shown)


dsarno:
Closing as superseded by #197, which already merged a fix for #191. Both PRs replace the narrow `_check_dict_coerce_failed` path with broader Vector/Color coercion guarding in `plugin/addons/godot_ai/handlers/node_handler.gd`. #197 went in at `0f3caab`; this branch's diff conflicts with the merged version.

PR #203, opened 2026-04-25
Collapse 118 MCP tools to 39 via &lt;domain&gt;_manage rollups

dsarno:
## Summary

The MCP server now exposes **~18 high-traffic verbs as named tools** plus **one rolled-up `<domain>_manage` tool per domain** that takes `op="<verb>"` + a `params` dict. Total tool count drops **118 → 39** — comfortably under hard tool-count caps (Antigravity is the immediate motivator) while every verb stays reachable.

Resources grow **7 → 16**: new `godot://node/{path}/properties|children|groups` and `godot://script/{path}` templates, plus list resources for `materials`, `input_map`, `performance`, `editor/state`, and `test/results`. Tool forms remain as the per-call `session_id`-pinning fallback for clients that don't surface resources.

## Why

Some MCP clients enforce a hard tool-count cap and ignore Anthropic's `defer_loading` hint (Antigravity rejects connections with >100 tools today). The previous shape — ~118 tools, mostly tagged deferred — broke for those clients. Following Anthropic's guidance to "group tools around intent, not endpoints," we keep the high-traffic verbs as named tools (so tool-search keeps per-verb autocomplete) and collapse each domain's long tail into a single rolled-up tool. The shape is independently confirmed by Unity MCP, which converged on the same `manage_<domain>` pattern with ~22 rollups + 24 resources.

## What changed

**New files**
- `src/godot_ai/tools/_meta_tool.py` — `register_manage_tool` helper. Builds a dynamic `Literal[...]` for `op` so schema-aware clients still see every valid verb. Unknown ops surface a structured `INVALID_PARAMS` error with fuzzy `data.suggestions` from `difflib.get_close_matches`. Handles JSON-string coercion for stringified array/object params (port of the old `JsonCoerced` validator).
- `src/godot_ai/resources/nodes.py` — `godot://node/{path}/properties|children|groups` templates.
- `src/godot_ai/resources/scripts.py` — `godot://script/{path}` template.
- `src/godot_ai/resources/library.py` — `godot://editor/state`, `godot://materials`, `godot://input_map`, `godot://performance`, `godot://test/results`.
- `tests/unit/test_meta_tool.py` — 17 tests covering schema construction, dispatch happy path, error paths (unknown op, non-dict params, handler TypeError), JSON-string coercion.

**Refactored**
[...]

PR #207, opened 2026-04-25
fix(server): accept stringified params on *_manage rollups (#206)

### LINKED ISSUE #206: <domain>_manage rollups reject stringified params from clients (Cline)
## Summary

PR #203 collapsed ~118 MCP tools into 39 named verbs + per-domain `<domain>_manage` rollups. Each rollup takes `op: Literal[...]` and `params: dict[str, Any] | None`. The `params` field is typed strict-dict via Pydantic.

Live smoke against `main` (e7cfe97) surfaced a regression: clients that auto-serialize nested-object arguments (Cline observed; likely others) send `params` as a JSON-encoded **string** rather than a dict. Pydantic strict-mode rejects every such call.

## Bug 1 — `*_manage.params` as JSON string

Reproducer (against `python -m godot_ai --transport streamable-http --port 8000` from `main`):

```python
from fastmcp import Client
import asyncio, json

async def main():
    async with Client("http://127.0.0.1:8000/mcp/") as c:
        await c.call_tool("resource_manage", {
            "op": "create",
            "params": json.dumps({"type": "BoxMesh", "properties": {"size": {"x": 5, "y": 5, "z": 5}}}),
        })

asyncio.run(main())
```

Observed:

```
ToolError: 1 validation error for call[resource_manage]
params
  Input should be a valid dictionary [type=dict_type, input_value='{"type": "BoxMesh", ...', input_type=str]
```

This affects every `<domain>_manage` rollup whenever a client stringifies the `params` object — `resource_manage`, `node_manage`, `script_manage`, `editor_manage`, `animation_manage`, `material_manage`, `audio_manage`, `particle_manage`, `camera_manage`, `signal_manage`, `input_map_manage`, `autoload_manage`, `filesystem_manage`, `theme_manage`, `ui_manage`, `client_manage`, `session_manage`, `test_manage`, `project_manage`, `scene_manage`.

## Bug 2 — `node_set_property.value` as string-of-class-dict

Same root cause but in a flat-arg tool. The `value` arg is documented as accepting `{"__class__": "BoxMesh", ...}` for fresh built-in resources (see `src/godot_ai/tools/node.py` around line 120). Cline sends that dict as a JSON string. The schema accepts string (it is part of the union), so the request reaches the handler — which then takes the `target_type == TYPE_OBJECT and value is String` branch in `plugin/addons/godot_ai/handlers/node_handler.gd:206-213` and treats it as a `res://` path.

Resulting error from a live editor session:

```
Resource not found: {"__class__": "BoxMesh", "properties": {"size": {"x": 5, "y": 5, "z": 5}}}
```

## Suggested fix

1. **Bug 1:** Mirror PR #199 (`StripClientWrapperKwargs`). Add a sibling middleware (`ParseStringifiedParams`) that, for tool calls whose name ends in `_manage`, checks whether `arguments["params"]` is a `str` and tries `json.loads` — replacing it with the parsed dict on success, leaving it untouched on failure so a genuine string-typed param later still surfaces a normal validation error.
2. **Bug 2:** In `node_handler.gd`'s `set_property`, before treating a `String` value as a `res://` path when target is `TYPE_OBJECT`, try to JSON-parse it. If it decodes to a Dictionary with `__class__`, route to the existing `__class__` instantiation branch. Otherwise fall through to the resource-load path. Keeps the `__class__` shape close to the other coercion logic in the same handler.

## Acceptance

- Reproducer above succeeds against the fix.
- Existing 598-test suite stays green.
- New unit tests covering middleware behavior and the `__class__`-string coercion shape.
- New integration test calling a `*_manage` rollup with stringified `params` through the full FastMCP stack.

dsarno:
## Summary

- Add `ParseStringifiedParams` middleware that JSON-decodes a string `params` slot on any `*_manage` tool call before Pydantic validates. Mirrors the `StripClientWrapperKwargs` middleware shape from #199.
- Fix the parallel case in `node_set_property.value` inside the GDScript handler: when target is `TYPE_OBJECT` and value arrives as a JSON string of `{"__class__": ...}`, parse it back to a Dictionary so the existing `__class__` instantiation branch handles it, instead of the next branch treating the JSON blob as a `res://` path and emitting `Resource not found: {...}`.
- 8 new unit tests covering the middleware happy/edge paths.
- 2 new integration tests calling `scene_manage` with stringified params through the full FastMCP stack.

Closes #206.

## Why

PR #203 collapsed ~118 per-verb tools into 39 named verbs + per-domain `<domain>_manage` rollups. Each rollup's `params` field is typed strict-dict via Pydantic. Some MCP clients (Cline confirmed) auto-serialize nested-object arguments before sending — every such call was rejected with `Input should be a valid dictionary` before any handler ran.

A blanket `extra="ignore"` would silence the error but also silently drop typo'd real params. The narrow middleware approach keeps strict validation everywhere except this one slot we know clients stringify.

## Test plan

- [x] `pytest` — full suite, 601 passed (was 591 + 10 new).
- [x] Live smoke against fixed server: pre-fix `resource_manage` with stringified `params` died at Pydantic; post-fix the request reaches the handler and returns a meaningful "Must provide either path+property…" error from the handler itself.
- [ ] Manual smoke from a Cline session with the fixed server — out of scope for this PR but the integration test exercises the same wire path.

### DIFF
--- a/plugin/addons/godot_ai/handlers/node_handler.gd
+++ b/plugin/addons/godot_ai/handlers/node_handler.gd
@@ -203,6 +203,16 @@ func set_property(params: Dictionary) -> Dictionary:
 
 	var instantiated_resource := false
 
+	# Some MCP clients (Cline) stringify the documented {"__class__": "BoxMesh", ...}
+	# value before sending. Promote that string back to a Dictionary here so the
+	# `__class__` branch below handles it, instead of the next branch treating
+	# the JSON blob as a res:// path and emitting "Resource not found: {...}".
+	# See #206.
+	if target_type == TYPE_OBJECT and value is String and value.begins_with("{"):
+		var json := JSON.new()
+		if json.parse(value) == OK and json.data is Dictionary and (json.data as Dictionary).has("__class__"):
+			value = json.data
+
 	if target_type == TYPE_OBJECT and value is String:
 		if value == "":
 			value = null
--- a/src/godot_ai/middleware/__init__.py
+++ b/src/godot_ai/middleware/__init__.py
@@ -6,5 +6,10 @@ from godot_ai.middleware.client_wrapper_kwargs import (
     CLIENT_WRAPPER_KWARGS,
     StripClientWrapperKwargs,
 )
+from godot_ai.middleware.parse_stringified_params import ParseStringifiedParams
 
-__all__ = ["CLIENT_WRAPPER_KWARGS", "StripClientWrapperKwargs"]
+__all__ = [
+    "CLIENT_WRAPPER_KWARGS",
+    "ParseStringifiedParams",
+    "StripClientWrapperKwargs",
+]
--- a/src/godot_ai/middleware/parse_stringified_params.py
+++ b/src/godot_ai/middleware/parse_stringified_params.py
@@ -1,0 +1,51 @@
+"""Decode stringified ``params`` on ``<domain>_manage`` calls before validation.
+
+PR #203 collapsed ~118 per-verb tools into 39 named verbs plus per-domain
+``<domain>_manage`` rollups. Each rollup takes ``op: Literal[...]`` and
+``params: dict[str, Any] | None``, validated strictly by Pydantic.
+
+Some MCP clients (Cline observed in the wild) auto-serialize nested-object
+arguments before sending. Their ``tools/call`` arrives with
+``arguments["params"]`` as a JSON string, and Pydantic strict-mode rejects
+the call before any handler runs. This middleware intercepts ``*_manage``
+calls and JSON-decodes a string ``params`` slot in place, so downstream
+validation sees the dict the client meant to send. See #206.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+
+from fastmcp.server.middleware import CallNext, Middleware, MiddlewareContext
+from fastmcp.tools.base import ToolResult
+from mcp.types import CallToolRequestParams
+
+logger = logging.getLogger(__name__)
+
+
+class ParseStringifiedParams(Middleware):
+    async def on_call_tool(
+        self,
+        context: MiddlewareContext[CallToolRequestParams],
+        call_next: CallNext[CallToolRequestParams, ToolResult],
+    ) -> ToolResult:
+        arguments = context.message.arguments
+        if arguments and context.message.name.endswith("_manage"):
+            raw = arguments.get("params")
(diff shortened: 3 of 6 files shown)


dsarno:
## Live smoke evidence

Smoked end-to-end against a fresh godot-ai server (this branch) + Godot 4.6.2 editor on `test_project`.

### Bug 1 — `*_manage.params` as JSON string

```python
await c.call_tool("resource_manage", {
    "op": "create",
    "params": json.dumps({"type": "BoxMesh", "properties": {"size": {"x": 5, ...}}}),
})
```

- **Pre-fix:** `ToolError: 1 validation error for call[resource_manage] / params / Input should be a valid dictionary [type=dict_type, input_value='{"type": "BoxMesh", ...']` — Pydantic rejects before any handler runs.
- **Post-fix:** request reaches the handler. (Returns a normal handler-level error about missing `path+property` / `resource_path` because the original prompt's payload didn't include either — that's `resource_create`'s own arg validation, not the bug.)

### Bug 2 — `node_set_property.value` as string-of-class-dict

Targeted `/root/Main/World/Ground` (a `MeshInstance3D` in `test_project/main.tscn` with an empty `mesh` slot).

```python
await c.call_tool("node_set_property", {
    "path": "/root/Main/World/Ground",
    "property": "mesh",
    "value": json.dumps({"__class__": "BoxMesh", "size": {"x": 5, "y": 5, "z": 5}}),
})
```

- **Pre-fix path:** GDScript `set_property` took the `value is String` branch, called `ResourceLoader.load(value)` on the JSON blob, and would have emitted `Resource not found: {"__class__": "BoxMesh", ...}`.
- **Post-fix:**
  ```
  >>> node_set_property(...)
  <<< { old_value: None,
        value: '():<BoxMesh#-9223369721967996262>',
        undoable: True }
  ```
  Follow-up `node_get_properties` confirms:
  ```
  { name: 'mesh', type: 'Object', value: '():<BoxMesh#-9223369721967996262>' }
  ```
  Property count went 32 → 33 (Godot exposes mesh-related fields once a mesh is assigned). `old_value: None` confirms nothing pre-existing was clobbered.

The `size` application itself isn't directly observable in the MCP response (it lives inside the resource), but my fix only changes the *entry path* into the existing `__class__` branch — once `value` becomes a `Dictionary`, the unchanged `ResourceHandler._apply_resource_properties` handles property application.

### Suite

`pytest` — 601 passed (8 new unit + 2 new integration; original 591 untouched).

PR #215, opened 2026-04-26
fix(handlers): hide editor-internal entries from list ops

### LINKED ISSUE #213: List ops surface editor-internal entries (signal_manage list, input_map_manage list)

```
```

```
```

```
```

## Suggested fix

## Environment

- Worktree HEAD `3a8ad17`

dsarno:
Closes #213.

## Summary

Two `list` ops surfaced editor-internal state as if it were user content:

- `signal_manage(op="list", params={"path": "/Main"})` returned 5 connections to `SceneTreeEditor` / dock UI observers (`visibility_changed`, `child_order_changed`, etc.) — the editor's own listeners on the scene root.
- `input_map_manage(op="list", params={})` (default `include_builtin=False`) returned 16 `spatial_editor/*` entries in a project with **zero** user-defined input actions — the editor runtime's freelook / orbit / zoom modifier bindings.

Agents reading either response saw the editor's plumbing attributed to the project.

## Changes

- **`signal_handler.gd::list_signals`** — drop connections whose target isn't part of the edited scene tree (and isn't a registered autoload directly under `/root`). Editor listeners are now tallied in a separate, additive `editor_connection_count` field so callers can still detect hidden observers without seeing each one in `connections[]`. Autoload-name lookup extracted to a small `_is_declared_autoload` helper, also reused by the existing `_resolve_node_or_autoload`.
- **`input_handler.gd::list_actions`** — when `include_builtin=False`, also skip any action that has no `input/<name>` entry in `ProjectSettings`. Editor-runtime actions added via `InputMap.add_action` (without ProjectSettings persistence) are now hidden by default; user-authored actions added via our `add_action` op are persisted to `ProjectSettings` and pass the filter.
- **Tool descriptions** for `signal_manage` and `input_map_manage` updated to document the new filtering shape.
- **GDScript tests** cover both filters. Both new tests follow a capture-then-tear-down-then-assert pattern so a mid-test failure can never leak the simulated fixture into a subsequent run; fixture nodes/actions use the `_McpTest…` prefix that the test runner's safety-net cleanup already recognises.

## Backwards compatibility

- `editor_connection_count` is a purely additive field — older Python integration tests don't read it and stay green (608/608 passing locally).
- `connections[]` shrinks to user wiring only; no existing caller iterates it semantically.
- The resource form `godot://input_map` calls `input_map_list` with the default `include_builtin=False`, so it gets the new filtering for free.

## Test plan

- [x] `pytest -q` — 608 passed
- [ ] **GDScript tests** — exercised in CI's Godot test matrix (`script/ci-godot-tests`). The new tests are `test_signal.gd::test_list_signals_filters_editor_internal_connections` and `test_input.gd::test_list_actions_filters_editor_runtime_actions` / `test_list_actions_includes_user_authored`.
- [ ] **Live editor smoke** — could not run in this sandbox (no Godot binary available). The fix follows the issue's repro exactly: open a fresh `Node3D` scene with zero user input actions, then `signal_manage(op="list", params={"path": "/Main"})` should now return `connection_count: 0, editor_connection_count: 5`, and `input_map_manage(op="list", params={})` should return `count: 0`.

https://claude.ai/code/session_01XYciuLk4eoJsJnbspC74p2

---
_Generated by [Claude Code](https://claude.ai/code/session_01XYciuLk4eoJsJnbspC74p2)_

### DIFF
--- a/plugin/addons/godot_ai/handlers/input_handler.gd
+++ b/plugin/addons/godot_ai/handlers/input_handler.gd
@@ -12,7 +12,12 @@ func list_actions(params: Dictionary) -> Dictionary:
 	for action_name in InputMap.get_actions():
 		var name_str := str(action_name)
 		var is_builtin := name_str.begins_with("ui_")
-		if is_builtin and not include_builtin:
+		if not include_builtin and is_builtin:
+			continue
+		# Editor-runtime actions (e.g. spatial_editor/freelook_left) live in
+		# InputMap but never appear under input/ in project.godot. Treat
+		# "user-authored" as "registered in ProjectSettings".
+		if not include_builtin and not ProjectSettings.has_setting("input/" + name_str):
 			continue
 		var events: Array[Dictionary] = []
 		for event in InputMap.action_get_events(action_name):
--- a/plugin/addons/godot_ai/handlers/input_handler.gd
+++ b/plugin/addons/godot_ai/handlers/input_handler.gd
@@ -15,8 +15,9 @@ func list_actions(params: Dictionary) -> Dictionary:
 		if not include_builtin and is_builtin:
 			continue
 		# Editor-runtime actions (e.g. spatial_editor/freelook_left) live in
-		# InputMap but never appear under input/ in project.godot. Treat
-		# "user-authored" as "registered in ProjectSettings".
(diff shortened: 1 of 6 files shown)


dsarno:

---

Issue #214, opened 2026-04-26
node_get_properties returns AABB and packed arrays as Godot debug strings instead of structured values


dsarno:
Caught during a smoke of the post-#203 rollup surface (see latest entry in `friction_log.md`).

## Repro

```
node_get_properties(path="/Main/Box")    # MeshInstance3D
```

Includes:

```json
{ "name": "custom_aabb", "type": "AABB", "value": "[P: (0.0, 0.0, 0.0), S: (0.0, 0.0, 0.0)]" }
```

The `value` is a literal print of Godot's debug repr of the AABB, not structured data. Same shape on `Label.tab_stops`:

```json
{ "name": "tab_stops", "type": "PackedFloat32Array", "value": "[]" }
```

— a string `"[]"`, not an empty array.

## Why this is friction

Agents read `node_get_properties` to reason about scene state and round-trip values back through `node_set_property`. `Vector2` / `Vector3` / `Color` already coerce to `{x,y,z}` / `{r,g,b,a}` dicts — both directions. AABB and packed arrays haven't gotten the same treatment, so:

- An agent can't programmatically inspect `custom_aabb` (e.g. "is the AABB zero-sized?") without parsing a debug string.
- An agent can't round-trip `tab_stops` — the string `"[]"` doesn't match `node_set_property`'s expected list shape.

## Suggested fix

Coerce in the variant-to-JSON helper that `node_handler.gd::get_node_properties` (and friends) leans on:

- `AABB` → `{"position": {"x":..., "y":..., "z":...}, "size": {"x":..., "y":..., "z":...}}`
- `Rect2` / `Rect2i` → `{"position": {"x":..., "y":...}, "size": {"x":..., "y":...}}`
- `Transform2D` / `Transform3D` / `Basis` / `Projection` → either matrix arrays or the full set of basis components
- `PackedFloat32Array` / `PackedFloat64Array` / `PackedInt32Array` / `PackedInt64Array` / `PackedByteArray` / `PackedStringArray` / `PackedVector2Array` / `PackedVector3Array` / `PackedColorArray` → JSON arrays of their element type

Coercion symmetry: every type that `node_set_property` *accepts* as a structured value should be *returned* in that same structured shape from `node_get_properties`. Right now the table is asymmetric — agents can write Vector3/Color/etc. as dicts but read AABB/packed arrays only as opaque strings.

[...]

Issue #212, opened 2026-04-26
ParseStringifiedParams (#206) doesn't handle a fully-stringified params arg


dsarno:
Follow-up to #206. Caught during a smoke of the post-#203 rollup surface (see latest entry in `friction_log.md`).

## Repro

```
node_manage(op="get_children", params='{"path":"/Main"}')
```

The whole `params` argument is a JSON string. The middleware shipped in #206 only walks values *inside* `params` and JSON-decodes those — it doesn't decode `params` itself when it arrives as a string. So Pydantic rejects:

```
1 validation error for call[node_manage]
params
  Input should be a valid dictionary [type=dict_type, input_value='{"path":"/Main"}', input_type=str]
```

## Why this matters

Some MCP clients (notably some Cline-family ones, per #199) auto-serialize the whole top-level argument when it contains nested structure. #206 fixed the case where individual values inside `params` arrive as strings; this is the symmetric outer case. Without coverage here, those clients still see a confusing `dict_type` error and have no path forward.

## Suggested fix

In `src/godot_ai/middleware/parse_stringified_params.py`, before the existing inner-value walk, check if `params` itself is a `str` that parses as a JSON object — if yes, replace it with the decoded dict and continue with the inner walk. Net change is ~5 lines.

Alternatively, if the project has decided not to support the fully-stringified-params client variant, document the limitation in `_meta_tool.py` so the next person doesn't assume #206 covers both shapes.

## Environment

- Worktree HEAD `3a8ad17` (already includes #206/#207)
- Reproducible with the Claude Code MCP client by manually wrapping `params` in `JSON.stringify(...)` before sending.

dsarno:
Closing as already addressed. The `ParseStringifiedParams` middleware shipped in #206/#207 does decode `arguments["params"]` when it arrives as a fully-stringified JSON string — the issue's premise ("the middleware … only walks values *inside* `params`") doesn't match what the code does today.

Reference test paths:

- `tests/unit/test_middleware_parse_stringified_params.py::TestParseStringifiedParams::test_decodes_string_params_on_manage_call` — exact `arguments={"op": "create", "params": '{"type": "BoxMesh", ...}'}` shape this issue describes.
- `tests/integration/test_mcp_tools.py::TestManageRollupAcceptsStringifiedParams::test_scene_manage_with_stringified_params_carrying_values` — same shape end-to-end through the FastMCP stack.

[...]

Issue #211, opened 2026-04-26
<domain>_manage op typos return plain Pydantic literal_error with no 'Did you mean' suggestion


dsarno:
Caught during a smoke of the post-#203 rollup surface (see latest entry in `friction_log.md`).

## Repro

```
node_manage(op="get_childen", params={"path":"/Main"})
```

Returns:

```
1 validation error for call[node_manage]
op
  Input should be 'get_children', 'get_groups', 'delete', 'duplicate', 'rename', 'move', 'reparent', 'add_to_group' or 'remove_from_group' [type=literal_error, input_value='get_childen', input_type=str]
```

No "Did you mean: `get_children`" — even though `difflib.get_close_matches('get_childen', valid_ops)` would obviously surface it.

## Why

CLAUDE.md already calls this out:

> The meta-tool's own `difflib`-based fallback in `dispatch_manage_op` only fires when the call somehow bypasses Pydantic (e.g. a future internal direct-dispatch caller).

Because `op` is typed `Literal[...]`, validation happens at the FastMCP/Pydantic schema boundary, **before** any handler code runs. The in-house difflib suggester (the same machinery #208 added for plugin-command typos) never gets a chance.

But this is the most common rollup-misuse error in practice — agents typo op names constantly. Recovering the suggestion would close the gap.

## Two ways to fix

**Option A (cleaner):** FastMCP middleware that intercepts `pydantic_core.ValidationError` for any rollup tool, detects when the failing field is `op`, and rewrites the message with a `difflib.get_close_matches` hint plus the valid alternatives. Centralised, applies to every rollup automatically.

**Option B (simpler):** in `tools/_meta_tool.py::register_manage_tool`, change the `op` parameter from `Literal[...]` to plain `str`, and validate inside `manage()` with the same `dispatch_manage_op` difflib path that already exists. Loses some schema introspection benefits but gives every client the same suggester behavior.

A would be my preference — keeps schema introspection working for tool-search-aware clients and only changes the user-facing error message, not the contract.

## Environment

- Worktree HEAD `3a8ad17`
- Pre-existing limitation per CLAUDE.md, but worth a tracked issue — typos on op names are *the* most frequent agent error in this surface.

dsarno:
[...]

Issue #210, opened 2026-04-26
Handler crash on bad input type surfaces as opaque INTERNAL_ERROR (e.g. add_to_group with non-string group)


dsarno:
Caught during a smoke of the post-#203 rollup surface (see latest entry in `friction_log.md`).

## Repro

```
node_manage(op="add_to_group", params={"path": "/Main/Box", "group": "[\"a\",\"b\"]"})
```

The agent passes a JSON-string-looking value for `group`. By the time it reaches the plugin, something upstream has decoded it — the recv log shows:

```
MCP | [recv] add_to_group({"group":["a","b"], "path":"/Main/Box"})
```

i.e. an actual `Array`, not a `String`. `node_handler.gd::add_to_group` has no input-type validation and lets GDScript blow up on the type mismatch. The dispatcher catches the bad return and emits the generic:

```
INTERNAL_ERROR: Handler 'add_to_group' returned malformed result (likely crashed — check Godot console)
```

Connection survives — this is **not** a server crash. But the agent has no actionable signal: the error doesn't say what the handler got, doesn't include a GDScript backtrace, and "check Godot console" isn't reachable from outside the editor.

## Why this is two bugs in one

1. **The handler is missing input validation.** `node_handler.gd::add_to_group` (and its siblings in the same family) should `assert(typeof(group) == TYPE_STRING)` at top and return `INVALID_PARAMS: group must be a string` instead of letting GDScript runtime-error.
2. **The `INTERNAL_ERROR: malformed result` message is a chronic UX trap.** Either capture `get_error_stack()` into the response's `data.gd_backtrace`, OR at minimum log the failing args to the plugin's log buffer so `logs_read` shows what blew up. Right now `logs_read` only shows the recv line and the error response — nothing about the actual GDScript runtime error.

The same coercion-then-handler-crash pattern almost certainly affects every other rollup op whose params include a JSON-stringified value where the handler expects a primitive. A sweep across all `*_handler.gd` files for missing `typeof()` guards on params sourced from JSON would be worthwhile.

## Suggested scope

- Fix `add_to_group` specifically as the reported case.
- Either land a small `_param_validators.gd` helper (`require_string(name, val)`, `require_int(name, val)`, etc.) that handlers can call uniformly, or audit the handlers and inline `assert(typeof(...) == TYPE_*)` at each site.
[...]

PR #217, opened 2026-04-26
fix(handlers): filter editor-internal entries from signal/input_map list ops (#213)

### LINKED ISSUE #213: List ops surface editor-internal entries (signal_manage list, input_map_manage list)

dsarno:
## Summary

Closes #213.

Two `list` ops were surfacing editor-runtime state as if it were user content. This bundles both fixes in one PR — the issue author called the bundling out explicitly ("Worth doing both in one PR — same conceptual change applied to two surface ops").

### `signal_manage(op="list")` was returning SceneTreeEditor observer wiring

```
signal_manage(op="list", path="/Main")
```

returned ~5 connections targeting paths like
`/Main/../../../../DockSlotLeftUR/Scene/.../@SceneTreeEditor@4919` — the editor's own SceneTree dock listening for `child_order_changed` / `editor_state_changed`. An agent reading the response would think the user wired up SceneTreeEditor by hand.

**Fix**: drop connections whose `Node` target isn't in the edited scene tree and isn't a declared autoload. New opt-in:

- `include_editor=False` (default) — drops them, returns `editor_connection_count` so callers know how many were hidden.
- `include_editor=True` — surfaces them with a per-connection `is_editor: bool` flag for filtering.
- Non-Node targets (RefCounted listeners, state machines, plugin classes) stay visible: they're typically legitimate script wiring, and hiding them by default could mask real bugs.

### `input_map_manage(op="list")` was returning `spatial_editor/*` actions

```
input_map_manage(op="list", params={})  # default include_builtin=False
```

returned 16 `spatial_editor/*` entries (`freelook_left`, `viewport_orbit_modifier_1`, etc.) in projects with zero user-defined input actions. The old filter only excluded the `ui_*` namespace; other editor-runtime namespaces leaked through.

**Fix**: `is_builtin` now means "not authored by the user" — i.e. not present as `ProjectSettings.has_setting("input/" + name)`. This catches `ui_*`, `spatial_editor/*`, and any future editor-runtime namespace Godot adds without further code changes here.

The `godot://input_map` MCP resource picks up the new filter automatically since it routes through the same handler.

### Tests

- Python unit: `signal_list` forwards the new `include_editor` kwarg to the plugin command.
- Python integration: `signal_manage(op="list", params={..., "include_editor": True})` round-trips with `editor_connection_count` in the response.
- GDScript: signal list filters editor observers, restores them with `include_editor=true`, and the count of `is_editor: true` entries matches the dropped count under the default filter. Input list excludes `spatial_editor/*` by default and surfaces non-user actions when `include_builtin=true`.

## Test plan

- [x] `ruff check src/ tests/` — passes
- [x] `pytest -q` — 610 passed (8 new)
- [ ] **Live smoke deferred**: this environment has no Godot binary, so the GDScript test suites and an interactive `signal_manage` / `input_map_manage` call against a live editor weren't run. Worth the next reviewer running `test_run` from `test_project/` and the two friction repros from #213 before merge.

## Triage notes for the rest of the open issues

- **#214** (AABB / packed array variant→JSON coercion) — single-PR candidate, scope clean.
- **#212** (ParseStringifiedParams fully-stringified `params`) — appears already fixed by #207. The middleware at `src/godot_ai/middleware/parse_stringified_params.py:36-45` decodes a string `params` slot before Pydantic; integration tests `TestManageRollupAcceptsStringifiedParams` exercise the exact repro and pass on this branch's parent commit. The issue body's claim that the middleware "only walks values inside `params`" is inverted — it does the outer decode and only the outer decode. Recommend closing as already-fixed after a quick verifier run.
- **#211** (op typo → no "Did you mean") — single-PR candidate, isolated to FastMCP middleware (Option A in the issue).
- **#210** (handler crash + opaque INTERNAL_ERROR) — bigger scope, deserves its own bisect-friendly PR (input-validation sweep + dispatcher backtrace surfacing).

https://claude.ai/code/session_019hTHwqTnRLzGSXPKNEfUk1

---
_Generated by [Claude Code](https://claude.ai/code/session_019hTHwqTnRLzGSXPKNEfUk1)_

### DIFF
--- a/plugin/addons/godot_ai/handlers/input_handler.gd
+++ b/plugin/addons/godot_ai/handlers/input_handler.gd
@@ -11,7 +11,12 @@ func list_actions(params: Dictionary) -> Dictionary:
 	var actions: Array[Dictionary] = []
 	for action_name in InputMap.get_actions():
 		var name_str := str(action_name)
-		var is_builtin := name_str.begins_with("ui_")
+		## "Built-in" = not authored by the user. Includes ui_* and editor-
+		## runtime namespaces like spatial_editor/* that InputMap exposes
+		## but project.godot doesn't store. ProjectSettings.has_setting is
+		## the authoritative origin check — it survives future Godot
+		## additions of new editor namespaces too.
+		var is_builtin := not ProjectSettings.has_setting("input/" + name_str)
 		if is_builtin and not include_builtin:
 			continue
 		var events: Array[Dictionary] = []
--- a/plugin/addons/godot_ai/handlers/signal_handler.gd
+++ b/plugin/addons/godot_ai/handlers/signal_handler.gd
@@ -51,7 +51,7 @@ func list_signals(params: Dictionary) -> Dictionary:
 					continue
 			connections.append({
 				"signal": sig.name,
-				"target": ScenePath.from_node(target, scene_root) if target is Node else str(target),
+				"target": _serialize_target(target, scene_root),
 				"method": callable.get_method(),
 				"is_editor": is_editor_target,
 			})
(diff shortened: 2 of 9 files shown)


dsarno:

---

PR #220, opened 2026-04-26
fix(handlers): drop editor-internal entries from list_signals/list_actions

### LINKED ISSUE #213: List ops surface editor-internal entries (signal_manage list, input_map_manage list)

dsarno:
## Summary

Closes #213. Bundles the two cases the issue itself recommended bundling — same conceptual fix (filter editor-runtime state out of "list" ops by default) applied to two surface ops.

Before this PR:

- `signal_manage(op="list", params={"path": "/Main"})` returned 5 `SceneTreeEditor` dock observer connections (`child_order_changed`, `editor_state_changed`, `script_changed`, …) with target paths escaping the edited scene root via `/Main/../../../../EditorNode/...`. An agent reading `connection_count: 5` would assume the user wired these up.
- `input_map_manage(op="list")` (default `include_builtin=False`) returned 16 `spatial_editor/*` entries on a project with zero user-defined actions. The old filter only stripped `ui_*`, missing every other editor-runtime namespace.

### `plugin/addons/godot_ai/handlers/signal_handler.gd`
- New static `_connection_origin(target, scene_root) -> "scene" | "autoload" | "editor"`. Walks up to the SceneTree root child to distinguish autoloads from editor docks.
- `list_signals` tags every connection with `origin`, drops `origin == "editor"` unless `include_editor_internal=true`, and surfaces `editor_connection_count` either way so callers can see how many were filtered.

### `plugin/addons/godot_ai/handlers/input_handler.gd`
- `list_actions` redefines "user-defined" as `ProjectSettings.has_setting("input/<name>")`. The old `name.begins_with("ui_")` heuristic missed `spatial_editor/*` entirely; the new check filters them and any future editor-runtime namespace.
- Behavior change for one edge case: a user who customized `ui_accept` (writing an override into `project.godot`) will now see it in the default list. That's the more correct semantics — it's their customization.

### Python + descriptions
- `signal_list` handler grows an `include_editor_internal: bool = False` kwarg.
- Tool descriptions for `signal_manage` and `input_map_manage` updated; `godot://input_map` resource docstring updated.

## Test plan

- [x] `pytest -q` — 609 passed, including new `TestSignalListEditorInternalFilter` class
- [x] New GDScript tests added: `test_list_signals_filters_editor_internal_by_default`, `test_list_signals_include_editor_internal_returns_them`, `test_list_actions_default_excludes_spatial_editor`, `test_list_actions_default_returns_user_defined_action`, `test_list_actions_with_builtins_includes_spatial_editor_when_present`
- [ ] **Live smoke against a real Godot editor was not possible from this sandbox** — no Godot binary available. Before merging, please run:
  - `test_run` via MCP — confirm all GDScript suites pass
  - `signal_manage(op="list", params={"path":"/Main"})` on a fresh `Node3D` scene — `connections` should be empty (or only autoloads), `editor_connection_count` should be ≥ 5
  - `input_map_manage(op="list", params={})` on a project with zero user actions — `count` should be 0
  - `input_map_manage(op="list", params={"include_builtin": true})` — should still surface `ui_*` and `spatial_editor/*` entries with `is_builtin: true`

https://claude.ai/code/session_01WUnuY27FZq21dM33rDCgPm

---
_Generated by [Claude Code](https://claude.ai/code/session_01WUnuY27FZq21dM33rDCgPm)_

### DIFF
--- a/plugin/addons/godot_ai/handlers/input_handler.gd
+++ b/plugin/addons/godot_ai/handlers/input_handler.gd
@@ -7,11 +7,17 @@ extends RefCounted
 
 
 func list_actions(params: Dictionary) -> Dictionary:
+	## Default returns only user-authored actions — those persisted to
+	## ``project.godot`` under ``input/<name>``. Editor-runtime actions like
+	## ``ui_*`` (built-ins) and ``spatial_editor/*`` (the 3D viewport's
+	## freelook/orbit/pan keybindings) live only on InputMap and are filtered
+	## out unless ``include_builtin`` is true.
 	var include_builtin: bool = params.get("include_builtin", false)
 	var actions: Array[Dictionary] = []
 	for action_name in InputMap.get_actions():
 		var name_str := str(action_name)
-		var is_builtin := name_str.begins_with("ui_")
+		var is_user_defined := ProjectSettings.has_setting("input/" + name_str)
+		var is_builtin := not is_user_defined
 		if is_builtin and not include_builtin:
 			continue
 		var events: Array[Dictionary] = []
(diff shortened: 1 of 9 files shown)


dsarno:

---

PR #225, opened 2026-04-26
Fix stale MCP session diagnostics and reconnect log spam

dialupdisaster:
## Summary
- Adds a structured, machine-actionable stale streamable-HTTP MCP session error so clients/LLMs can reinitialize instead of repeatedly retrying an invalid `mcp-session-id`.
- Reduces editor WebSocket reconnect log spam by extending capped backoff to 60s and logging only initial/periodic reconnect attempts while preserving indefinite automatic reconnect.
- Adds unit and GDScript coverage for stale-session response rewriting and reconnect backoff/log-throttling behavior.

## Problem
When a streamable-HTTP MCP client keeps using an `mcp-session-id` that the MCP SDK no longer knows about, the Python MCP SDK rejects the request before Godot AI tool handlers run. The current response is a plain JSON-RPC error:

```json
{
  "code": -32600,
  "message": "Session not found"
}
```

That error is technically correct, but it gives clients and LLM agents no clear recovery action. In practice, a client can keep retrying the same stale session forever even though a fresh MCP session would work.

Separately, the editor WebSocket reconnect loop currently retries forever with a 10-second cap and logs every attempt. If the WebSocket cannot reconnect overnight, this can produce thousands of repeated reconnect log lines.

## Why not silently recover the old MCP session?
Server-side resurrection of a stale streamable-HTTP session is not safe here. The missing `mcp-session-id` refers to transport/session-manager state that no longer exists. Reusing or silently recreating that session ID risks violating MCP initialization/session semantics.

The correct recovery is for the MCP client to initialize a fresh streamable-HTTP MCP session. This PR makes that explicit and machine-readable.

## New stale-session response
For the MCP SDK's stale session response, Godot AI now preserves the HTTP 404 and JSON-RPC `-32600` error code, but rewrites the message and adds structured recovery data:

```json
{
  "jsonrpc": "2.0",
  "id": "server-error",
  "error": {
    "code": -32600,
    "message": "MCP session expired or was not found; reinitialize the streamable HTTP session",
    "data": {
      "recoverable": true,
      "action": "reinitialize_mcp_session",
      "reason": "stale_streamable_http_session"
    }
  }
}
```

Clients and LLM tooling can key off:

```text
error.data.action == "reinitialize_mcp_session"
```

and create a fresh MCP session instead of retrying the same stale one.

## Reconnect behavior change
The editor WebSocket reconnect loop still retries indefinitely, but less noisily:

[...]

PR #227, opened 2026-04-26
fix(asgi): address Copilot feedback on stale-session middleware (#225)

dsarno:
Addresses the four Copilot review comments on #225. Built on top of #225's commits, so this branch contains the original PR plus these fixes.

## Copilot feedback addressed

1. **`_send_response` types `start_message` as `Message | None` but always uses it.** Tightened the parameter (and `_rewrite_stale_session_body`) to non-`Optional` since the call site already guarantees non-None via the `if start_message is None` guard in `capture_send`.

2. **`_is_sdk_session_not_found` matched only on `error.message`, which is too broad.** Now also requires `id == "server-error"` and `error.code == -32600`. An unrelated JSON-RPC 404 that happens to share the message text will no longer be rewritten.

3. **Workflow duplication.** Deleted `.github/workflows/pr-checks.yml` — `ci.yml` already triggers on `pull_request` to `main` and runs `ruff check src/ tests/` + the full pytest suite (across two Python versions, with coverage). The new workflow added no signal but tripled CI runtime.

4. **Middleware buffered all HTTP 404 responses, not just stale-session ones.** Now gated on the request having an `mcp-session-id` header — the SDK only emits "Session not found" when the client sent that header, so any 404 without one passes through with original streaming/timing semantics.

## Test changes

- Added `test_stale_mcp_session_diagnostic_passes_through_when_request_has_no_session_id` — proves the new gate doesn't rewrite even a perfect stale-looking body when the header is absent.
- Added `test_stale_mcp_session_diagnostic_requires_matching_jsonrpc_id` — proves the tighter matcher rejects a payload whose only difference from the SDK's response is the `id` field.
- Updated the chunked-body test and the two "leaves other 404s unchanged" tests to send the `mcp-session-id` header so they actually exercise the rewriter's reject path (not just the gate).

## Validation

- `pytest -q` — 632 passed (was 630 before; +2 new tests)
- Middleware coverage stays at 100% (the new gate and the tightened matcher are both exercised by the new and updated tests).

## Notes

- Did not touch the `connection.gd` reconnect changes from #225 — Copilot didn't flag them and they look right.
- The `__getattr__` proxy on the middleware is still there; it's covered by `test_stale_mcp_session_diagnostic_preserves_fastmcp_app_state` and required for FastMCP's runner to read `app.state` before handing it to uvicorn.

https://claude.ai/code/session_01Lnxpctxzfc5sY5rBUQitXJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lnxpctxzfc5sY5rBUQitXJ)_


dsarno:
Closing in favor of posting the patch directly on #225 (https://github.com/hi-godot/godot-ai/pull/225#issuecomment-4322969646) so credit stays with the original PR author. The patch contains the same four Copilot-feedback fixes plus the coverage tests.

---

PR #228, opened 2026-04-26
Fix editor refocus client status refresh

### LINKED ISSUE #226: Investigate editor freeze after returning focus with Godot AI enabled
## Summary

When Godot AI is enabled, the Godot editor can freeze/stutter for a few seconds after the user tabs away for a while and then returns focus to the editor. This has been observed on both Windows and macOS. Disabling the Godot AI plugin appears to stop the behavior.

This issue is investigatory for now. I am not planning to start a PR until #225 lands, but wanted to capture the findings in case someone else wants to pick it up first.

## Observed behavior

- Leave the Godot editor unfocused for some time.
- Tab/click back into the editor.
- Editor freezes/stutters for a few seconds.
- Reproduced on Windows and macOS.
- Reportedly only happens when the Godot AI plugin is enabled.

## Most likely suspect

A static code pass points at the focus-in client status refresh path:

```text
mcp_dock.gd::_notification(NOTIFICATION_APPLICATION_FOCUS_IN)
  -> _refresh_all_client_statuses.call_deferred()
  -> _refresh_all_client_statuses()
  -> per-client McpClientConfigurator.check_status(...)
```

That path appears to run synchronous config/status work on the editor main thread. The highest-risk bit is the Claude Code client status check, which can invoke a blocking CLI call:

```gdscript
OS.execute(cli, ["mcp", "list"], output, true)
```

Since `OS.execute(..., true)` blocks, a slow CLI startup, auth/config read, shell/PATH lookup, antivirus scan, cloud-synced profile access, or update check could stall the editor exactly when focus returns.

## Secondary suspects worth measuring

- `connection.gd::_process()` drains all available WebSocket packets in one frame. If packets accumulate while the editor is backgrounded/throttled, refocus may process a burst on the main thread.
- Developer log UI catch-up may append many buffered lines when visible.
- Godot itself has known editor refocus/filesystem rescan behavior that may amplify plugin-side synchronous file reads.

## Useful upstream context / references

- Godot macOS editor refocus lag / filesystem rescan discussion: https://github.com/godotengine/godot/issues/104246
- Godot filesystem/update storm freezes: https://github.com/godotengine/godot/issues/92485
- Related older freeze issue: https://github.com/godotengine/godot/issues/92081
- `OS.execute()` is blocking when `blocking=true`: https://docs.godotengine.org/en/stable/classes/class_os.html

## Suggested diagnostic pass

Before changing behavior, add slow-path timing logs around:

- focus-in notification handling
- `_refresh_all_client_statuses()` total duration
- each individual client status check duration and client id
- every plugin-side `OS.execute()` call duration and command category
- WebSocket `_peer.poll()` duration
- packet count/drain duration in `connection.gd::_process()`
- dispatcher queue length before/after refocus
- log UI append line count and duration

Suggested thresholds:

- log client status checks >50ms
- warn for client checks >250ms
- treat any focus-in sweep >1000ms as a confirmed freeze contributor
- log WebSocket drains if packet count is high or drain time exceeds ~4ms

## Likely mitigation if confirmed

If the focus-in sweep is confirmed:

- Do not run blocking CLI checks automatically on focus-in.
- Cache client status and only refresh on explicit user action or when the Clients panel opens.
- Time-slice any automatic client refresh across frames.
- Prefer config-file reads over CLI calls where possible.
- Add a timeout/unknown state for slow client checks rather than blocking the editor.
- Consider capping WebSocket packet draining per frame if backlog is also confirmed.

dialupdisaster:
## Summary
- Keeps automatic client status refresh on editor focus-in, but moves status probes to a background thread so blocking CLI checks do not freeze the editor.
- Adds stale-while-refreshing UI behavior, one-in-flight coalescing, and a 15s cooldown for automatic refocus refreshes while preserving manual refresh.
- Adds focused regression tests for refocus behavior and client-status refresh semantics.

## Validation
- `uv run ruff check src tests`
- `uv run pytest tests/unit/test_asgi_session_diagnostics.py tests/unit/test_cli_reload.py tests/unit/test_editor_focus_refocus.py` — 29 passed
- `uv run pytest tests` — 637 passed
- Manual validation in Godot: repeated tab-away/tab-back no longer freezes while Godot-AI is enabled

Fixes #226

### DIFF
--- a/plugin/addons/godot_ai/mcp_dock.gd
+++ b/plugin/addons/godot_ai/mcp_dock.gd
@@ -11,6 +11,7 @@ const DEV_MODE_SETTING := "godot_ai/dev_mode"
 const MODE_OVERRIDE_VALUES := ["", "user", "dev"]
 const MODE_OVERRIDE_LABELS := ["Auto", "Force user", "Force dev"]
 const CLIENT_STATUS_REFRESH_COOLDOWN_MSEC := 15 * 1000
+const CLIENT_STATUS_REFRESH_TIMEOUT_MSEC := 30 * 1000
 static var COLOR_MUTED := Color(0.7, 0.7, 0.7)
 static var COLOR_HEADER := Color(0.95, 0.95, 0.95)
 ## Used for "in-progress" / "stale, action needed" UI: the startup-grace
@@ -90,6 +91,11 @@ var _client_status_refresh_in_flight := false
 var _client_status_refresh_pending := false
 var _client_status_refresh_pending_force := false
 var _last_client_status_refresh_completed_msec: int = 0
+var _client_status_refresh_started_msec: int = 0
+var _client_status_refresh_generation: int = 0
+var _client_status_refresh_shutdown_requested := false
+var _client_status_refresh_timed_out := false
+static var _orphaned_client_status_refresh_threads: Array[Thread] = []
 
 # Dev-mode only
 var _dev_section: VBoxContainer
@@ -158,9 +164,22 @@ func _ready() -> void:
 func _process(_delta: float) -> void:
 	if _connection == null:
 		return
+	_prune_orphaned_client_status_refresh_threads()
+	_check_client_status_refresh_timeout()
 	_update_status()
 	if _log_section.visible:
 		_update_log()
+
+
+func _exit_tree() -> void:
+	_client_status_refresh_shutdown_requested = true
+	_client_status_refresh_generation += 1
+	if _client_status_refresh_thread != null:
+		_orphaned_client_status_refresh_threads.append(_client_status_refresh_thread)
+		_client_status_refresh_thread = null
+	_client_status_refresh_in_flight = false
+	_client_status_refresh_pending = false
+	_client_status_refresh_pending_force = false
 
 
 func _notification(what: int) -> void:
@@ -509,7 +528,7 @@ func _build_ui() -> void:
 	# Apply initial dev-mode visibility
 	_apply_dev_mode_visibility()
 	_refresh_setup_status.call_deferred()
-	_refresh_clients_summary.call_deferred()
+	_request_client_status_refresh.call_deferred(true)
 
 
 func _make_header(text: String) -> Label:
@@ -574,6 +593,7 @@ func _build_client_row(client_id: String) -> void:
 
 	_client_rows[client_id] = {
 		"dot": dot,
+		"status": McpClient.Status.NOT_CONFIGURED,
 		"name_label": name_label,
 		"configure_btn": configure_btn,
 		"remove_btn": remove_btn,
@@ -1097,9 +1117,11 @@ func _on_refresh_clients_pressed() -> void:
 
 
 func _on_configure_all_clients() -> void:
+	if _client_status_refresh_in_flight:
+		return
 	for client_id in _client_rows:
-		var dot: ColorRect = _client_rows[client_id]["dot"]
-		if dot.color == Color.GREEN:
+		var status: McpClient.Status = _client_rows[client_id].get("status", McpClient.Status.NOT_CONFIGURED)
+		if status == McpClient.Status.CONFIGURED:
 			continue
 		_on_configure_client(String(client_id))
 	_refresh_clients_summary()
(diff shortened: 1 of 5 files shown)


plugin/addons/godot_ai/mcp_dock.gd
```diff
@@ -503,7 +509,7 @@ func _build_ui() -> void:
 	# Apply initial dev-mode visibility
 	_apply_dev_mode_visibility()
 	_refresh_setup_status.call_deferred()
-	_refresh_all_client_statuses.call_deferred()
+	_refresh_clients_summary.call_deferred()
```
dsarno: **Initial-paint regression.** Pre-PR, `_build_ui` ended with `_refresh_all_client_statuses.call_deferred()`, so dock dots reflected real status as soon as the editor finished startup. Now only the summary label is refreshed, and dot colors stay at their default until the first `NOTIFICATION_APPLICATION_FOCUS_IN`.

On a cold open of an editor that already has focus, focus-in never fires until the user alt-tabs (or clicks the new Refresh button). Suggest calling `_request_client_status_refresh(true)` here — now safe because it's async, and you'll get the same first-frame UX as before without any blocking work on startup.

---

plugin/addons/godot_ai/mcp_dock.gd
```diff
@@ -1403,15 +1414,83 @@ func _on_copy_manual_command(client_id: String) -> void:
...
+		Callable(self, "_run_client_status_refresh_worker").bind(client_ids, server_url)
+	)
+	if err != OK:
+		_client_status_refresh_in_flight = false
+		_client_status_refresh_thread = null
+		return false
+	return true
+
+
+func _run_client_status_refresh_worker(client_ids: Array[String], server_url: String) -> void:
```
dsarno: **No timeout on the worker — risk of permanent in-flight stickiness.** `OS.execute(..., true)` inside `cli_status_check` has no timeout. If `claude mcp list` hangs (auth prompt, network stall, AV/Defender delay, hung shim), the worker thread sits in `OS.execute` forever. Then:

- `_client_status_refresh_in_flight` never clears
- every automatic refocus coalesces into `pending = true` that never fires
- the manual Refresh button does nothing (it short-circuits at the in-flight guard in `_request_client_status_refresh`)
- the dock is stuck on stale dots until the plugin is reloaded

Two ideas, neither a merge-blocker but worth a follow-up:

1. **Watchdog**: track refresh start time; if a refresh hasn't completed within e.g. 30s, surface a "client probe still running" hint and allow a *forced* refresh to abandon the orphan slot. The orphan thread keeps running but no longer blocks new requests. Trickier to do safely — you have to either cancel the in-flight `call_deferred` (`_apply_client_status_refresh_results` would otherwise stomp newer state) or have the apply path no-op when its run id is stale. A run-id counter incremented per refresh request, captured by the worker, and checked in `_apply_client_status_refresh_results` is the cleanest pattern.
2. At minimum, give the manual Refresh button visible "checking…" feedback while in-flight so users have a signal something *is* running. Today there's no UI difference between "in progress" and "stuck."

---

dsarno (commented):
Reviewed alongside Copilot's automated review. Approach is right — moving the probes off the editor main thread is the correct fix, and the stale-while-refreshing + single-flight coalescing design is well-judged. Thanks for the careful work.

One blocker before merge, plus a few smaller issues. Copilot caught three of them; agreeing with all three:

1. **[Blocker] Lifecycle cleanup for `_client_status_refresh_thread`** — Copilot's comment at line ~1458 is right. If the dock is freed mid-refresh (plugin disable from Project Settings, plugin reload via `editor_reload_plugin`, editor shutdown), the worker's `call_deferred("_apply_client_status_refresh_results", ...)` lands on a freed Object and crashes; `_client_status_refresh_in_flight` can also stick true across reload.

   Caveat on the suggested fix: a blind `wait_to_finish()` in `_exit_tree()` re-introduces the very freeze this PR fixes — a hung `claude mcp list` would block editor close indefinitely. Prefer a **cancel flag** the worker checks before `call_deferred`, plus an `is_instance_valid(self)` guard in `_apply_client_status_refresh_results`. Use `wait_to_finish` only as a fallback once the flag has been set and the worker is known to be near completion.

2. **`McpCliFinder` static cache mutated from worker** — Copilot's race comment at line ~1468. Today single-flight gating means only one worker mutates `_cache`/`_searched` at a time, but other main-thread paths (`is_installed` from elsewhere, the manual refresh button's pre-checks, etc.) could race with it. Cleanest fix: **pre-resolve CLI paths on the main thread** at the top of `_request_client_status_refresh` and pass the resolved string into the worker. Keeps `_run_client_status_refresh_worker` free of `McpCliFinder` calls entirely — no mutex needed.

3. **`_on_configure_all_clients` reads stale dot colors** — Copilot's comment at line ~1105. Agree. Stash the last-known `Status` per row dict alongside `dot`/`name_label`/etc. and switch on that. As-is, the button can re-Configure a client whose cached dot is wrong (e.g. just after `claude mcp remove` was run externally, before the next refresh lands).

Two more points Copilot didn't catch — see inline:

4. **Initial-paint regression** at line 512 — `_build_ui` no longer kicks off a status sweep, so on cold open dots stay default until the user alt-tabs.
5. **No worker timeout** at line 1461 — a hung `claude mcp list` pins `_client_status_refresh_in_flight = true` indefinitely.

**Tests note:** `tests/unit/test_editor_focus_refocus.py` is grep-against-source — it will catch accidental removal of the constants/wiring but not behavioral breakage. Manual validation claim is plausible and CI is green across all three OSes, so not a blocker. Worth a future test that injects a fake client whose `cli_status_check` sleeps and asserts the editor stays responsive — the kind of test that would have caught this regression in the first place.

Verdict: blocker (#1) plus the cache pre-resolve (#2) before merge; the rest are good follow-ups.

---

dialupdisaster:
Applied the review follow-ups in 118075b:

- Restored cold-open initial client status refresh via async `call_deferred(true)`.
- Pre-resolve CLI paths/installed state on the main thread before worker refreshes so the worker avoids `McpCliFinder` cache mutation and `is_installed()` probes.
- Added cached row status values so Configure all does not depend on stale dot colors; it is disabled while a refresh is in flight.
- Added a 30s refresh watchdog with generation IDs, stale-result no-op handling, and orphan-thread pruning so a hung CLI probe does not permanently block manual refreshes.
- Added dock shutdown cleanup to mark in-flight workers stale and prevent late result application.

Local validation:
- `uv run ruff check src tests`
- `uv run pytest tests/unit/test_editor_focus_refocus.py` (8 passed)
- `uv run pytest tests` (641 passed)
- `"/Applications/Godot_mono.app/Contents/MacOS/Godot" --headless --path "test_project" --import` (no GDScript parse errors; existing test_project autoload/plugin-path warnings remain unrelated)

dsarno:
This is going to be joined by https://github.com/hi-godot/godot-ai/pull/230, which refactors a bunch of stuff with client checking hotpaths -- so this overall bug and fix helped me smoke out a deeper architectural issue with the client configs -- so, very helpful, thanks again!

Issue #233, opened 2026-04-27
Editor crashes on self-update: worker thread races with GDScript hot-reload of strategies


dsarno:
## Summary

Editor crashes (`SIGABRT` in worker thread) on the dock's self-update path immediately after the new plugin is loaded. Reproduced on a real user update from v2.0.1 → v2.1.0 minutes after the v2.1.0 release was published.

This is the residual leg of #229 that was [explicitly flagged as untested](https://github.com/hi-godot/godot-ai/issues/229#issuecomment-4323327346) when #230 / #232 landed. Filing it for closure.

## Repro
1. Run a project with godot-ai v2.0.1 (or any pre-#230 version) installed.
2. Click **Update** in the dock when a newer release banner appears.
3. `_install_update` extracts the new release zip over `addons/godot_ai/`, then calls `set_plugin_enabled(false)` followed by `(true)`.
4. Editor crashes ~5–10 seconds later.

## Crash report (macOS DiagnosticReports)

```
Process:           Godot 4.6.2
OS:                macOS 26.4.1 (25E253)
Exception:         EXC_BAD_ACCESS / SIGABRT
Subtype:           KERN_INVALID_ADDRESS at 0x0000000000000018
Crashed thread:    VBoxContainer(McpDock)::_run_client_status_refresh_worker

Crashed thread stack (top):
  Thread::thread_callback
  CoreBind::Thread::_start_func
  CallableCustomBind::call
  Callable::callp
  Object::callp
  GDScriptInstance::callp
  GDScriptFunction::call          ← bytecode walked off
  GDScriptFunction::call
  GDScriptFunction::call
  GDScriptFunction::call
  __abort
  __pthread_kill
```

The thread name confirms it's the auto-refresh worker spawned by `_build_ui`'s `_request_client_status_refresh.call_deferred(true)`.

## Root cause

The self-update flow extracts dozens of `.gd` files over `addons/godot_ai/` on disk, then triggers a plugin enable that creates a fresh `McpDock` and immediately fires the auto-refresh worker. Godot's GDScript hot-reload of the just-overwritten files is **lazy** — script-class registration runs to completion before enable returns, but the actual bytecode replacement of arbitrary referenced scripts happens as those scripts are first executed/dereferenced post-reload. The worker walks into `_json_strategy.check_status` / `client_configurator.*` / etc. while their bytecode pages are mid-swap → invalid IP → abort.

[...]

Issue #229, opened 2026-04-26
Worker thread races with GDScript hot-reload during plugin reload (follow-up to #228)


dsarno:
## Summary
The async client-status refresh worker introduced in #228 races with Godot's GDScript hot-reload. When the plugin's `.gd` files change on disk while the worker is mid-probe (e.g. `git checkout`, in-IDE save of a plugin script, the dock's self-update flow), the editor's main thread swaps script bytecode out from under the running worker, producing IP-bounds errors, `Opcode: 0` script errors, and intermittent `SIGSEGV`.

This is a separate crash from the static-var orphan bug fixed in [crazyfacestudios#2](https://github.com/crazyfacestudios/godot-ai/pull/2) (follow-up to #228). That fix prevents `Thread`s from being GC'd mid-execution on plugin disable; this issue covers the orthogonal "worker is running while files are being reloaded" race, which can fire even when the threading lifecycle is otherwise clean.

## Severity
Developer workflow only — real users don't `git checkout` mid-editor session. Triggers in:

- `git checkout` / `git stash` / rebase from a terminal while an editor with the plugin loaded is running
- Editing a plugin `.gd` file in another IDE and saving while the editor is open
- The dock's self-update flow (`_install_update`) which extracts a release zip over the addons dir, then calls `set_plugin_enabled(false/true)`
- Possibly any user who triggers a plugin reload right after a CI/script that updates the addons dir

The first three are routine for plugin developers. The self-update path in particular is the one user-facing trigger — worth checking whether the staggered scan + reload there actually settles before the new dock kicks off its first refresh.

## Repro

Against the [crazyfacestudios#2](https://github.com/crazyfacestudios/godot-ai/pull/2) head (i.e. with the `_exit_tree` thread-drain fix already applied):

1. Open the editor with the plugin loaded.
2. From a terminal in the same checkout, swap the plugin's `mcp_dock.gd` to a different revision: `git checkout <other-branch> -- plugin/addons/godot_ai/mcp_dock.gd`
3. Call `editor_reload_plugin` (or toggle the plugin in Project Settings).
4. Editor stdout shows:

```
SCRIPT ERROR: Bad address index.
   at: suggest_free_port (res://addons/godot_ai/clients/_cli_strategy.gd:118)
   GDScript backtrace (most recent call first):
       [0] suggest_free_port (res://addons/godot_ai/client_configurator.gd:118)
       [1] _exit_tree (res://addons/godot_ai/mcp_dock.gd:240)
[...]

PR #234, opened 2026-04-27
fix(dock): defer first auto-refresh past GDScript hot-reload settle (#233)

dsarno:
## Summary

**Interim heuristic stopgap** for #233 — editor crashes (`SIGABRT` in worker thread) on the dock's self-update path immediately after the new plugin loads. Reproduced in the wild minutes after v2.1.0 was tagged: real user updated v2.0.1 → v2.1.0, install completed, editor died with:

```
Exception: EXC_BAD_ACCESS / SIGABRT
Subtype:   KERN_INVALID_ADDRESS at 0x18
Thread:    VBoxContainer(McpDock)::_run_client_status_refresh_worker
Stack:     Thread::_start_func
           → CallableCustomBind::call
           → 4× nested GDScriptFunction::call
           → __abort
```

> ⚠️ **Caveat (not bulletproof):** the 1.5s delay is empirical, not derived from any Godot signal that confirms hot-reload completion. On a slow CI runner, a larger plugin tree, or a future Godot version that changes hot-reload timing, this race could resurface. **A deterministic fix is tracked in #235** (investigate `EditorFileSystem.script_classes_updated`, fall back to sync first refresh on main thread, or refactor strategies to share no GDScript state with the worker). This PR ships the timer-based fix because:
> - The observed crash repros at 100% on the affected user path
> - Every hour without a fix risks more user crashes
> - Investigating + verifying a deterministic gate takes meaningful effort
> - The timer is a defensible interim that takes us from 100% crash rate to "no observed crashes on tested systems"

## Root cause

The self-update flow extracts a release zip over `addons/godot_ai/` then toggles `set_plugin_enabled(false/true)`. Godot's GDScript hot-reload of the just-overwritten files is **lazy** — class registration runs to completion before enable returns, but bytecode replacement of arbitrary referenced scripts happens as those scripts are first dereferenced post-reload. The new dock's `_build_ui` was firing the auto-refresh worker via `call_deferred(true)` (next idle frame), and the worker walks deep into `_json_strategy.check_status` / `client_configurator.*` — exactly the scripts still mid-swap.

This is the residual leg of #229 that #230 and #232 didn't close:

| PR | Side | What it fixed |
|---|---|---|
| #230 | both | Per-client `clients/<name>.gd` Callables → data-only |
| #232 | **disable** | Worker draining via `_exit_tree.wait_to_finish` |
| **this** | **enable** | New worker spawn deferred past hot-reload settle |

## Fix

Replace the immediate `_request_client_status_refresh.call_deferred(true)` in `_build_ui` with a 1500ms timer-deferred call. Empirical settle is <500ms; 1500ms is 3× margin.

```gdscript
func _schedule_initial_client_status_refresh() -> void:
    await get_tree().create_timer(CLIENT_STATUS_REFRESH_INITIAL_DELAY_MSEC / 1000.0).timeout
    if _client_status_refresh_shutdown_requested:
        return
    if not is_inside_tree():
        return
    _request_client_status_refresh(true)
```

## Why a signal-based gate wasn't used (yet)

- `is_scanning()` / `filesystem_changed` flip back to false **before** the bytecode swap completes — confirmed empirically in [#229's discussion](https://github.com/hi-godot/godot-ai/issues/229#issuecomment-4323327346).
- `NOTIFICATION_APPLICATION_FOCUS_IN` doesn't fire on plugin reload (editor stays focused through the toggle), so dropping the auto-spawn entirely would leave the dock unpopulated.
- `EditorFileSystem.script_classes_updated` is the most promising deterministic signal but is **unverified** to bracket the hot-reload window — see #235 for the investigation plan.

A timer is the only mechanism that actually works in observed conditions today. #235 tracks replacing it.

## Test surface

- **Updated** Python test `test_initial_paint_requests_async_status_refresh` — locks in the new contract: `_build_ui` must call the helper, which must reference the delay constant and ultimately call `_request_client_status_refresh(true)`.
- **Added** GDScript regression test `test_initial_refresh_delay_is_past_typical_hot_reload_settle` — asserts the constant is ≥1000ms so a "0-delay would be snappier" refactor can't silently re-introduce the crash.
- 911/914 GDScript suite, 652 pytest, ruff clean.

## Test plan

- [ ] CI: pytest + ruff (Linux/macOS/Windows × py3.11/3.13)
- [ ] CI: GDScript suite via `script/ci-godot-tests` on all three OSes
- [ ] Manual: install pre-#233 plugin (e.g. v2.1.0), trigger an in-place update via the dock — should now complete without crash. The dock briefly shows cached state for ~1.5s before the dots refresh.

## Follow-up
- **#235** tracks the deterministic replacement for this timer-based fix.

### DIFF
--- a/plugin/addons/godot_ai/mcp_dock.gd
+++ b/plugin/addons/godot_ai/mcp_dock.gd
@@ -12,15 +12,11 @@ const MODE_OVERRIDE_VALUES := ["", "user", "dev"]
 const MODE_OVERRIDE_LABELS := ["Auto", "Force user", "Force dev"]
 const CLIENT_STATUS_REFRESH_COOLDOWN_MSEC := 15 * 1000
 const CLIENT_STATUS_REFRESH_TIMEOUT_MSEC := 30 * 1000
-## Delay before the very first auto-refresh fires after `_build_ui`. Godot's
-## lazy GDScript hot-reload of `addons/godot_ai/*.gd` doesn't fully settle by
-## the time the new dock is in the tree on the self-update path
-## (`_install_update` extracts a new release zip then toggles
-## `set_plugin_enabled`). Worker walking into a strategy whose bytecode is
-## mid-swap aborts the editor (KERN_INVALID_ADDRESS / SIGABRT, see #233).
-## 1.5s is comfortable margin past typical settle (<500ms empirically) while
-## still feeling instant in the dock.
-const CLIENT_STATUS_REFRESH_INITIAL_DELAY_SEC := 1.5
+## Delay before the very first auto-refresh fires after `_build_ui` —
+## settle margin past Godot's lazy GDScript hot-reload of plugin scripts
+## on the self-update path. Empirical settle is <500ms; 1500 is 3× margin.
+## See issue #233.
+const CLIENT_STATUS_REFRESH_INITIAL_DELAY_MSEC := 1500
 static var COLOR_MUTED := Color(0.7, 0.7, 0.7)
 static var COLOR_HEADER := Color(0.95, 0.95, 0.95)
 ## Used for "in-progress" / "stale, action needed" UI: the startup-grace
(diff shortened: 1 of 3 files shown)


dsarno:
Good catch — the `get_tree()` could be null on the test path where the dock is instantiated via `McpDockScript.new()` without being added to the tree (which is exactly what `test_dock.gd:18` does). The error was getting swallowed by the orphan-coroutine semantics so the suite still passed, but it left vestigial "null instance" errors in the editor log on every affected test.

Fixed in `0ea377b` by capturing `var tree := get_tree()` and bailing when null / not in tree before creating the timer. Thanks!

PR #237, opened 2026-04-27
feat(logs): add source="editor" capturing parse + @tool errors (#231)

### LINKED ISSUE #231: logs_read: capture editor-side script errors (parse errors, @tool, EditorPlugin)
## Problem

`logs_read` today exposes two sources:

- `source="plugin"` — MCP plugin's own recv/send/event traffic
- `source="game"` — game subprocess stdout/stderr/`push_error`/`push_warning`, captured by `game_helper.gd`'s autoloaded `Logger`
- `source="all"` — both, merged

Neither covers **editor-process** script output. That means when an LLM debugs:

- **Parse errors** (script fails to compile on save/load — red lines in Godot's Output panel)
- **`@tool` script errors** (runtime errors from tool scripts running inside the editor)
- **EditorPlugin runtime errors** (errors from user-authored editor plugins)

…it sees nothing in `logs_read` and concludes "the tool can't read editor logs" — even though the errors are clearly visible in Godot's Output panel.

This came up in a Discord report where the user pointed at `game_helper.gd:49`'s `if Engine.is_editor_hint(): return` guard as the culprit. That guard is correct (it's what makes game-side capture work), but the underlying confusion is real: there's no editor-side analog.

## What's needed

A new `source="editor"` (and inclusion in `source="all"`) that captures editor-process script errors and `push_error` / `push_warning` calls.

```python
logs_read(source="editor", count=50)   # parse errors + @tool/EditorPlugin runtime errors
logs_read(source="all", count=50)      # plugin + game + editor, merged
```

## Mechanism

`OS.add_logger` works in the editor process the same way it does in the game subprocess. Register a `Logger` subclass on plugin `_enter_tree` and drain into a new `EditorLogBuffer` analogous to `GameLogBuffer`. Godot's internal `print_error` chain — which parse errors flow through — routes to registered loggers.

## Gotchas to design around

1. **Timing.** Parse errors fired *before* the plugin loads (initial filesystem scan, scripts that fail on first project open) happen before `add_logger` is called. We'll miss those. Document the limitation; consider a one-shot sweep of `EditorFileSystem` parse state on plugin enable to backfill known-bad scripts.

2. **Noise filtering.** An editor-side logger sees every internal Godot warning, every editor message, plus user errors. Without filtering it's a firehose that buries signal. Filter to:
   - Errors/warnings whose `script_path` (or message-extracted path) lives under `res://` and ends in `.gd` or `.cs`
   - Drop messages from `addons/godot_ai/` itself to avoid feedback loops
   - Keep `push_error` / `push_warning` regardless of origin (user opted in)

3. **Structured fields.** Parse errors include file path, line, column, parser message. Surface those as structured fields (`{path, line, col, level, message}`) rather than a flat string blob, so the LLM can navigate to the offending line directly. Same shape used for runtime errors when available.

4. **Buffer lifecycle.** Editor errors persist across `project_run` (unlike game logs which clear on each run). Cap at the same 500-line ring used for plugin logs; surface `dropped_count`.

5. **Logger gating on Godot version.** `Logger` virtual class is 4.5+, same constraint as `game_helper.gd`. Gate behind `ClassDB.class_exists("Logger") and OS.has_method("add_logger")` and no-op gracefully on 4.4.

## Proposed API

```python
logs_read(
    count: int = 50,
    offset: int = 0,
    source: Literal["plugin", "game", "editor", "all"] = "plugin",
    since_run_id: str = "",
)
```

Response shape stays the same; entries from `source="editor"` carry `{"source": "editor", "level": "error|warning", "path": "res://foo.gd", "line": 42, "col": 8, "text": "Parser Error: ..."}` when structured fields are available, falling back to `{"text": ...}` only when they aren't.

## Tasks

- [ ] Spike: confirm `OS.add_logger` on editor-process side reliably catches parse errors (vs. only runtime `push_error`)
- [ ] Add `EditorLogBuffer` alongside `McpLogBuffer` and `GameLogBuffer`
- [ ] New `editor_logger.gd` analogous to `game_logger.gd`, attached in `plugin.gd::_enter_tree`
- [ ] Filter rules (user-script paths, exclude plugin's own output)
- [ ] Extract structured `{path, line, col}` from parser error format where possible
- [ ] Extend `logs_read` Python tool + GDScript handler to accept `source="editor"` and include it in `source="all"`
- [ ] Update `logs_read` tool description so the LLM picks the right source — "for parse errors and @tool script errors, use `source="editor"` or `source="all"`"
- [ ] Python unit + integration tests
- [ ] GDScript test: write a `.gd` file with a deliberate parse error, trigger reimport, assert it appears in `logs_read(source="editor")`
- [ ] Update `docs/TOOLS.md`

## Out of scope

- Streaming push-based subscription (poll via `logs_read` is enough)
- Capturing errors that fire before the plugin's `_enter_tree` (best-effort backfill via `EditorFileSystem` is a stretch goal, not a requirement)
- Catching errors from other plugins' loggers (ours is one of N — only what `print_error` routes through is in scope)

## Related

- #73 — original game-side `logs_read` gap (closed, shipped)
- CLAUDE.md "Game-side code: gate on `Engine.is_editor_hint()`" — relevant context for why the editor/game split exists

dsarno:
Closes #231.

## Summary
- New `logs_read(source="editor")` captures editor-process script errors that were previously invisible to LLM debugging — parse errors, `@tool` runtime errors, EditorPlugin errors, `push_error`/`push_warning`. Errors show in Godot's Output panel today; now they show in `logs_read` too.
- Implementation: a Godot 4.5+ `Logger` subclass registered via `OS.add_logger`, draining into a mutex-protected ring buffer. Filtered to user `.gd`/`.cs` paths; `addons/godot_ai/` dropped to break feedback loops.
- Structured entries `{source, level, text, path, line, function}` so the LLM can navigate straight to the offending line. `source="all"` interleaves plugin → editor → game.

## Known limitation (called out in #231)
Parse errors fired during the editor's initial filesystem scan happen before `_enter_tree` and aren't captured. Documented in the `_attach_editor_logger` docstring and the tool description.

## Follow-up
The new `editor_logger.gd` / `editor_log_buffer.gd` and the existing `game_logger.gd` / `game_log_buffer.gd` share ~75% of their bodies. Cross-cutting refactor (extract `script_backtrace.gd` + `RingBuffer` base) intentionally deferred — touches existing tested code, not in scope for this issue. Tracked locally; happy to spin up a follow-up PR.

## Test plan
- [x] `pytest -q` — 654 passed (incl. new `test_logs_read_handler_source_editor_passes_through`, updated `source="all"` test, integration `test_source_editor_returns_structured_script_errors`)
- [x] `test_run suite=editor` — 77 passed / 0 failed / 2 pre-existing scene-conditional skips
- [x] Live smoke on Godot 4.6.2: wrote a deliberately-broken `.gd`, called `filesystem_manage(op="reimport")`, confirmed `logs_read(source="editor")` returned `{level: "error", path: "res://_smoke_broken.gd", line: 2, function: "GDScript::reload", text: "Parse Error: Expected parameter name."}`
- [ ] Reviewer to spot-check on Linux/Windows if possible (capture path is the same Logger virtual either way, but worth one round-trip)

## Notes
- Code-quality / efficiency / reuse review pass before commit applied: typed positional `EditorLogBuffer.append`, single-lock `get_recent` (TOCTOU fix), constructor injection in `editor_logger`, cheap-reject filter ordering on the hot path, dropped redundant `_editor_logger_attached` bool, value-checked `dropped_count` assertion per CLAUDE.md test hygiene, compressed tool description.

### DIFF
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -29,7 +29,7 @@ not the MCP tool names.
 | `script_create` / `script_attach` / `script_patch` | Create, attach, anchor-edit GDScript files |
 | `project_run` | Play the project (autosave persists in-memory MCP edits unless `autosave=False`) |
 | `test_run` | Run GDScript test suites in the editor |
-| `logs_read` | Read plugin / game / combined log buffers (game source needs `_mcp_game_helper` autoload) |
+| `logs_read` | Read plugin / game / editor / combined log buffers. `source="editor"` surfaces parse errors + @tool/EditorPlugin runtime errors + push_error/push_warning (Godot 4.5+, filtered to user .gd/.cs) — use this when the editor's Output panel shows red lines but `logs_read` returned nothing |
 | `editor_screenshot` | Capture editor viewport, cinematic camera, or running game framebuffer |
 | `editor_reload_plugin` | Reload the plugin and wait for reconnect (server must be external) |
 | `animation_create` | Create an Animation clip (auto-creates AnimationPlayer + library if missing) |
--- a/plugin/addons/godot_ai/handlers/editor_handler.gd
+++ b/plugin/addons/godot_ai/handlers/editor_handler.gd
@@ -8,13 +8,15 @@ var _log_buffer: McpLogBuffer
 var _connection: Connection
 var _debugger_plugin: McpDebuggerPlugin
 var _game_log_buffer: GameLogBuffer
+var _editor_log_buffer: EditorLogBuffer
 
 
-func _init(log_buffer: McpLogBuffer, connection: Connection = null, debugger_plugin: McpDebuggerPlugin = null, game_log_buffer: GameLogBuffer = null) -> void:
+func _init(log_buffer: McpLogBuffer, connection: Connection = null, debugger_plugin: McpDebuggerPlugin = null, game_log_buffer: GameLogBuffer = null, editor_log_buffer: EditorLogBuffer = null) -> void:
 	_log_buffer = log_buffer
 	_connection = connection
 	_debugger_plugin = debugger_plugin
 	_game_log_buffer = game_log_buffer
+	_editor_log_buffer = editor_log_buffer
 
 
 func get_editor_state(_params: Dictionary) -> Dictionary:
@@ -43,7 +45,7 @@ func get_selection(_params: Dictionary) -> Dictionary:
 	return {"data": {"selected_paths": paths, "count": paths.size()}}
 
 
-const VALID_LOG_SOURCES := ["plugin", "game", "all"]
+const VALID_LOG_SOURCES := ["plugin", "game", "editor", "all"]
 
 
 func get_logs(params: Dictionary) -> Dictionary:
@@ -56,7 +58,7 @@ func get_logs(params: Dictionary) -> Dictionary:
 	if not source in VALID_LOG_SOURCES:
 		return McpErrorCodes.make(
 			McpErrorCodes.INVALID_PARAMS,
-			"Invalid source '%s' — use 'plugin', 'game', or 'all'" % source,
+			"Invalid source '%s' — use 'plugin', 'game', 'editor', or 'all'" % source,
 		)
 
 	match source:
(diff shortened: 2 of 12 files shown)


dsarno:
Closing as duplicate of #240, which was merged on 2026-04-28 and explicitly supersedes the `pensive-curran-7e7828` worktree this branch came from. Both PRs close #231.

Reviewed Copilot's 3 comments — none need separate follow-up:
- Backtrace iteration nit: already addressed in #240's `LogBacktrace.resolve_error()`, which iterates `script_backtraces` and picks the first non-empty frame.
- Two `.uid` trailing-newline nits: cosmetic; the `.uid` files merged via #240 are the canonical format.

---

PR #243, opened 2026-04-28
fix(plugin, dock): unbreak self-update across new-class_name releases (#242)

### LINKED ISSUE #242: SIGABRT during dock teardown on v2.1.x → v2.1.x+1 self-update (Dictionary::keys on freed dict)
## Summary

`SIGABRT` on the editor main thread during plugin teardown when the user clicks "Update" in the dock's self-update banner. Reproducible on at least the **v2.1.1 → v2.1.2** transition path (project: "Project Boost"; user reported "still getting a crash when I go from 2.1.1 to 2.1.2"). Two crash reports captured back-to-back at 22:31 and 22:33 with the same signature.

This is **distinct from [#233](https://github.com/hi-godot/godot-ai/issues/233)** (Race A — worker spawned by new dock's `_build_ui` racing the bytecode swap) **and from the Race B path closed by [#236](https://github.com/hi-godot/godot-ai/pull/236)** (focus-in during install spawning a new worker). Workers are not involved here at all — all `WorkerThread 0..9` are idle in `cond_wait` at the moment of crash. The crash is on `com.apple.main-thread`.

## Repro

1. Project on v2.1.1 plugin, dock visible.
2. Wait for the auto-update banner to appear (dock detects v2.1.2 release on GitHub).
3. Click "Update".
4. Plugin downloads v2.1.2 zip ✓
5. Plugin extracts ✓
6. `fs.scan()` completes, `filesystem_changed` fires ✓
7. `_reload_after_update.call_deferred()` is queued ✓
8. `CallQueue::flush()` runs the deferred call → `set_plugin_enabled(false)` → plugin `_exit_tree` cascade → **CRASH**

Both crash reports captured: `~/Library/Logs/DiagnosticReports/Godot-2026-04-27-223158.ips` and `Godot-2026-04-27-223326.ips`. Same backtrace.

## Symbolicated backtrace (Godot 4.6.2-stable.mono.official)

```
Signal: SIGABRT (EXC_BAD_ACCESS, KERN_INVALID_ADDRESS at 0x38)
Thread: com.apple.main-thread

Frame 34: CallQueue::flush()                                       ← deferred dispatch
Frame 33: CallQueue::_call_function
Frame 32: Callable::callp
Frame 31: Object::callp
Frame 30: GDScriptInstance::callp
Frame 29: GDScriptFunction::call (+45384)                          ← _reload_after_update
Frame 28: <trampoline>
Frame 27: EditorNode::set_addon_plugin_enabled                     ← line: set_plugin_enabled(false)
Frame 26: EditorNode::remove_editor_plugin
Frame 25: Node::remove_child
Frame 24: Node::_propagate_exit_tree                               ← plugin._exit_tree fires
Frame 23: GDScriptInstance::callp                                  ← plugin.gd::_exit_tree (GDScript)
Frame 22: GDScriptFunction::call (+45384)
Frame 21: <trampoline>
Frame 20: EditorPlugin::remove_control_from_docks                  ← plugin.gd:275 — remove_control_from_docks(_dock)
Frame 19: EditorDockManager::remove_dock
Frame 18: EditorDockManager::_move_dock
Frame 17: Node::remove_child
Frame 16: Node::_propagate_exit_tree (+100)                        ← dock-side children iteration
Frame 15: Node::_propagate_exit_tree (+196)                        ← a child's own notify-self
Frame 14: GDScriptInstance::callp                                  ← child's _exit_tree (GDScript)
Frame 13: GDScriptFunction::call (+42792)
Frame 12: Variant::callp                                           ← child._exit_tree calls method via typed-Variant
Frame 11: Object::callp
Frame 10: GDScriptInstance::callp                                  ← second GDScript dispatched
Frame 9: GDScriptFunction::call (+33476)
Frame 8: <trampoline>
Frame 7: Dictionary::keys() const                                  ← CRASH (Dictionary's internal pointer is null)
```

The KERN_INVALID_ADDRESS at 0x38 is a small offset from null — classic struct-pointer dereference on a freed object. `Dictionary::keys()` walks the HashMap's internal storage; if that pointer is null, the call lands at `null+(offset of internal field)`.

## Hypothesis: teardown ordering in `plugin.gd::_exit_tree`

v2.1.1 (and unchanged in v2.1.2) does:

```gdscript
func _exit_tree() -> void:
    if _connection: _connection.teardown()
    if _dispatcher: _dispatcher.clear()
    _handlers.clear()                              # ← line 273: handlers freed here
    if _dock:
        remove_control_from_docks(_dock)            # ← line 275: dock removed AFTER handlers are gone
        _dock.queue_free()
    ...
```

The comment on this function explicitly references [#46](https://github.com/hi-godot/godot-ai/issues/46) and says the order is intentional — clear `_dispatcher` and `_handlers` early so the RefCounted handler destructors run *while their class_name scripts are still loaded*. **But this means: by the time `remove_control_from_docks` fires the dock's children's `_exit_tree` cascade (frame 20→14), `_handlers` has already been freed.** If anything in the dock or its children holds a `Callable` bound to a method on one of those handlers — or to a Dictionary owned by one of those handlers — calling it now lands in invalid memory.

The frame-12 `Variant::callp` followed by frame-11 `Object::callp` is the GDScript pattern for "call a method on a typed Variant" — i.e. `var x: SomeType = something; x.method(...)`. If the typed Variant holds a `Callable` or `Object` that was indirectly anchored to a freed handler, this is exactly how it'd manifest.

## Open question: which child of the dock has the GDScript `_exit_tree`?

Frame 14 is a GDScript `_exit_tree` on a *child* of the dock (frames 17→16→15 walk dock removal → children iteration → notify-self). I checked every `add_child()` call in v2.1.1's `mcp_dock.gd` and **all explicit children are stock Godot built-ins** without custom scripts: `Label`, `Button`, `RichTextLabel`, `Polygon2D`, `HTTPRequest`, `PanelContainer`, `HBoxContainer`, `HSeparator`. None should have a GDScript `_exit_tree`.

Three possibilities:
1. A child added via `setup()` or another path I missed.
2. Godot's `EditorDockManager` wraps the dock in a custom Control internally (the C++ `EditorDock` wrapper at frames 18/19 might cascade through such a wrapper).
3. One of the "stock" children has been script-attached at runtime in a way I missed when grepping.

## What this is NOT

- **NOT** Race A ([#233](https://github.com/hi-godot/godot-ai/issues/233)) — that's the new dock's `_build_ui` worker, on a *worker thread*, racing bytecode swap. Different code path entirely.
- **NOT** Race B (closed by [#236](https://github.com/hi-godot/godot-ai/pull/236)) — focus-in during install spawning a new worker. Workers are idle here.
- **NOT** the CLI-probe deadlock ([#238](https://github.com/hi-godot/godot-ai/issues/238) / [#239](https://github.com/hi-godot/godot-ai/issues/239)) — those are sync `OS.execute` hangs, not crashes.

## Suggested next steps

1. **Reproduce locally** with v2.1.1 base + click Update → v2.1.2. The dock's existing GitHub release check should surface the v2.1.2 banner without any synthetic setup.
2. **Identify frame 14 / the dict source.** Two cheap probes that would crack this open:
   - Instrument `mcp_dock.gd::_exit_tree` to walk `get_children()` recursively and print each `node.name + node.get_class() + (node.get_script() if it has one)` *before* the drain runs. Will tell us exactly which child node is involved.
   - Instrument every `_handlers` element to add a `_notification(NOTIFICATION_PREDELETE)` print at the top, so we know the order in which handlers are destroyed vs the dock teardown.
3. **Test the fix hypothesis.** Reorder `_exit_tree` so `remove_control_from_docks(_dock)` fires *before* `_handlers.clear()`. If that closes this crash without re-opening [#46](https://github.com/hi-godot/godot-ai/issues/46) (`GameLogBuffer._storage` SIGSEGV), the teardown ordering is the root cause.

## Severity / scope

- Affects **everyone** upgrading from v2.1.1 → v2.1.2 via the dock's "Update" button. Rough fraction of users on v2.1.1 isn't known but is non-trivial — that's the most recent stable release before v2.1.2.
- The same teardown order is in v2.1.2, so **future v2.1.2 → v2.1.3 upgrades will also hit this** unless fixed.
- Workaround for affected users: skip the in-editor self-update and reinstall the plugin manually (download zip from Releases, extract over `addons/godot_ai/`).

dsarno:
## Summary

Closes [#242](https://github.com/hi-godot/godot-ai/issues/242).

The `v2.1.1 → v2.1.2` in-place self-update SIGABRT'd in `_propagate_exit_tree` → `Dictionary::keys()` for users in the wild (Project Boost). Root cause: v2.1.2's `plugin.gd:42` declared `var _editor_log_buffer: EditorLogBuffer` — a typed-var against a brand-new `class_name` introduced by [#240](https://github.com/hi-godot/godot-ai/pull/240). During `_install_update`'s extract → `fs.scan()` cycle, the parser hits that typed-var BEFORE `EditorLogBuffer`'s class_name has been registered in the global table → `plugin.gd` parse fails → v2.1.x plugin enters a degraded state → the follow-up `set_plugin_enabled(false)` fires its `_exit_tree` cascade against that degraded instance → crash downstream.

`GameLogBuffer`'s `extends` chain also changed in v2.1.2 to inherit from `StructuredLogRing` (also new), giving the four-error chain we observed in the repro:

```
SCRIPT ERROR: Could not parse global class "GameLogBuffer"
SCRIPT ERROR: Could not find type "EditorLogBuffer" in the current scope (plugin.gd:42)
SCRIPT ERROR: Could not resolve class "GameLogBuffer", because of a parser error
SCRIPT ERROR: Identifier "EditorLogBuffer" not declared in the current scope (plugin.gd:92)
```

### Fix 1 — untype `_game_log_buffer` and `_editor_log_buffer` in plugin.gd

Same lesson as the existing untyped `_editor_logger` field: typed declarations in `plugin.gd` against class_names whose inheritance or new class_name siblings can change between releases are self-update hazards. Untyped + `preload(...).new()` resolves at script-load time without needing the global class_name registry to be ahead of `plugin.gd`'s parse.

The runtime parameter checks on `EditorHandler._init`'s typed `editor_log_buffer: EditorLogBuffer` parameter still enforce the type fence — we just move it from `plugin.gd`'s parse to the call site's runtime check, breaking the chicken-and-egg.

### Fix 2 — parse-check fallback in `_on_filesystem_scanned_for_update`

Belt-and-suspenders for the next time someone introduces a parse hazard we don't catch in review. Before triggering `_reload_after_update`, load the new `plugin.gd` via `ResourceLoader.load`. `null` return = parse error → bail out to the same "Restart editor to apply" surface the pre-Godot-4.4 path already uses. User manually restarts, picks up the new files fresh from disk, dodges the in-place reload entirely.

## Tests

Two new structural tests in `tests/unit/test_editor_focus_refocus.py`:

- `test_plugin_gd_avoids_typed_class_name_for_log_buffers` — locks the untyped+preload pattern (so a future contributor "fixing" the apparent oversight by re-typing the field will fail CI).
- `test_install_update_falls_back_on_plugin_parse_failure` — locks the parse-check + bail-out: `ResourceLoader.load` gate, install flag cleared, "Restart editor" message, AND the parse check happens *before* `_reload_after_update.call_deferred()`.

All 653 Python tests pass locally. ruff clean.

## What this does NOT close

- **Users currently on v2.1.1**: this fix lands in v2.1.3+ and protects v2.1.x → v2.1.x+1 transitions where v2.1.x has the fix. Users on v2.1.1 hitting #242 still need to manually reinstall (download zip from Releases, extract over `addons/godot_ai/`, restart editor).
- **Direct identification of the crashing dock child** in #242's `Dictionary::keys()` SIGABRT: the crash is downstream of the parse failure plus Project-Boost-specific dock state. Eliminating the parse failure eliminates the conditions for the crash, so it's fixed transitively. The exact dock child remains an open thread on #242 if some other crash signature later needs it.
- **Broader untyping of every class_name field in plugin.gd** (defense-in-depth against future regressions): filed as a separate follow-up for evaluation.

## Test plan

- [x] `pytest` — 653 passed locally
- [x] `ruff check plugin/ tests/` — clean
- [ ] CI passes (will populate)
- [ ] Live smoke verifying the parse-check fallback fires when a hand-crafted "broken" plugin.gd is dropped in (next session — not blocking merge given structural tests cover the path)

### DIFF
--- a/plugin/addons/godot_ai/mcp_dock.gd
+++ b/plugin/addons/godot_ai/mcp_dock.gd
@@ -2168,6 +2168,34 @@ func _install_update() -> void:
 
 
 func _on_filesystem_scanned_for_update() -> void:
+	## Belt-and-suspenders: verify the new release's plugin.gd actually parses
+	## before we fire `set_plugin_enabled(false/true)`. If the new release adds
+	## a class_name'd file that's referenced by a typed-var declaration in
+	## plugin.gd (or its transitive dependencies), GDScript's parser can fail
+	## to resolve the new class_name during hot-reload — `plugin.gd` then
+	## fails to compile and the v2.1.x plugin enters a degraded state. The
+	## subsequent `_exit_tree` cascade has been observed to SIGABRT in the
+	## wild (#242).
+	##
+	## `ResourceLoader.load` returns `null` when a script has a parse error,
+	## so a single load is enough to detect the hazard. On failure, fall back
+	## to the same "Restart editor to apply" surface the pre-Godot-4.4 path
+	## already uses — the user manually restarts, picks up the new files
+	## fresh from disk, and dodges the in-place reload entirely.
+	##
+	## The check itself is cheap (the script's already in the editor's cache
+	## post-`fs.scan()`) and only runs once per update install.
+	if ResourceLoader.load("res://addons/godot_ai/plugin.gd") == null:
+		_self_update_in_progress = false
+		_update_btn.text = "Restart editor to apply"
+		_update_btn.disabled = true
+		_update_label.text = (
+			"Updated! Restart the editor to apply — auto-reload skipped because "
+			"the new release added a script class plugin.gd parses against "
+			"(safe-mode fallback, see #242)."
+		)
+		_update_label.add_theme_color_override("font_color", Color.GREEN)
+		return
 	_update_btn.text = "Reloading..."
 	_reload_after_update.call_deferred()
 
--- a/plugin/addons/godot_ai/mcp_dock.gd
+++ b/plugin/addons/godot_ai/mcp_dock.gd
@@ -2189,11 +2189,7 @@ func _on_filesystem_scanned_for_update() -> void:
 		_self_update_in_progress = false
 		_update_btn.text = "Restart editor to apply"
 		_update_btn.disabled = true
-		_update_label.text = (
-			"Updated! Restart the editor to apply — auto-reload skipped because "
-			"the new release added a script class plugin.gd parses against "
-			"(safe-mode fallback, see #242)."
-		)
+		_update_label.text = "Updated! Restart the editor to apply — auto-reload skipped (safe-mode fallback, see #242)."
 		_update_label.add_theme_color_override("font_color", Color.GREEN)
 		return
 	_update_btn.text = "Reloading..."
--- a/plugin/addons/godot_ai/plugin.gd
+++ b/plugin/addons/godot_ai/plugin.gd
@@ -38,8 +38,22 @@ const SPAWN_GRACE_MS := 5 * 1000
 var _connection: Connection
 var _dispatcher: McpDispatcher
 var _log_buffer: McpLogBuffer
-var _game_log_buffer: GameLogBuffer
-var _editor_log_buffer: EditorLogBuffer
+## Untyped — `GameLogBuffer` and `EditorLogBuffer` are class_names whose
+## inheritance chain has changed between releases (v2.1.2 refactored both to
+## extend `StructuredLogRing`, a class_name that didn't exist in v2.1.1).
+## A typed declaration here is a self-update hazard: when `_install_update`
+## extracts the new release's files and `fs.scan()` triggers script reload,
+## the parser hits these typed-vars before the new dependent class_names
+## are registered in the global table → plugin.gd parse error → v2.1.x
+## plugin fails to load → set_plugin_enabled(false)'s _exit_tree cascade
+## fires on a degraded dock instance → SIGABRT. See #242.
+##
+## Same lesson as the existing `_editor_logger` field below: untyped + load
+## via `preload()` at construction time. The `preload()` is resolved at
+## script-load and doesn't need the global class_name registry to be ahead
+## of plugin.gd's parse, breaking the chicken-and-egg.
+var _game_log_buffer
+var _editor_log_buffer
 ## Untyped — script extends Godot 4.5+'s Logger class, loaded via load() so
 ## the plugin still parses on 4.4. Null on Godot < 4.5 or before
 ## `_attach_editor_logger` runs; "attached" state IS exactly "non-null".
@@ -88,8 +102,10 @@ func _enter_tree() -> void:
 	_start_server()
 
 	_log_buffer = McpLogBuffer.new()
-	_game_log_buffer = GameLogBuffer.new()
-	_editor_log_buffer = EditorLogBuffer.new()
+	## See the `_game_log_buffer` / `_editor_log_buffer` declaration comment
+	## above for why these go through `preload()` instead of `ClassName.new()`.
+	_game_log_buffer = preload("res://addons/godot_ai/utils/game_log_buffer.gd").new()
+	_editor_log_buffer = preload("res://addons/godot_ai/utils/editor_log_buffer.gd").new()
 	_attach_editor_logger()
 	_dispatcher = McpDispatcher.new(_log_buffer)
 
--- a/tests/unit/test_editor_focus_refocus.py
+++ b/tests/unit/test_editor_focus_refocus.py
@@ -209,6 +209,112 @@ def test_install_update_drains_workers_and_blocks_spawning_before_extract() -> N
     )
 
 
+def test_plugin_gd_avoids_typed_class_name_for_log_buffers() -> None:
+    """plugin.gd's `_game_log_buffer` / `_editor_log_buffer` must be untyped (#242).
+
+    Hot-reload during self-update parses the new release's plugin.gd before
+    its dependent class_name'd files are guaranteed to be in the global class
+    table. A typed declaration like `var _editor_log_buffer: EditorLogBuffer`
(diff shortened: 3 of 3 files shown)


dsarno:
Live smoke test ran: launched a Godot editor against this branch's `test_project/`, plugin loaded cleanly (`plugin=2.1.2`, session `test-project@0464`), zero parse errors in the editor log, scene goes `importing → ready` normally, no diagnostic reports. Validates the runtime behavior of the untyped+preload change.

What this didn't cover: the parse-check fallback path itself (would need a synthetically-broken plugin.gd + triggering `_install_update`'s download flow against a fake newer release). The structural test `test_install_update_falls_back_on_plugin_parse_failure` locks the code shape, and the path is small and straightforward — accepting that as adequate without live exercise. Flagged as a follow-up smoke item if it bites later.

dsarno:
Closing as superseded by #250. The investigation here was useful, but this PR depends on a restart-required fallback and narrower typed-field mitigation, which no longer matches the #245 constraint: the editor process must stay alive and future typed Dict/Array additions must not reintroduce the crash.

PR #248, opened 2026-04-28
Avoid Windows-reserved WebSocket ports

thekeunpie-hash:
## Summary
- Detect when the configured Godot plugin WebSocket port is inside a Windows excluded TCP port range.
- Fall back to the next non-excluded port instead of repeatedly trying a port that fails with WinError 10013.
- Recompute the WebSocket URL on every reconnect attempt so a changed/fallback `ws_port` is picked up without keeping a stale URL.
- Add parser-backed tests for skipping adjacent Windows reserved ranges.

## Context
On Windows, Hyper-V / WSL2 / Docker can reserve TCP ranges via winnat. If `godot_ai/ws_port` falls inside one of those ranges, `websockets.serve()` fails with WinError 10013 even though no process owns the port. The HTTP server can still be running, leaving the dock stuck in a reconnect loop.

Observed locally:
- `godot_ai/ws_port = 9500`
- Windows excluded range included `9491-9590`
- Python could not bind `127.0.0.1:9500`
- Moving WS to `10500` allowed the plugin to connect

## Testing
- `git diff --check`
- Godot `--check-only --script` against the modified GDScript files showed no parse errors.
- `uv run --extra dev pytest tests/unit -q` was attempted, but failed due existing Windows locale behavior: 17 tests hit `UnicodeDecodeError` reading UTF-8 GDScript files through default cp949; 388 tests passed before those unrelated failures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,54 @@ The plugin starts or reuses the Python server, connects over WebSocket, and expo
 </details>
 
 <details>
+<summary><strong>Windows: <code>uvx mcp-proxy</code> won't start (<code>pywin32</code> install fails)</strong></summary>
+
+Symptom (in your MCP client's server log):
+
+```text
+error: Failed to install: pywin32-311-cp313-cp313-win_amd64.whl (pywin32==311)
+  Caused by: failed to remove directory `C:\Users\<you>\AppData\Local\uv\cache\builds-v0\.tmpXXXXXX\Lib\site-packages\pywin32-311.data`: ... os error 32
+```
+
+Cause: uv hard-links shared `.pyd` files (notably
+`pydantic_core/_pydantic_core.cp313-win_amd64.pyd`) from `archive-v0\` into
+each new `builds-v0\.tmpXXXXXX\` build venv. The running `godot-ai` Python
+process has the same `.pyd` mapped via `LoadLibrary` — and because hard
+links share the inode, Windows refuses to delete it under any path until
+every process unmaps it. uv's post-install cleanup of the build venv then
+dies on a stale lock; the misleading `pywin32` mention is just the last
+package in the resolution order, not the actual lock holder.
+
+**Mitigation in this plugin:** `_stop_server` and `force_restart_server`
+both call `McpUvCacheCleanup.purge_stale_builds()` immediately after
+killing the server children, while the `.pyd` is briefly unmapped. See
+[`plugin/addons/godot_ai/utils/uv_cache_cleanup.gd`](plugin/addons/godot_ai/utils/uv_cache_cleanup.gd).
+
+**Recommended belt-and-braces:** tell uv to copy instead of hard-link by
+setting `UV_LINK_MODE=copy` in the MCP launcher's environment. This also
+removes the reverse race where an MCP client spawns `uvx mcp-proxy`
+*while* a server child is still holding the `.pyd`. Example for Claude
+Desktop's `claude_desktop_config.json`:
+
+```json
+{
+  "mcpServers": {
+    "godot-ai": {
+      "command": "uvx",
+      "args": ["mcp-proxy", "--transport", "streamablehttp", "http://127.0.0.1:8000/mcp"],
+      "env": { "UV_LINK_MODE": "copy" }
+    }
+  }
+}
+```
+
+If you've already hit the lock, kill stray `python.exe` children whose
+command line contains `spawn_main(parent_pid=...)` and delete
+`%LOCALAPPDATA%\uv\cache\builds-v0\.tmp*` manually before retrying.
+
+</details>
+
+<details>
 <summary><strong>Contributing</strong></summary>
 
 See [CONTRIBUTING.md](docs/CONTRIBUTING.md) for development setup, testing, and PR guidelines.
--- a/plugin/addons/godot_ai/client_configurator.gd
+++ b/plugin/addons/godot_ai/client_configurator.gd
@@ -46,7 +46,10 @@ static func http_port() -> int:
 
 ## Active WebSocket port: user override (if in range) or `DEFAULT_WS_PORT`.
 static func ws_port() -> int:
-	return _read_port_setting(SETTING_WS_PORT, DEFAULT_WS_PORT)
+	var port := _read_port_setting(SETTING_WS_PORT, DEFAULT_WS_PORT)
+	if McpWindowsPortReservation.is_port_excluded(port):
+		return suggest_free_port(port + 1, 2048)
+	return port
 
 
 static func http_url() -> String:
@@ -111,15 +114,9 @@ static func excluded_domains() -> String:
 ## currently excluded by Windows' winnat reservation table. Falls back to
 ## `start` if nothing clears (caller can apply anyway — user may just
 ## retry). On non-Windows this is a no-op: all ports pass, returns `start`.
-static func suggest_free_port(start: int, span: int = 100) -> int:
+static func suggest_free_port(start: int, span: int = 2048) -> int:
 	var candidate := clampi(start, MIN_PORT, MAX_PORT - span + 1)
-	for i in span:
-		var p := candidate + i
-		if p > MAX_PORT:
-			break
-		if not McpWindowsPortReservation.is_port_excluded(p):
-			return p
-	return candidate
+	return McpWindowsPortReservation.suggest_non_excluded_port(candidate, span, MAX_PORT)
 
 
 # --- Client operations (string id) ---------------------------------------
--- a/plugin/addons/godot_ai/client_configurator.gd
+++ b/plugin/addons/godot_ai/client_configurator.gd
@@ -46,10 +46,7 @@ static func http_port() -> int:
 
 ## Active WebSocket port: user override (if in range) or `DEFAULT_WS_PORT`.
 static func ws_port() -> int:
-	var port := _read_port_setting(SETTING_WS_PORT, DEFAULT_WS_PORT)
-	if McpWindowsPortReservation.is_port_excluded(port):
-		return suggest_free_port(port + 1, 2048)
-	return port
+	return _read_port_setting(SETTING_WS_PORT, DEFAULT_WS_PORT)
 
 
 static func http_url() -> String:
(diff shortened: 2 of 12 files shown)


dsarno (commented):
Thanks for the careful problem write-up — the WinError 10013 / winnat case is a real footgun and the diagnosis is good. A few correctness/safety concerns beyond what Copilot already flagged:

### 1. `ws_port()` is now non-idempotent across calls — spawn/connect can disagree

Before this PR, `ws_port()` was a pure read of an `EditorSetting`: stable across calls within a session. After this PR it shells out to `netsh` and can return a different value on each call.

The plugin reads `ws_port()` at several independent moments and assumes they agree:

- `plugin.gd:410` — passed to the spawned server as `--ws-port`
- `plugin.gd:661` — same, on `--refresh` respawn
- `plugin.gd:1070` — same, for the dev-server launcher
- `connection.gd:38` (`_ready`) — initial WS URL
- `connection.gd:116` (new in this PR) — re-resolved on every reconnect
- `mcp_dock.gd:870` — server-label display

If the netsh excluded-range table changes between two of these calls (winnat restart, Hyper-V toggle, transient `netsh` failure), the server binds one port and the dock dials another, producing a silent "reconnect forever" exactly like the bug this PR is trying to fix. `is_port_excluded` returns `false` on netsh failure, so a transient subprocess error after a successful spawn-time check inverts the decision.

Recommendation: resolve the swap once at spawn time, cache the result on the plugin, and have all readers (including `connection.gd`) read the cached value. The reconnect-time recompute can stay, but it should consult the cache, not re-run netsh.

### 2. Main-thread netsh from the frame loop

`Connection._process()` calls `_attempt_reconnect()` → `_connect_to_server()` → `ws_port()` → `OS.execute("netsh", …, blocking=true)`. That's a synchronous subprocess on the editor's main thread, on every reconnect attempt, during a reconnect storm (delays start at 1s). With Copilot's "doubled netsh" point still standing, that's up to two ~50–100ms subprocess invocations per reconnect tick. Cache-on-spawn (above) makes this go away.

### 3. Dead assignment in `connection.gd:_ready`

`_ready()` line 38 still sets `_url = "ws://…%d" % ws_port()`, but `_connect_to_server()` (called on the next line) now overwrites it unconditionally. That's a redundant netsh call at startup and a dead store. Either drop the `_ready` assignment or drop the new one in `_connect_to_server` and route the reconnect refresh through a different seam.

### 4. UX asymmetry between HTTP and WS reservation

When the HTTP port is reserved, `plugin.gd:475` surfaces `PORT_EXCLUDED` state and the dock shows the port-picker UI. When the WS port is reserved, this PR silently swaps to a different port. A user who set `godot_ai/ws_port = 9500` in EditorSettings sees the dock label flip to `WS: 10500` with no other indication of why. Worth either (a) routing WS through the same `PORT_EXCLUDED`/picker flow for symmetry, or (b) at minimum logging once via `log_buffer` that the configured WS port was unavailable and was swapped to N.

### 5. Test gap: no end-to-end coverage of the `ws_port()` swap

The new `suggest_non_excluded_port_from_output` parser tests are nice, but no test exercises the actual `ws_port()` decision (configured-and-reserved → fallback returned). Hard to write portably because it's gated on `OS.get_name() == "Windows"`, but a refactor that splits `ws_port()` into a pure "given netsh output, return port" helper plus a thin Windows-only shell would make this testable on every platform — and would also dovetail with the cache-on-spawn refactor in #1.

### What looks right

- Re-resolving the URL in `_connect_to_server` is the correct shape for "EditorSetting changed at runtime"; the implementation just needs to read from a cached resolved port instead of re-running netsh.
- `suggest_non_excluded_port_from_output` is a clean parser-backed seam, well tested.
- Widening `span` to 2048 is reasonable for the adjacent-Hyper-V-ranges case the PR describes.
- No new attack surface — `OS.execute("netsh", …)` was already used by existing code.

Net: design intent is right, but the call-site fanout for `ws_port()` makes the "consult netsh on every read" implementation racy and main-thread-heavy. Caching the resolved port at spawn time addresses #1 and #2 together and makes #5 trivial.

---

dsarno:
@thekeunpie-hash Thanks for the submission! If you can address the feedback I'll merge this.

thekeunpie-hash:
Addressed the review feedback in `cbd02e2`:

- `ws_port()` is back to a pure EditorSettings read.
- The resolved WS port is cached on the plugin/server lifecycle and reused by spawn, refresh respawn, reconnect, dev-server launch, and the dock label.
- Managed server records now persist the resolved WS port so same-version adoption does not re-run `netsh` and drift from the bound server port.
- `Connection` no longer calls the configurator or shells out during reconnects.
- WS fallback logs a one-time dock-visible message when the configured port is swapped.
- `suggest_non_excluded_port_from_output()` parses excluded ranges once and jumps across reserved ranges instead of re-scanning output per candidate.
- Added parser-backed tests for the pure WS resolution path.

Validation run locally:
- `git diff --check`
- Godot `--check-only` for `test_windows_port_reservation.gd`
- Godot `--check-only` for `test_plugin.gd`

thekeunpie-hash:
Rebased the branch onto `v2.2.1` / current `main` and force-pushed the PR branch.

Follow-up changes in `edf98d0` are now compatible with the 2.2.1 class-name scoping work:

- Uses `McpConnection`, `McpLogBuffer`, and `McpWindowsPortReservation` names.
- Keeps the 2.2.1 live-server status compatibility/adoption checks intact.
- Caches the resolved WS port on the plugin lifecycle and persists it in the managed-server record.
- Uses the cached resolved WS port for spawn, refresh respawn, reconnect, dev server launch, dock label, and incompatible-server messaging.
- Leaves `McpClientConfigurator.ws_port()` as a pure EditorSettings read.

Validation run locally after cache refresh:
- `git diff --check`
- Godot `--check-only` for `test_windows_port_reservation.gd`
- Godot `--check-only` for `test_plugin.gd`

dsarno:
Thanks again for iterating on this. The earlier review items around repeated `netsh`, reconnect-time port drift, dead URL assignment, and dock-visible logging look addressed.

One new concern came up only because today’s server-adoption safety work landed after this branch was last based/reviewed. PR #259 changed the ownership model so a compatible live server with a stale/different managed record is treated as external and the stale record is cleared, instead of letting old metadata authorize killing an unrelated process.

Could you please rebase this on current `main` and make one adjustment to preserve that behavior?

In the merged shape, `_start_server()` reads `record["ws_port"]` and uses it as the expected WS port before checking whether `record_version` matches the current plugin version. That means a stale managed record can still influence the compatibility check:

- stale record: old version, old resolved WS port, e.g. `9500`
- live compatible current server: same HTTP port, current version, actual WS port, e.g. `10500`
- plugin expects the stale `9500`, reports WS mismatch, then the version-drift branch can still authorize killing the port owner

That cuts against the stale-foreign-server protection that #259 just merged.

Requested fix: only trust/persisted `managed_server_ws_port` when the managed record is current ownership proof, e.g. `record_version == current_version`. If the record version is missing or stale, resolve the expected WS port from current settings / Windows reservation logic before probing compatibility, then let the compatible-external adoption path clear the stale record.

A focused regression test for “stale record has stale ws_port, live current server reports a different current ws_port, adoption remains external and clears the record” would make this safe to merge.

This is not a new design direction for #248; it’s just aligning the resolved-WS-port cache with the ownership rules that landed today in #259.

thekeunpie-hash:
Rebased onto a2fafb4 (v2.2.2) and force-pushed.

Ownership-proof gate — squashed into the originating "Cache resolved WebSocket port" commit so the cached-ws-port logic obeys the gate from inception in branch history. New static helper _resolved_ws_port_for_existing_server discards the cached ws_port unless record_version == current_version; otherwise it returns the freshly resolved value. A stale record's port is dropped before the compatibility probe, the live current-version server passes compatibility, adoption flows through the external path, and the stale record is cleared.

Regression coverage in test_plugin_lifecycle.gd:

test_resolved_ws_port_drops_stale_record_value — five helper cases (stale, matching, missing cache, empty record, empty current)
test_stale_ws_port_does_not_authorize_killing_external_server — stale 9500 / live 10500 / record 2.1.0 vs current 2.2.0 → external adoption, _server_pid == -1, stale record cleared, no restart authority
teardown also clears MANAGED_SERVER_WS_PORT_SETTING.

One extra commit — Sweep stale uvx build venvs after server stop. Windows-only: uv hard-links pydantic_core.pyd from archive-v0\ into each builds-v0\.tmpXXX\ build venv, and the running server child holds that .pyd mapped, so uvx mcp-proxy post-install cleanup dies on the lock. New McpUvCacheCleanup sweeps .tmp* orphans on _stop_server / force_restart_server. README adds a recommendation to set UV_LINK_MODE=copy. Happy to split it into a separate PR if you'd prefer.

Heads-up: heading back to game work for a stretch, so I'll be running this branch as-is and pulling updates rather than iterating much in the near term. Will respond promptly if review surfaces anything blocking.

dsarno:
> Rebased onto [a2fafb4](https://github.com/hi-godot/godot-ai/commit/a2fafb405011885e0aa38e48f28053e4b908672e) (v2.2.2) and force-pushed.
> 
> Ownership-proof gate — squashed into the originating "Cache resolved WebSocket port" commit so the cached-ws-port logic obeys the gate from inception in branch history. New static helper _resolved_ws_port_for_existing_server discards the cached ws_port unless record_version == current_version; otherwise it returns the freshly resolved value. A stale record's port is dropped before the compatibility probe, the live current-version server passes compatibility, adoption flows through the external path, and the stale record is cleared.
> 
> Regression coverage in test_plugin_lifecycle.gd:
> 
> test_resolved_ws_port_drops_stale_record_value — five helper cases (stale, matching, missing cache, empty record, empty current) test_stale_ws_port_does_not_authorize_killing_external_server — stale 9500 / live 10500 / record 2.1.0 vs current 2.2.0 → external adoption, _server_pid == -1, stale record cleared, no restart authority teardown also clears MANAGED_SERVER_WS_PORT_SETTING.
> 
> One extra commit — Sweep stale uvx build venvs after server stop. Windows-only: uv hard-links pydantic_core.pyd from archive-v0\ into each builds-v0.tmpXXX\ build venv, and the running server child holds that .pyd mapped, so uvx mcp-proxy post-install cleanup dies on the lock. New McpUvCacheCleanup sweeps .tmp* orphans on _stop_server / force_restart_server. README adds a recommendation to set UV_LINK_MODE=copy. Happy to split it into a separate PR if you'd prefer.
> 
> Heads-up: heading back to game work for a stretch, so I'll be running this branch as-is and pulling updates rather than iterating much in the near term. Will respond promptly if review surfaces anything blocking.

Let me know if you can fix the failed CI tests.  When it passes CI I'll merge.  Thanks for the work.

dsarno:
Thanks @thekeunpie-hash !  Welcome to the club!

Issue #247, opened 2026-04-28
v2.1.1 → v2.1.2 self-update half-fails: EditorHandler.new(5 args) hits cached 4-arg _init at plugin.gd:103 (not caught by #243)


dsarno:
## Summary

Reproducible on Linux/non-mono and Linux/mono, Godot 4.6.2-stable, with a fresh empty project: the **v2.1.1 → v2.1.2 in-place self-update flow fails during `_reload_after_update`**'s re-enable step. v2.1.2's `plugin.gd::_enter_tree` calls `EditorHandler.new(_log_buffer, _connection, _debugger_plugin, _game_log_buffer, _editor_log_buffer)` with 5 args, but the cached `EditorHandler` class is still v2.1.1's 4-arg `_init`. Result:

```
ERROR: Error constructing a GDScriptInstance: 'RefCounted(editor_handler.gd)::_init':
       Method expected 4 argument(s), but called with 5
   at: _create_instance (modules/gdscript/gdscript.cpp:200)
   GDScript backtrace (most recent call first):
       [0] _enter_tree (res://addons/godot_ai/plugin.gd:103)
       [1] _reload_after_update (res://addons/godot_ai/mcp_dock.gd:1932)
SCRIPT ERROR: Invalid call to function 'new' in base 'GDScript'. Expected 4 argument(s).
```

Plugin enters a half-loaded state (no handlers wired, no dock UI, server still running). Distinct from #242 (which is a SIGABRT with a different signature). Distinct from the parse-failure path closed by #243.

## Reproduction (deterministic, Linux/non-mono)

```bash
mkdir -p /tmp/repro && cd /tmp/repro
curl -L -o v211.zip https://github.com/hi-godot/godot-ai/releases/download/v2.1.1/godot-ai-plugin.zip
curl -L -o v212.zip https://github.com/hi-godot/godot-ai/releases/download/v2.1.2/godot-ai-plugin.zip

mkdir proj && cd proj
cat > project.godot <<'EOF'
config_version=5
[application]
config/name="repro"
config/features=PackedStringArray("4.6")
[autoload]
_mcp_game_helper="*res://addons/godot_ai/runtime/game_helper.gd"
[editor_plugins]
enabled=PackedStringArray("res://addons/godot_ai/plugin.cfg")
EOF
unzip -q ../v211.zip

# Stage v2.1.2 zip where _install_update reads it
USER_DIR="$HOME/.local/share/godot/app_userdata/repro"
[...]

PR #249, opened 2026-04-28
fix(dock): pre-detach dock from self-update to dodge hot-reload Dict-NIL crash (#245)

### LINKED ISSUE #245: Identify the "crashing dock child" GDScript _exit_tree from #242 (forensic followup)
## Summary

Investigative follow-up to [#242](https://github.com/hi-godot/godot-ai/issues/242) / [#243](https://github.com/hi-godot/godot-ai/pull/243).

[#243](https://github.com/hi-godot/godot-ai/pull/243) closes the upstream parse-failure hazard that produced the SIGABRT in #242. But the *immediate* crash signature — `Dictionary::keys()` at `null+0x38`, fired by a GDScript `_exit_tree` on a **child of the dock** during `_propagate_exit_tree` cascade — remains unidentified.

The investigation in #242 walked every `add_child(...)` call in `mcp_dock.gd` and confirmed every explicit child is a stock Godot built-in (Label, Button, RichTextLabel, HTTPRequest, PanelContainer, HBoxContainer, etc.) without custom GDScript attached. A live `editor_reload_plugin` against a fresh test project printed the dock subtree top-to-bottom — every node showed `script=(no script)`. So under "fresh test project" conditions there *is* no scripted child, and the crash doesn't reproduce.

But the crash *did* fire in Project Boost. Something there has a custom-scripted dock descendant we didn't account for. Possibilities:

- **A dock child added by `setup()` or another path the grep missed.** Lower-likelihood — `setup()` was inspected and didn't reveal anything custom — but worth a second pass with fresh eyes.
- **Godot's editor `EditorDockManager` wraps the dock control internally.** Frame 18 in the crash is `EditorDockManager::_move_dock`. The C++ wrapper itself wouldn't have a GDScript `_exit_tree`, but it might own a Control-with-script that we don't see from our side.
- **Project-Boost-specific runtime state**: another EditorPlugin, an autoload that reaches into the dock's tree, or some script-attach the user did at runtime. Project Boost is a real game project; its editor session has more state than the empty `/tmp` test.
- **One of our "stock" children has a script attached at runtime by something we don't track.** Grep at the `add_child` site shows stock — but `set_script(...)` could happen later. Worth checking.

## Why this matters even after [#243](https://github.com/hi-godot/godot-ai/pull/243)

[#243](https://github.com/hi-godot/godot-ai/pull/243) eliminates the parse-failure path that *led to* this crash, and adds a parse-check fallback so similar future hazards bail out before the `set_plugin_enabled(false)` `_exit_tree` cascade fires. So the most likely production trigger is closed.

But the underlying child-of-dock teardown bug is still present, just not reachable via the parse-failure path. If a different scenario produces a similar "_exit_tree on a dock child holding a stale Dictionary reference" condition, this same crash signature will fire — and we'd be back at the same forensics with the same blind spot.

## Suggested investigation paths

Pick whichever's lowest-friction:

1. **Reproduce Project Boost-style.** Check Project Boost's `project.godot` for editor plugins or autoloads that interact with dock UI. If something obvious shows up, mimic that setup in a test project and re-run the dock-subtree walk from #242. If frame 14 has a custom-script child this time, we have the answer.
2. **Symbolicated debug build of Godot 4.6.2.** The crash backtrace we have includes two unsymbolicated frames (image+offset, no symbol). With a debug build's `dSYM`, `atos` would resolve those — likely revealing which built-in C++ wrapper is at frame 8, and whether frame 14's GDScript instance is on a node we *can* see in the subtree walk after all.
3. **Instrument plugin teardown to log every node-with-script in the editor tree at `_install_update` time.** Walk the entire editor scene tree (not just the dock subtree) and log script-attached nodes. The crashing child might be reachable from outside our `add_child` grep.
4. **Review the `EditorDock` wrapper internals** in Godot source. Frame 18 (`EditorDockManager::_move_dock`) is the closest C++ callsite to the crash. If `EditorDockManager` adds a wrapper Control between our dock and its children, that wrapper might be where the custom GDScript lives.

## Reproduction notes

Two crash reports captured: `~/Library/Logs/DiagnosticReports/Godot-2026-04-27-223158.ips` and `Godot-2026-04-27-223326.ips`. Both same backtrace, both fired by clicking "Update" on Project Boost's v2.1.1 → v2.1.2 transition. The full symbolicated stack is in [#242](https://github.com/hi-godot/godot-ai/issues/242).

Repro environment that would *not* trigger this crash (we tried): empty Godot project at `/tmp/`, v2.1.1 → v2.1.2 transition via cp + `editor_reload_plugin`. The parse failures fire identically, but the child-of-dock crash doesn't — the empty project doesn't have the dock-child state Project Boost has.

dsarno:
## Summary

- Pre-detaches the McpDock instance during `_install_update` so it never runs hot-reloaded `_exit_tree` bytecode against pre-update field storage. Closes [#245](https://github.com/hi-godot/godot-ai/issues/245) for v(N) → v(N+1) upgrades from this release forward.
- Adds a forensic synthetic repro (`docs/forensic/issue-245/repro-dictinject/`) and a one-shot CI workflow (`.github/workflows/forensic-issue-245.yml`) that reproduces the underlying Godot 4.6.2 hot-reload bug on Linux + macOS + Windows runners. CI run: [25061868538](https://github.com/hi-godot/godot-ai/actions/runs/25061868538).
- Filed upstream Godot bug at [dsarno/godot#1](https://github.com/dsarno/godot/issues/1).

## What this fixes

The original SIGABRT in [#242](https://github.com/hi-godot/godot-ai/issues/242) (Project Boost v2.1.1 → v2.1.2 self-update) traces to a Godot 4.6.2 script-hot-reload bug: when a script class hot-reloads on a live instance and the new version declares a new typed `Dictionary` field with `= {}`, the field's backing storage is left as `Variant::NIL` instead of running the declared initializer. The new bytecode then reads it as `Dictionary` (e.g. `.keys()`) and dereferences a null `_p` → SIGSEGV in `Dictionary::keys() const + 52`. **Confirmed cross-platform** on Godot 4.6.2-stable per the CI grid above (not macOS-only as the first forensic comment claimed; corrected on [#245](https://github.com/hi-godot/godot-ai/issues/245#issuecomment-4336713248)).

In v2.1.2 the trigger field was `var _client_action_threads: Dictionary = {}` newly introduced in `mcp_dock.gd`, read via `_drain_client_action_workers()` from `_exit_tree`. Any future plugin update that adds a new typed Dictionary or Array field used in `_exit_tree`'s reach has the same hazard.

## What this doesn't fix

- **v2.1.1 / v2.1.2 → v2.1.3 upgrades** still hit the original crash, because the running v2.1.1 / v2.1.2 `_install_update` doesn't have `detach_dock_for_update`. From v2.1.3 onwards the in-place upgrade path is safe. Affected users on v2.1.1 / v2.1.2 need to reinstall the plugin manually one time. Will document in the v2.1.3 release notes.
- [#247](https://github.com/hi-godot/godot-ai/issues/247) (`EditorHandler.new(5 args)` against cached 4-arg `_init` after hot-reload) is a different layer of the same upgrade window — script-class cache rather than instance-field cache. Not addressed here; tracked as a follow-up.

## How the fix works

Pre-update: while we still own bytecode that matches the dock's pre-update field shape, free the McpDock instance before triggering the filesystem rescan.

```
v(N) bytecode running:
   _install_update(): extract zip
   _install_update(): _plugin.detach_dock_for_update()
                       └─ remove_control_from_docks(_dock)  ← runs _exit_tree on
                          _dock.queue_free()                  v(N) bytecode against
                          _dock = null                        v(N) field shape ✓
                       └─ filesystem_changed.connect(...)
                       └─ fs.scan()
[hot-reload happens here — McpDock class swaps to v(N+1), but no instance
 to crash; plugin instance survives, but its _exit_tree reads only nullable
 refs guarded by `if _xxx:`, no Dictionary/Array fields read via builtin]
   _on_filesystem_scanned_post_extract():
     _reload_after_extract():
       set_plugin_enabled(false)  ← plugin._exit_tree: _dock is null, skips dock
       set_plugin_enabled(true)   ← plugin._enter_tree: builds fresh dock under
                                    v(N+1) class, every field initialized cleanly
```

## Test plan

- [x] `pytest -x` (608 tests pass)
- [x] `ruff check src/ tests/` (clean)
- [x] `godot --headless --path test_project --import` (no parse errors on edited files)
- [x] Cross-platform synthetic CI ([forensic workflow](https://github.com/hi-godot/godot-ai/actions/runs/25061868538)) demonstrates the underlying Godot bug on all three OSes
- [ ] Full CI matrix on this PR (will run when opened)
- [ ] Manual end-to-end smoke: build a v(N) zip from current main, install, click Update with a v(N+1) zip that introduces a new typed Dictionary field, verify no crash and dock recreates

## Closes / refs

- Closes [#245](https://github.com/hi-godot/godot-ai/issues/245)
- Refs [#242](https://github.com/hi-godot/godot-ai/issues/242), [#247](https://github.com/hi-godot/godot-ai/issues/247)
- Upstream: [dsarno/godot#1](https://github.com/dsarno/godot/issues/1)

### DIFF
--- a/.github/workflows/forensic-issue-245.yml
+++ b/.github/workflows/forensic-issue-245.yml
@@ -1,0 +1,85 @@
+name: Forensic / issue-245 dictionary hot-reload
+
+# Manual one-shot to verify the macOS-only "newly-injected typed Dictionary
+# field stays NIL after hot-reload" bug across Linux, macOS, and Windows
+# Godot 4.6.2-stable runners. Triggered via workflow_dispatch only.
+
+on:
+  workflow_dispatch:
+
+jobs:
+  synthetic:
+    name: Synthetic / ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: chickensoft-games/setup-godot@v2
+        with:
+          version: 4.6.2
+          use-dotnet: false
+
+      - name: Run synthetic dictionary-injection test (Linux/macOS)
+        if: matrix.os != 'windows-latest'
+        shell: bash
+        run: |
+          set +e
+          godot --headless --editor \
+            --path docs/forensic/issue-245/repro-dictinject \
+            --quit-after 1500 > /tmp/synthetic.log 2>&1
+          rc=$?
+          echo "::group::Full Godot output"
+          cat /tmp/synthetic.log
+          echo "::endgroup::"
+          echo "exit code: $rc"
+          echo
+          echo "===== Verdict ====="
+          if grep -q "post-reload property names" /tmp/synthetic.log; then
+            grep "post-reload property names" /tmp/synthetic.log
+          fi
+          if grep -q "inst.get('injected_dict')" /tmp/synthetic.log; then
+            grep "inst.get('injected_dict')" /tmp/synthetic.log
+          fi
+          if grep -q "Program crashed with signal" /tmp/synthetic.log; then
+            echo "RESULT: CRASH (Dictionary::keys on null _p)"
+          elif grep -q "DONE — no SIGABRT" /tmp/synthetic.log; then
+            echo "RESULT: OK (no crash, hot-reload initialized field correctly)"
+          else
+            echo "RESULT: INDETERMINATE (test did not complete normally; rc=$rc)"
+          fi
+
+      - name: Run synthetic dictionary-injection test (Windows)
+        if: matrix.os == 'windows-latest'
+        shell: pwsh
+        run: |
+          $log = "$env:TEMP\synthetic.log"
+          # `godot.exe` is on PATH after setup-godot
+          & godot --headless --editor `
+            --path docs/forensic/issue-245/repro-dictinject `
+            --quit-after 1500 *> $log
+          $rc = $LASTEXITCODE
+          Write-Host "::group::Full Godot output"
+          Get-Content $log
+          Write-Host "::endgroup::"
+          Write-Host "exit code: $rc"
+          Write-Host ""
+          Write-Host "===== Verdict ====="
+          $content = Get-Content $log -Raw
+          if ($content -match "post-reload property names: .*") {
+            Write-Host $matches[0]
+          }
+          if ($content -match "inst\.get\('injected_dict'\): .*") {
+            Write-Host $matches[0]
+          }
+          if ($content -match "Program crashed with signal") {
+            Write-Host "RESULT: CRASH (Dictionary::keys on null _p)"
+          } elseif ($content -match "DONE — no SIGABRT") {
+            Write-Host "RESULT: OK (no crash, hot-reload initialized field correctly)"
+          } else {
+            Write-Host "RESULT: INDETERMINATE (test did not complete normally; rc=$rc)"
+          }
--- a/.github/workflows/forensic-issue-245.yml
+++ b/.github/workflows/forensic-issue-245.yml
@@ -6,6 +6,12 @@ name: Forensic / issue-245 dictionary hot-reload
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - dlight/wizardly-payne-37369d
+    paths:
+      - .github/workflows/forensic-issue-245.yml
+      - docs/forensic/issue-245/**
 
 jobs:
   synthetic:
--- a/.github/workflows/forensic-issue-245.yml
+++ b/.github/workflows/forensic-issue-245.yml
@@ -1,8 +1,9 @@
 name: Forensic / issue-245 dictionary hot-reload
 
-# Manual one-shot to verify the macOS-only "newly-injected typed Dictionary
-# field stays NIL after hot-reload" bug across Linux, macOS, and Windows
-# Godot 4.6.2-stable runners. Triggered via workflow_dispatch only.
+# Manual one-shot to verify the "newly-introduced typed Dictionary field
+# stays NIL after hot-reload" bug across Linux, macOS, and Windows Godot
+# 4.6.2-stable runners. Initial macOS run from a Mac confirmed the crash;
+# this workflow checks Linux + Windows + a second macOS for completeness.
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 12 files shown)


dsarno:
## Manual end-to-end smoke results

Ran three headless smoke configs against macOS Godot 4.6.2-mono on this Mac:

1. **Fixed (PR-branch v2.1.2 + this fix → synthetic v2.1.3-test with new typed Dictionary field)**: exit 0, no SIGSEGV, plugin update cycle completed, v2.1.3-test loaded. Stale-class opcode error at `plugin.gd:1085` during the post-reload init (recoverable, non-fatal — separate issue, not introduced by this PR).
2. **Control (v2.1.2 release zip → synthetic v2.1.3-test, no fix)**: exit 0, no SIGSEGV either. Headless's hot-reload timing apparently doesn't fire the #242 trigger window — same finding as the previous session's Linux test.
3. **Real upgrade path (v2.1.1 release zip → v2.1.2 release zip, no fix)**: exit 0, no #242 SIGSEGV — but #247 reproduces cleanly:

   ```
   ERROR: Error constructing a GDScriptInstance: 'RefCounted(editor_handler.gd)::_init':
          Method expected 4 argument(s), but called with 5
      at: _create_instance (modules/gdscript/gdscript.cpp:200)
      GDScript backtrace (most recent call first):
          [0] _enter_tree (res://addons/godot_ai/plugin.gd:103)
          [1] _reload_after_update (res://addons/godot_ai/mcp_dock.gd:1903)
   ```

   That's #247 verbatim. Confirms the headless framework works for at least one of the upgrade-window bugs; #242 specifically is the one that needs interactive-editor hot-reload semantics.

### What this means for #249

- The fix doesn't introduce regressions in any path I exercised (every config that ran the cycle reached `exit 0`).
- The fix's correctness for #242 is established by source analysis + the cross-platform CI synthetic at [docs/forensic/issue-245/repro-dictinject/](docs/forensic/issue-245/repro-dictinject/) (which crashes byte-identically to #242 on all three OSes — see [run 25061868538](https://github.com/hi-godot/godot-ai/actions/runs/25061868538)).
- I cannot verify the fix prevents the #242 SIGSEGV via this headless smoke because the trigger doesn't fire there. The previous Linux session's full v2.1.1 → v2.1.2 path also didn't crash on Linux/headless; this matches.
- For interactive-editor verification, the user (or a reviewer) would need to install v2.1.3 (this PR), open Godot, install a synthetic v2.1.3.x with a new typed Dictionary field referenced in `_exit_tree`, and click Update — the fix should cleanly detach the dock pre-reload.

dsarno:
## Manual interactive smoke caught a regression — converted to draft

User ran the manual smoke: opened `~/Documents/godot-ai-pr249-smoke/` in interactive Godot 4.6.2-mono with this PR's plugin, saw the Update banner, clicked Update. Result: editor crashed back to project picker.

### New crash signature (different from #242)

Crash log: `~/Library/Logs/DiagnosticReports/Godot-2026-04-28-090708.ips`

```
Exception: EXC_BAD_ACCESS (SIGBUS), KERN_PROTECTION_FAILURE at 0x18570af48
Thread: com.apple.main-thread
Top frames:
  GDScriptFunction::call ... + 88512   ← actually mid-bytecode in some other function
  GDScriptInstance::callp + 224
  Object::callp + 184
  Callable::callp + 224
  CallQueue::_call_function + 312
  CallQueue::flush + 348
  SceneTree::process + 324
```

Faulting address `0x18570af48` is in macOS reserved/commpage VM space (`vmRegionInfo: commpage (reserved)`), not null+0x38 like #242. This is **not** the Dictionary-NIL hot-reload hazard — it's stack corruption.

The crash PC instruction is `ldr x8, [sp, #0x18]` — a load from the local stack frame. For that to fault, the stack itself must be in an invalid VM region. Consistent with calling a method on a freed instance whose vtable resolves through stale pointers, scribbling the stack along the way.

### Why this PR's approach causes the regression

`detach_dock_for_update` queue_frees the McpDock pre-reload. But McpDock has many *other* deferred Callables pending against `self` — client status refresh callbacks, UI update callbacks (`_apply_client_action_result`, `_apply_row_status`, etc.), focus-in handlers. Once queue_free runs at the end of the current frame, any pending deferred Callables targeting the destroyed instance fire on freed memory → SIGBUS in CallQueue::flush.

The original v2.1.2 in-place reload path kept the dock alive across the toggle, so deferred callbacks always had a valid target. This PR's approach surgically dodges #242 (the Dictionary-NIL crash) but introduces this new crash mode.

### Reproducer state

- Crash log: `Godot-2026-04-28-090708.ips`
- Plugin state on disk after crash: `version="2.1.3-test"` (extract completed before crash)
- User had `Mode override: Force user`, MCP server connected, 7/18 clients configured. The 7 client rows mean lots of pending deferred status callbacks at the moment Update was clicked — likely amplifies the stale-Callable issue.

### Path forward

The `detach_dock_for_update` approach as-written is unsafe. Two options I see:

**Option A (programmatic editor restart)**: skip in-place reload entirely. After extract, `OS.create_process(OS.get_executable_path(), OS.get_cmdline_args())` + `get_tree().quit()`. New editor process loads v(N+1) cleanly, no hot-reload, no NIL fields, no stale Callables. UX: editor flickers briefly, relaunches in same project. Unsaved scenes are lost, but that's the same risk as any plugin enable/disable cycle today.

**Option B (drain ALL deferred Callables targeting the dock before freeing)**: theoretically possible but Godot's CallQueue isn't enumerable from GDScript, so this requires either upstream Godot work or a brittle "wait several frames before reload" delay.

Leaning toward A. Will redesign and re-open or supersede this PR.

dsarno:
Closing as superseded by #250. This PR was the rejected pre-detach + queue_free attempt; #250 uses the detached runner path, preserves the no-restart constraint, fixes the scan wait review issue, and passed the updated interactive smoke.

Issue #235, opened 2026-04-27
Replace timer-based first-refresh deferral with deterministic gate (follow-up to #233/#234)


dsarno:
## Summary

Follow-up to #233 / #234. PR #234 fixes the self-update crash with a 1.5s timer-deferred first refresh. That works in observed conditions but is **heuristic, not deterministic** — there's no Godot signal we're observing to confirm GDScript hot-reload has actually settled. On slow systems, larger plugins, or future Godot releases that change hot-reload timing, the race could resurface.

This issue tracks the investigation + implementation of a deterministic fix.

## Why a timer was chosen as the immediate stopgap
- The observed crash was reproduced in the wild on a real user's update from v2.0.1 → v2.1.0
- Every hour without a fix risks more user crashes
- The deterministic alternatives all need investigation or carry meaningful UX cost
- A heuristic that works in practice >> 100% crash rate

### Option A — Synchronous first refresh on main thread
Run the very first refresh inline rather than via the worker. No thread → no race.

- **Pro**: Bulletproof. No reliance on timing, signals, or assumptions.
- **Con**: 18 CLI probes on main thread = 1–10s editor freeze on plugin reload, depending on which CLIs are installed. Partial regression of #228's win (CLI probes were moved off main precisely because they froze focus events).
- **Mitigation**: Skip CLI probes on the first refresh; only do JSON/TOML strategies sync. Defer CLI probes to focus-in. Means CLI clients show stale dots briefly.

### Option B — Drop auto-refresh from `_build_ui`; fire only on focus-in or manual click
No worker spawned automatically → no race.

- **Pro**: Dead simple, deterministic by elimination.
- **Con**: On in-place reload (the self-update path), the editor stays focused — `NOTIFICATION_APPLICATION_FOCUS_IN` doesn't fire. Dock would show blank/cached state until the user manually refocuses or clicks Refresh.
- **Mitigation**: Add a banner prompting the user to click Refresh after update.

### Option C — Listen for `EditorFileSystem.script_classes_updated`
One-shot listener before firing the worker. Wait for the signal to confirm hot-reload completed.

- **Pro**: Genuinely deterministic if the signal brackets the hot-reload window.
- **Con**: **Unverified**. Earlier empirical testing showed `is_scanning()` and `filesystem_changed` fire *before* bytecode swap completes (see #229 comment thread). `script_classes_updated` may have the same problem.
- **Investigation needed**:
[...]

PR #250, opened 2026-04-28
[codex] fix(dock): reload self-update from detached runner

### LINKED ISSUE #245: Identify the "crashing dock child" GDScript _exit_tree from #242 (forensic followup)

dsarno:
## Summary
- hand off downloaded self-updates from `McpDock` to a detached update runner before any extraction writes
- synchronously remove the live dock from editor docks so its `_exit_tree` runs on pre-update bytecode, then disable the old plugin before extraction and filesystem scan
- re-enable the plugin after scan and lazily `queue_free` the orphaned old dock after deferred callbacks have drained
- add static guard tests for ordering, no restart calls, no `await` across hot-reload, and no typed `Dictionary`/`Array` fields on the runner

## Why
Godot 4.6.2 can hot-reload a live script instance with newly added typed `Dictionary`/`Array` storage left as NIL. In the v(N) -> v(N+1) self-update path, that made the old `McpDock` execute new `_exit_tree` bytecode against old storage and crash in `Dictionary.keys()`. This path avoids keeping plugin-owned instances alive across the scan.

Fixes #245. Refs #242 and #247.

## Validation
- `pytest -q`
- `ruff check src tests`
- `godot --headless --path test_project --import` with no parse/load errors in the import log
- PR CI: https://github.com/hi-godot/godot-ai/actions/runs/25067331195 passed, including cross-platform Godot 4.6.2 reload smoke on Linux, macOS, and Windows
- interactive smoke: Godot 4.6.2 mono, `~/Documents/godot-ai-pr249-smoke`, synthetic `2.1.3-test` update adding `_vnp1_trigger_dict: Dictionary = {}`; clicked Update; no crash and no new `.ips`; console saw runner disable old plugin, old plugin unload, runner enable new plugin, and new plugin load. The `[v(N+1) _exit_tree]` trigger only ran later on manual editor shutdown and returned 0 entries from `.keys()`.

## Forensic synthetic note
The dedicated #245 synthetic forensic workflow from PR #249 is not part of this implementation PR. I ran it separately on temporary branch `tmp/forensic-issue-245-run`, based on this PR head plus only the forensic workflow/docs: https://github.com/hi-godot/godot-ai/actions/runs/25067538231.

That run intentionally exercises the upstream Godot bug, not the godot-ai self-update workaround. It still reports `RESULT: CRASH (Dictionary::keys on null _p)` on Linux, macOS, and Windows, with `inst.get('injected_dict'): typeof=0 value=<null>` before the crash. That confirms the engine bug still exists; the plugin fix avoids keeping godot-ai plugin-owned instances alive across the hot-reload boundary.

### DIFF
--- a/plugin/addons/godot_ai/mcp_dock.gd
+++ b/plugin/addons/godot_ai/mcp_dock.gd
@@ -95,6 +95,8 @@ var _client_status_refresh_started_msec: int = 0
 var _client_status_refresh_generation: int = 0
 var _client_status_refresh_shutdown_requested := false
 var _client_status_refresh_timed_out := false
+var _client_status_refresh_deferred_until_filesystem_ready := false
+var _client_status_refresh_deferred_force := false
 ## Set for the duration of `_install_update` — extract-overwrite of plugin
 ## scripts on disk would crash any worker mid-`GDScriptFunction::call`
 ## (confirmed via SIGABRT in `VBoxContainer(McpDock)::_run_client_status_refresh_worker`).
@@ -192,6 +194,7 @@ func _process(_delta: float) -> void:
 		return
 	_prune_orphaned_client_status_refresh_threads()
 	_check_client_status_refresh_timeout()
+	_retry_deferred_client_status_refresh()
 	_update_status()
 	if _log_section.visible:
 		_update_log()
@@ -1728,6 +1731,9 @@ func _perform_initial_client_status_refresh() -> void:
 		return
 	if _self_update_in_progress:
 		return
+	if _is_editor_filesystem_busy():
+		_defer_client_status_refresh_until_filesystem_ready(false)
+		return
 	if _client_status_refresh_in_flight:
 		return
 
@@ -1821,6 +1827,10 @@ func _request_client_status_refresh(force: bool = false) -> bool:
 		return false
 	if _client_rows.is_empty():
 		return false
+	if _is_editor_filesystem_busy():
+		if force:
+			_defer_client_status_refresh_until_filesystem_ready(force)
+		return false
 
 	var client_probes: Array[Dictionary] = []
 	for client_id in _client_rows:
@@ -1839,6 +1849,36 @@ func _request_client_status_refresh(force: bool = false) -> bool:
 		_refresh_clients_summary()
 		return false
 	return true
+
+
+func _is_editor_filesystem_busy() -> bool:
+	var fs := EditorInterface.get_resource_filesystem()
+	return fs != null and fs.is_scanning()
+
+
+func _defer_client_status_refresh_until_filesystem_ready(force: bool) -> void:
+	## Godot can still be reparsing/reloading plugin scripts while the editor
+	## filesystem is busy. Do not spawn a worker into that window: the worker
+	## can call plugin GDScript while the main thread is reloading it, which
+	## crashes in `GDScriptFunction::call`.
+	_client_status_refresh_deferred_until_filesystem_ready = true
+	_client_status_refresh_deferred_force = _client_status_refresh_deferred_force or force
+
+
+func _retry_deferred_client_status_refresh() -> void:
+	if not _client_status_refresh_deferred_until_filesystem_ready:
+		return
+	if _self_update_in_progress or _client_status_refresh_shutdown_requested:
+		return
+	if _client_status_refresh_in_flight:
+		return
+	if _is_editor_filesystem_busy():
+		return
+
+	var force := _client_status_refresh_deferred_force
+	_client_status_refresh_deferred_until_filesystem_ready = false
+	_client_status_refresh_deferred_force = false
+	_request_client_status_refresh(force)
 
 
 func _run_client_status_refresh_worker(client_probes: Array[Dictionary], server_url: String, generation: int) -> void:
--- a/plugin/addons/godot_ai/mcp_dock.gd
+++ b/plugin/addons/godot_ai/mcp_dock.gd
@@ -97,6 +97,7 @@ var _client_status_refresh_shutdown_requested := false
 var _client_status_refresh_timed_out := false
 var _client_status_refresh_deferred_until_filesystem_ready := false
 var _client_status_refresh_deferred_force := false
+var _client_status_refresh_deferred_initial := false
 ## Set for the duration of `_install_update` — extract-overwrite of plugin
 ## scripts on disk would crash any worker mid-`GDScriptFunction::call`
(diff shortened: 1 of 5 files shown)


dsarno:
Follow-up on the review/focus findings:

- Removed the frame-count filesystem-scan timeout from `update_reload_runner.gd`; the runner now enables the new plugin only after Godot emits `filesystem_changed`.
- Treated automatic client-status refresh as low-priority editor UI state: focus-in still uses the #228 async/cooldown path, but if `EditorFileSystem.is_scanning()` is true it returns without spawning or queueing probe work. Explicit/manual refreshes can defer until the filesystem is idle.
- Added regression coverage for both: no scan-timeout-as-success path, no synchronous focus-in probe/sweep, and deferred refresh replaying only through `_request_client_status_refresh`.

Validation on `6e66ee8`:
- `pytest -q` -> 667 passed
- `ruff check src tests` -> clean
- `godot --headless --path test_project --import` -> no parse/load errors
- CI run 25068864821 -> green across Python, release smoke, Godot tests, and game capture smoke on Linux/macOS/Windows
- Interactive Mac smoke (`~/Documents/godot-ai-pr249-smoke`, Godot 4.6.2 mono): v2.1.2 -> synthetic v2.1.3-test update completed, old plugin disabled, new plugin enabled, no new `.ips`, and no `[v(N+1) _exit_tree]` crash-trigger print during the update.

dsarno:
Review follow-up on `8696da5`:

- Fixed the scan-deferred initial refresh path so it replays through `_perform_initial_client_status_refresh()` and preserves the #235 main-thread strategy warm-up before any status worker can start.
- Took both Copilot comments as useful/actionable:
  - runner extraction now fails closed if `FileAccess.open(..., WRITE)` or `f.get_error()` reports a write failure, instead of enabling a partially extracted plugin;
  - `install_downloaded_update()` now also removes the `source_dock` fallback from docks before handing it to the runner.

Validation:
- `pytest -q` -> 668 passed
- `ruff check src tests` -> clean
- `godot --headless --path test_project --import` -> no parse/load errors
- Interactive Godot 4.6.2 mono smoke on `~/Documents/godot-ai-pr249-smoke`: v2.1.2 -> synthetic v2.1.3-test update completed, old plugin unloaded/new plugin loaded, temp update dir consumed, no new `.ips`, and the later v(N+1) `_exit_tree` trigger returned 0 keys on a fresh initialized instance.

PR #256, opened 2026-04-29
[codex] fix camera current undo flake

dsarno:
## Summary
- harden Camera2D current switching with bounded settling instead of a single immediate retry
- displace stale viewport Camera2D references when headless editor state still points at the old current camera
- verify post-commit make_current paths for camera create, configure, and preset operations

## Root Cause
CI has been flaking because headless Godot can report stale Camera2D current state immediately after make_current/undo. The undo action was atomic, but the synchronous one-shot retry could still observe the previous viewport camera before the current slot settled.

## Validation
- bash script/verify-worktree
- Godot import + bash script/ci-check-gdscript /tmp/godot-ai-camera-fix-import.log
- focused MCP test: camera.test_configure_current_sibling_unmark_single_undo passed
- MCP camera suite: 32 passed, 0 failed
- bash script/ci-reload-test: post-churn suite 935 passed, 0 failed

### DIFF
--- a/plugin/addons/godot_ai/handlers/camera_handler.gd
+++ b/plugin/addons/godot_ai/handlers/camera_handler.gd
@@ -58,6 +58,8 @@ const _KEYS_3D := [
 ]
 
 const _DAMPING_MARGIN_KEYS := ["left", "top", "right", "bottom"]
+const _CURRENT_SETTLE_ATTEMPTS := 3
+const _CURRENT_SETTLE_DELAY_MSEC := 2
 
 
 var _undo_redo: EditorUndoRedoManager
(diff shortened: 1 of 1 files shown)


dsarno:
Follow-up plan for this flake:

This PR hardens the production camera-current path. If the same `camera.test_configure_current_sibling_unmark_single_undo` flake shows up again after this lands, the next move should be a test strategy change rather than another retry tweak.

Recommended next step if it flakes again:
- keep coverage for the behavior: `camera_configure(current=true)` switches current camera, and one undo restores the previous current camera
- refactor the assertion away from same-call-stack `Camera2D.is_current()` checks in headless editor timing
- add a bounded frame-settle/poll helper in the test runner, or split the undo-contract coverage from the Godot viewport-current integration smoke

In short: keep the behavior covered, but remove synchronous headless viewport timing as the CI gate if this mitigation is not enough.

PR #270, opened 2026-04-30
Skip MCP startup in headless editor launches

dialupdisaster:
## Summary
- Skip MCP server adoption/spawn, WebSocket, dock, logger, debugger, autoload, and handler setup when Godot runs headless.
- Add `GODOT_AI_ALLOW_HEADLESS=1` as an explicit opt-in for CI and intentional headless MCP sessions.
- Update CI headless editor jobs and add GDScript coverage for the headless gate.

This keeps consumer project CI/import runs from unexpectedly starting Python subprocesses, touching MCP ports, or writing plugin autoload settings; godot-ai's own CI opts back in because headless MCP is the thing being tested there.

## Testing
- `.venv/bin/python -m pytest -q tests/unit/test_gdscript_no_adjacent_string_concat.py tests/unit/test_script_create_import_settle.py`
- `git diff --check`

Note: Local Godot Mono headless import could not be completed because this machine reports `.NET: Assemblies not found` and crashes/hangs during headless editor startup.


dsarno:
The default-disable behavior makes sense once you frame it from the consumer's side: a user who installs this plugin in their game project doesn't want the MCP server, dock, WebSocket, and Python subprocess all spinning up every time they run `godot --headless --editor` for asset import or their own CI. Port conflicts, slow startup, `project.godot` autoload pollution — all bad surprises in a context they didn't sign up for. So the asymmetry is right: end users get the safe default; godot-ai-the-repo's CI is the unusual case (headless + MCP is exactly what we're testing) and pays an env-var on three workflow steps to opt in. Worth adding that one-sentence motivation to the PR description so it lives in the merge commit and isn't lost on future readers.

A couple of code-level things I'd push back on, though:

**1. The preload relocation doesn't do what its comment claims.** The new comment at the top of the moved block says handlers are loaded after the headless guard "so `godot --headless` does not parse the full tool graph." But `preload()` in GDScript is a compile-time directive — the referenced scripts load as dependencies of `plugin.gd` when `plugin.gd` itself is parsed, regardless of whether the call sits at script scope or inside a function body. Moving them into `_enter_tree` does not skip any loading. The real savings (no instantiation, no WebSocket, no dock, no debugger registration, no autoload writes) come entirely from the early return — which is genuine and worthwhile, just not what the comment describes.

**2. The typing regression that comes with #1.** The move converts ~25 locals from `const X := preload(...)` / `var x := X.new(...)` to `var X = preload(...)` / `var x = X.new(...)`, dropping the `:=` inferred types throughout. With #1 deflated there's no parse-time benefit to justify the move at all — I'd suggest leaving the typed top-level `const` block exactly as it was. The early return at the top of `_enter_tree` already prevents instantiation in headless mode, which is where the actual cost lives.

Net: a revised version that's just **early-return guard + env override + CI opt-in + tests** — without churning the preload block — would be a cleaner merge with the same behavior. The static helpers and test coverage are nice as-is.

dsarno (approved):
Thanks for the quick turnaround on both points — motivation is now captured in the PR body, and reverting the preload relocation makes the diff exactly what it should be. LGTM.

PR #274, opened 2026-04-30
[codex] Fix self-update orphan server recovery

dsarno:
## Summary

This PR fixes the managed server lifecycle around self-update so an old `godot-ai` server cannot survive on the HTTP port and block the updated plugin from starting a compatible server.

The safety model is cross-platform: automatic cleanup now requires strong ownership proof, records and pid files are preserved unless the port is actually released, and the plugin never falls through to spawn while the port remains held. The Windows-specific part is the listener discovery path, which now uses PowerShell first so PID lookup does not depend on localized `netstat` output.

## Root Cause

The old lifecycle paths could clear managed-server records or continue startup after a failed kill attempt. On Windows, listener PID discovery depended on parsing `netstat`, which is fragile under localization. For legacy v2.2.0 servers, the plugin also needed a deterministic recovery path because those servers write `user://godot_ai_server.pid` but do not expose `/godot-ai/status`.

## Changes

- Add PowerShell-first Windows listener PID lookup with netstat fallback coverage retained.
- Replace boolean ownership checks with proof dictionaries that carry proof kind and target PIDs.
- Add strong proof tiers for managed record, legacy pidfile listener, and status matching the managed record version.
- Keep status-name-only identity out of automatic kill paths; allow it only for user-click recovery.
- Preserve managed records and pid files when a kill fails or the port remains held.
- Block spawn fall-through while the HTTP port is still occupied.
- Add incompatible-server recovery status fields and recovery APIs.
- Update the dock so Restart is shown for recoverable incompatible servers and dispatches to recovery instead of force restart.
- Add lifecycle, dock, and listener parser tests for the new proof and recovery behavior.

## Validation Run

- `git diff --check`
- `ruff check src/ tests/`
- `pytest -v` - 697 passed
- Godot headless import: `Godot --headless --path test_project --import`
- `bash script/ci-godot-tests` - 994/1008 passed, 0 failed

## Approval Test Plan

Automated checks to keep in the approval gate:

- Run `ruff check src/ tests/` and `pytest -v`.
- Run the Godot import and `bash script/ci-godot-tests` on a machine with Godot available.
- Confirm the proof tests cover managed-record proof, v2.2-style pidfile listener proof, status-version match proof, strong rejection of status-name-only proof, and recovery acceptance of status-name-only proof.
- Confirm the lifecycle tests prove drift kill with no strong targets sets incompatible, preserves the record, and does not spawn.
- Confirm the lifecycle tests prove a failed drift kill keeps the record and pid file intact and does not spawn into the held port.
- Confirm the lifecycle tests prove `force_restart_server()` preserves the record when the port stays held.
- Confirm the lifecycle tests prove `recover_incompatible_server()` returns false and leaves incompatible state intact when the port stays held.
- Confirm the dock tests prove Restart is visible only when `can_recover_incompatible=true`, and dispatch goes to `recover_incompatible_server()` for incompatible state and `force_restart_server()` otherwise.
- Confirm Windows listener tests prove PowerShell PID output parsing and netstat fallback coverage.
- Confirm command-line fingerprint tests prove case-insensitive `godot-ai` or `godot_ai` matching and require a server flag such as `--pid-file` or `--transport`.

Manual regression matrix for the original orphaned-server failure:

1. Windows v2.2.0 orphan recovery
   - Start a v2.2.0 managed server that writes `user://godot_ai_server.pid` and does not expose `/godot-ai/status`.
   - Install or reload the updated plugin.
   - Verify the old pidfile PID is a listener on the configured HTTP port and has a `godot-ai` or `godot_ai` branded command line with a server flag.
   - Verify self-update cleanup kills all branded listener PIDs on that port, waits for the port to become free, clears the record and pid file only after release, and starts exactly one new compatible server.
   - Verify Task Manager or PowerShell no longer shows the old PID or its spawned child listener.

2. Windows localized shell safety
   - On a non-English Windows locale, repeat listener detection while a server is listening.
   - Verify PowerShell `Get-NetTCPConnection` finds the listener PID even if `netstat` text would be localized.
   - Temporarily make PowerShell lookup return no PID or fail, then verify the netstat fallback path still works where possible.

3. Cross-platform managed drift cleanup
   - On macOS and Linux, start an older managed server with a valid managed record whose version differs from the plugin version.
   - Reload the plugin.
   - Verify automatic kill occurs only when strong proof exists, the port is confirmed free before record cleanup, and the new server is spawned only after the port is free.

4. Failed-kill preservation
   - Simulate or force a listener that cannot be killed, or immediately rebind the port during cleanup.
   - Trigger drift cleanup and force restart.
   - Verify the managed record and pid file remain present, the plugin enters incompatible-server state, and no replacement server is spawned into the held port.

5. Status-name-only occupant safety
   - Put a process on the configured port whose `/godot-ai/status` returns `name=godot-ai` but has no matching managed record or pidfile proof.
   - Trigger update reload or drift startup.
   - Verify the plugin does not automatically kill it.
   - Verify the incompatible banner appears with recovery only when identity proof is present.
   - Click Restart and verify user-confirmed recovery can kill the occupant; if the port remains held, verify recovery returns false and the banner/state remain.

6. Non-godot occupant safety
   - Bind the configured port with an unrelated local service.
   - Trigger plugin startup, update reload, force restart, and recovery attempts.
   - Verify no automatic kill occurs, no raw port sweep occurs, and the plugin reports incompatible or blocked state without clearing managed records prematurely.

7. Compatible adoption sanity
   - Start a compatible managed server on the configured port.
   - Reload the plugin.
   - Verify the plugin adopts it, sets `actual_name` to `godot-ai`, clears `can_recover_incompatible`, and does not show the recovery Restart path.

### DIFF
--- a/plugin/addons/godot_ai/handlers/camera_handler.gd
+++ b/plugin/addons/godot_ai/handlers/camera_handler.gd
@@ -283,7 +283,7 @@ func configure(params: Dictionary) -> Dictionary:
 			if prop_name in _NODE_TRANSFORM_KEYS:
 				msg += (
 					". Transforms live on the Node, not on the camera config — "
-					+ "use node_set_property(path=%s, properties={\"%s\": ...})" % [node_path, prop_name]
+					+ "use node_set_property(path=%s, property=\"%s\", value=...)" % [node_path, prop_name]
 				)
 			return McpErrorCodes.make(McpErrorCodes.INVALID_PARAMS, msg)
 		if prop_name == "current":
--- a/plugin/addons/godot_ai/handlers/camera_handler.gd
+++ b/plugin/addons/godot_ai/handlers/camera_handler.gd
@@ -57,6 +57,14 @@ const _KEYS_3D := [
 	"current",
 ]
 
+# Transform-shaped keys live on Node2D / Node3D, not in the camera-specific
+# schema — rejecting them without a hint sends agents searching for the wrong
+# tool.
+const _NODE_TRANSFORM_KEYS := [
+	"position", "rotation", "scale", "transform",
+	"global_position", "global_rotation", "global_scale", "global_transform",
+]
+
 const _DAMPING_MARGIN_KEYS := ["left", "top", "right", "bottom"]
 const _CURRENT_SETTLE_ATTEMPTS := 3
 const _CURRENT_SETTLE_DELAY_MSEC := 2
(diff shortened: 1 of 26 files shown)


dsarno:
Closing this draft as superseded by #275. This first branch was accidentally based on the smoke aggregation history and included unrelated commits; #275 is the clean one-commit, five-file PR against `main`.

PR #286, opened 2026-05-02
Thread live-status snapshot through the spawn path

dsarno:
## Summary

Follow-on to #285. The user's startup trace on the occupied-port path showed `http_status_probe: 5` and `phase=server_start delta_ms=2646` — five back-to-back ~500 ms HTTPClient poll loops dominating the first-paint window. This PR replaces the redundant probes with explicit data flow: probe once at the top of `_start_server`, thread the snapshot through the helpers that need it.

> **Note:** an earlier version of this PR used a per-startup probe cache (memoization). After review, that approach was rejected as treating a symptom (hidden mutable state, manual cache invalidation) rather than fixing the call graph. The current diff is the parameter-threading version.

## Approach

Each downstream helper now takes an optional `live: Dictionary = {}` and falls back to its historical fresh probe when the caller doesn't have one. This preserves behavior for non-startup callers (the dock's recover/restart buttons, `_on_server_version_*`) without any branching:

- `_evaluate_strong_port_occupant_proof(port, live={})`
- `_evaluate_recovery_port_occupant_proof(port, live={})` — threads `live` to the strong-proof delegate
- `_recover_strong_port_occupant(port, wait_s, pre_kill_live={})` — uses the snapshot for proof determination only; the kill itself invalidates it, so the function never reuses `pre_kill_live` past the kill
- `_set_incompatible_server(live, ...)` — already accepted `live`; now also forwards it to its internal `_evaluate_recovery_port_occupant_proof` call

`_start_server` becomes the orchestrator: probes once at the top, passes `live` to recovery, and re-probes unconditionally on recovery failure before handing data to `_set_incompatible_server`.

## Latent staleness bug fix (bonus)

The original no-drift branch reused the pre-recovery `live` for the failure-mode diagnostic even when `_recover_strong_port_occupant` had attempted (and failed) a kill via `legacy_pidfile_kill_targets`. After a kill the port owner can change, so the diagnostic could describe a process that no longer exists. The unified failure arm now re-probes after every recovery failure — costs one extra probe in the rare no-targets sub-case, fixes the bug structurally.

## Trace verification (Linux 4.6.2 / `python -m http.server 8000`)

| Variant | `http_status_probe` count | `server_start` phase |
|---|---|---|
| #285 baseline | 4 | 285 ms |
| **This PR** | **2** | **222 ms** |

On the user's Windows environment (~530 ms per probe) the same factor of 2 fewer probes should translate to ~**1 s saved** on the occupied-port path. In the user's drift+kill-fail scenario specifically (5 probes in their trace) the refactor collapses to 2 probes — so ~1.5 s saved on Windows for that specific path.

## Why parameter threading vs the cache approach

| | Cache (rejected) | Parameter threading (this PR) |
|---|---|---|
| `http_status_probe` (Linux baseline) | 1 | 2 |
| Hidden mutable state | yes (`_startup_probe_cache_active`) | no |
| Cache invalidation correctness | manual; one site fixed in review, others latent | n/a |
| Latent no-drift staleness bug | not addressed | **fixed** |
| API surface change | additive helper | one optional kwarg per helper, default preserves behavior |

The cache version was 1 probe faster on Linux but added hidden state and didn't fix the staleness bug. The user explicitly preferred the structural fix.

## Tests

Five new unit tests in `test_plugin_lifecycle.gd` assert the threading contract via a `probe_calls` counter on the existing `_ProofPlugin` override:

- `test_strong_proof_uses_provided_live_without_probing` — `live` short-circuits internal probe.
- `test_strong_proof_probes_when_live_omitted` — historical fall-back preserved.
- `test_recovery_proof_threads_live_through_strong_call` — `live` passes through the strong-proof delegate, both internal probe sites skipped.
- `test_recover_strong_port_occupant_threads_live_to_proof` — recovery uses `pre_kill_live` for proof determination.
- `test_set_incompatible_server_threads_live_to_recovery_proof` — diagnostic and proof determination share the same `live`.

## Test plan

- [x] `pytest -v` (691 passed)
- [x] `bash script/ci-check-gdscript` — all GDScript files OK
- [x] Headless trace A/B with port 8000 occupied: 4 probes → 2 probes
- [ ] `test_run` via MCP in editor — five new unit tests pass
- [ ] Live smoke: launch editor with port 8000 occupied; confirm dock shows incompatible-server panel quickly (no multi-second hang) and the Restart Server button still works
- [ ] Live smoke: launch editor with port 8000 free; confirm clean spawned path is unchanged
- [ ] Stack on top of #285 in CI to re-run the user's headless Windows trace; expect `http_status_probe: 2`

### DIFF
--- a/plugin/addons/godot_ai/plugin.gd
+++ b/plugin/addons/godot_ai/plugin.gd
@@ -655,26 +655,25 @@ func _start_server() -> void:
 			## kill the process currently holding the port.
 			print("MCP | managed server v%s does not match plugin v%s, restarting"
 				% [record_version, current_version])
-			if not _recover_strong_port_occupant(port, 3.0):
-				_server_started_this_session = true
-				_set_incompatible_server(_probe_live_server_status_for_port(port), current_version, port)
-				_startup_path = "incompatible"
-				push_warning(_server_status_message)
-				return
-			## Fall through to spawn.
-		else:
-			if not _recover_strong_port_occupant(port, 3.0):
-				## No recorded version drift and the live status probe did not
-				## verify an exact/current-compatible godot-ai server. A stale
-				## matching record alone is not enough ownership proof to kill
-				## the current port owner, and opening the WebSocket could route
-				## clients to an incompatible HTTP/MCP tool surface.
-				_server_started_this_session = true
-				_set_incompatible_server(live, current_version, port)
-				_startup_path = "incompatible"
-				push_warning(_server_status_message)
-				return
-			## Fall through to spawn.
+		## No recorded drift case: a stale matching record alone is not
+		## enough ownership proof to kill the current port owner, and
+		## opening the WebSocket could route clients to an incompatible
+		## HTTP/MCP tool surface — `_recover_strong_port_occupant` only
+		## acts on strong proof (managed_record / pidfile_listener /
+		## status_matches_record) so this branch is structurally safe.
+		##
+		## `live` is forwarded so the recovery path's proof helpers reuse
+		## the snapshot we already paid for. The kill itself invalidates
+		## that snapshot, so the failure-mode arm below re-probes before
+		## handing data to `_set_incompatible_server`.
+		if not _recover_strong_port_occupant(port, 3.0, live):
+			_server_started_this_session = true
+			var post_recovery_live := _probe_live_server_status_for_port(port)
+			_set_incompatible_server(post_recovery_live, current_version, port)
+			_startup_path = "incompatible"
+			push_warning(_server_status_message)
+			return
+		## Fall through to spawn.
 	else:
 		_startup_path = "free"
 
@@ -751,7 +750,12 @@ func _set_incompatible_server(live: Dictionary, expected_version: String, port: 
 	_server_actual_version = _live_version_for_message(live)
 	_server_dev_version_mismatch_allowed = false
 	_server_status_message = _incompatible_server_message(live, expected_version, port, _resolved_ws_port)
-	var proof := _evaluate_recovery_port_occupant_proof(port)
+	## `live` is the caller's most-current snapshot — pass it through to
+	## the recovery proof helper so it doesn't fire another probe of the
+	## same port. The `_set_incompatible_server` contract is "use exactly
+	## this live", so threading it down keeps the proof determination
+	## consistent with the diagnostic message we just built.
+	var proof := _evaluate_recovery_port_occupant_proof(port, live)
 	var proof_name := str(proof.get("proof", ""))
 	_can_recover_incompatible = not proof_name.is_empty()
 	print("MCP | proof: %s" % (proof_name if _can_recover_incompatible else "(none)"))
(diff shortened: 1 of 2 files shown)


dsarno:
## Verified with `GODOT_AI_STARTUP_TRACE=1` — A/B comparison

Stacked this PR's cache on top of #285's trace locally and ran headless Godot 4.6.2 against `python -m http.server 8000` (returns 404 on `/godot-ai/status`, matching the user's environment from #285).

**Without cache (PR #285 only):**
```
MCP startup trace | begin platform=Linux http_port=8000 ws_port=9500
MCP startup trace | phase=settings_registered delta_ms=3 total_ms=3
MCP startup trace | phase=server_start delta_ms=285 total_ms=288
MCP startup trace | phase=core_objects delta_ms=11 total_ms=299
MCP startup trace | phase=handlers_registered delta_ms=1 total_ms=300
MCP startup trace | phase=dock_attached delta_ms=408 total_ms=708
MCP startup trace | done path=incompatible total_ms=708 counters={ ..., "http_status_probe": 4, ... }
```

**With cache (this PR stacked on #285):**
```
MCP startup trace | begin platform=Linux http_port=8000 ws_port=9500
MCP startup trace | phase=settings_registered delta_ms=6 total_ms=6
MCP startup trace | phase=server_start delta_ms=133 total_ms=139
MCP startup trace | phase=core_objects delta_ms=9 total_ms=148
MCP startup trace | phase=handlers_registered delta_ms=1 total_ms=149
MCP startup trace | phase=dock_attached delta_ms=597 total_ms=746
MCP startup trace | done path=incompatible total_ms=746 counters={ ..., "http_status_probe": 1, ... }
```

**Direct comparison (same machine, same scenario):**

|  | Without cache | With cache | Δ |
|---|---|---|---|
| `http_status_probe` count | 4 | **1** | **−75 %** |
| `server_start` phase | 285 ms | **133 ms** | **−152 ms (−53 %)** |

On Linux each probe costs ~38 ms (fast localhost HTTPClient). On the original user's Windows trace (#285 comment, port 8000 occupied) each probe cost ~530 ms — so the same factor of 4 fewer probes should translate to **~2.0 s saved** on Windows. That matches the PR description's projection of 2.6 s → ~0.7 s on the occupied-port path.

`dock_attached` varied 408 → 597 ms between runs (run-to-run noise — it's the next ditch and what PR B in the plan would target).

(The 4-probe count without cache is fewer than the user's 5 because their Windows trace went through the version-drift branch — `record_version` matched the managed-server record, triggering an additional `_set_incompatible_server` → `_evaluate_recovery_port_occupant_proof` probe. My local setup has no managed-server record, so the helper takes the simpler path. Even in the user's 5-probe scenario the cache collapses all five into one because all five probe the same port.)

---

dsarno:
## Pivot: replaced the cache with parameter threading

After reviewing the cache approach we decided to do it the right way: thread the live-status snapshot through `_evaluate_*_port_occupant_proof`, `_recover_strong_port_occupant`, and `_set_incompatible_server` as an explicit parameter instead of memoizing via hidden mutable state.

### Why

The cache was treating a symptom (five helpers redundantly probing the same data) rather than fixing the call graph. Two specific concerns:

1. **Hidden mutable state on a hot path that previously had none.** Any future probe added inside the startup window would silently join the cache, even if its logic depended on observing live state.
2. **Cache invalidation is a perpetual maintenance hazard.** Copilot caught one staleness site (post-kill); a no-drift + `legacy_pidfile_kill_targets` + kill-failure path was still latent.

The parameter-threading version makes the data flow explicit, removes the hidden state, and structurally fixes the latent staleness (the unified failure arm in `_start_server` now re-probes on every recovery failure before handing data to `_set_incompatible_server`).

### Trade-off measured

Linux trace, port 8000 occupied (`python -m http.server`):

|  | `http_status_probe` | `server_start` |
|---|---|---|
| #285 baseline | 4 | 285 ms |
| Cache (reverted) | 1 | 133 ms |
| **This PR** | **2** | **222 ms** |

The cache was 1 probe faster on Linux (~89 ms). The parameter-threading version costs that extra probe in failure paths but gains:

- Explicit data flow, no hidden state.
- Structural fix for the post-kill staleness bug.
- No "cache invalidation" maintenance burden for future contributors.

On Windows (~530 ms per probe) the user's drift+kill-fail scenario goes 5 → 2 probes, so the user-perceived saving is still ~1.5 s on that path.

### What changed

- New: every probe-consuming helper takes `live: Dictionary = {}` and falls back to a fresh probe when omitted (preserves dock UI button + version-handshake paths verbatim).
- `_start_server` body re-organized to probe once at the top, thread `live` through, re-probe after recovery fails.
- Five new unit tests assert the threading contract via a `probe_calls` counter.

### Force-push note

Branch was force-pushed to replace the cache implementation. The two prior Copilot review comments (post-kill cache staleness, `_start_server_locked` naming) are now obsolete by construction — neither concept exists in this version.

---

dsarno:
@copilot Review.

Issue #301, opened 2026-05-03
flake: Camera2D current undo test still fails after #296


dsarno:
## Summary

`camera.test_configure_current_sibling_unmark_single_undo` is still flaky after #296. PR #300 hit the same Camera2D current-state undo symptom in `Godot tests / macOS`, even though PR #300 does not touch camera code and the same test matrix passed locally plus on Linux/Windows CI.

This looks like a recurrence of the camera-current race tracked in #140 and #278, which #296 closed by widening the settle window, checking `Viewport.get_camera_2d()` for Camera2D, and cleaning camera-suite fixtures per test.

## New failing example

- PR: https://github.com/hi-godot/godot-ai/pull/300
- Run: https://github.com/hi-godot/godot-ai/actions/runs/25291900443
- Failed job: `Godot tests / macOS` job `74144847847`
- Step: `Run handler tests`

Failure:

```text
Godot tests: 1043/1060 passed, 1 failed
FAIL: camera.test_configure_current_sibling_unmark_single_undo: After undo second should not be current. viewport_cam=UndoSecond:<Camera2D#1720486792976> first_in_tree=true second_in_tree=true
```

## Why this is probably not PR #300

PR #300 only touches lifecycle reliability code (`connection.gd`, `dispatcher.gd`, `sessions/registry.py`, error-code parity, and related tests). It does not modify `camera_handler.gd` or `test_camera.gd`.

Observed signal on the same head SHA:

- Local `script/ci-godot-tests`: `1045/1060 passed, 0 failed`
- CI `Godot tests / Linux`: passed
- CI `Godot tests / Windows`: passed
- CI `Godot tests / macOS`: failed with the camera undo symptom above

## Prior context

- #140 tracked the original macOS-only `camera.test_configure_current_sibling_unmark_single_undo` flake and noted that #120's `_verify_current_after_commit` covered the forward/apply side but not the undo-side viewport lag.
- #278 tracked a newer recurrence on PR #275, later broadened beyond macOS by a Windows failure.
- #296 closed #278 by increasing the bounded settle budget, tightening Camera2D settled-state checks against `Viewport.get_camera_2d()`, and clearing current cameras during camera test teardown.

This PR #300 failure suggests #296 reduced but did not eliminate the undo-side race.

## Suggested investigation

- Reproduce on macOS headless by looping `camera.test_configure_current_sibling_unmark_single_undo` against a post-#296 tree.
[...]

PR #305, opened 2026-05-04
camera: track logical current to make camera_get deterministic

dsarno:
### Motivation
- Headless CI shows nondeterministic `camera_get("")`/current-camera resolution because `Camera2D.is_current()` and `Viewport.get_camera_*()` can lag after create/undo/reload, causing flaky tests (Issue #301). 
- The intent is to provide a deterministic MCP-facing notion of “current camera” that is updated inside the same undo/redo callbacks that call `make_current()`/`clear_current()` so tests and tools do not depend solely on transient viewport timing.

### Description
- Added per-scene-root metadata keys `"godot_ai/logical_current_2d"` and `"godot_ai/logical_current_3d"` and helper functions ` _logical_meta_key_for`, `_set_logical_current`, `_clear_logical_current`, and `_logical_current_camera` to read/write and self-heal stale entries. (file: `plugin/addons/godot_ai/handlers/camera_handler.gd`)
- Updated `_apply_make_current` to call `_set_logical_current(cam)` and `_apply_clear_current` to call `_clear_logical_current(cam)` so logical marker updates occur inside the same do/undo callbacks as the real viewport operations. 
- Changed `get_camera` selection order to prefer the logical current first, then the viewport current, then `is_current()` scanning, and finally fall back to the first camera; and updated `list_cameras` to consider the logical marker when reporting `current`.
- Preserved existing bounded settling/force-refresh logic so the change is a deterministic metadata-backed fallback rather than a replacement of `make_current()`/`clear_current()` behavior.

### Testing
- Ran `pytest -q tests/unit/test_runtime_handlers.py -k "camera_get_empty_path or camera_list_handler"` in the environment, but the run failed with `ModuleNotFoundError: No module named 'websockets'` so unit tests could not complete. 
- No other automated tests were executed in this environment; the change includes self-healing metadata behavior to avoid stale entries during test runs.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f819133798832781713424d18b0054)

### DIFF
--- a/plugin/addons/godot_ai/handlers/camera_handler.gd
+++ b/plugin/addons/godot_ai/handlers/camera_handler.gd
@@ -182,6 +182,13 @@ static func _logical_current_camera(scene_root: Node, type_str: String = "") -> 
 
 
 
+static func _is_logical_current(scene_root: Node, cam: Node) -> bool:
+	if scene_root == null or cam == null:
+		return false
+	var logical := _logical_current_camera(scene_root, _camera_type_str(cam))
+	return logical != null and logical == cam
+
+
 # Register a current=true switch on `node` in the open undo action,
 # unmarking previously-current siblings of the same class so a single
 # Ctrl-Z reverts the whole switch.
@@ -795,9 +802,10 @@ func get_camera(params: Dictionary) -> Dictionary:
 	var keys: Array = _KEYS_2D if type_str == "2d" else _KEYS_3D
 	var prop_types := _property_type_map(node)
 	var props: Dictionary = {}
+	var is_current_effective := _is_effective_current(node) or _is_logical_current(scene_root, node)
 	for key in keys:
 		if key == "current":
-			props[key] = _is_effective_current(node)
+			props[key] = is_current_effective
 			continue
 		if prop_types.has(key):
 			props[key] = CameraValues.serialize(node.get(key))
@@ -807,7 +815,7 @@ func get_camera(params: Dictionary) -> Dictionary:
 			"path": McpScenePath.from_node(node, scene_root),
 			"type": type_str,
 			"class": node.get_class(),
-			"current": _is_effective_current(node),
+			"current": is_current_effective,
 			"properties": props,
 			"resolved_via": resolved_via,
 		}
@@ -825,12 +833,15 @@ func list_cameras(_params: Dictionary) -> Dictionary:
 
 	var cams := _list_cameras_in_scene(scene_root, "")
 	var out: Array[Dictionary] = []
+	var logical_2d := _logical_current_camera(scene_root, "2d")
+	var logical_3d := _logical_current_camera(scene_root, "3d")
 	for cam in cams:
+		var logical_current := logical_2d if cam is Camera2D else logical_3d if cam is Camera3D else null
 		out.append({
 			"path": McpScenePath.from_node(cam, scene_root),
 			"class": cam.get_class(),
 			"type": _camera_type_str(cam),
-			"current": _is_effective_current(cam) or _logical_current_camera(scene_root, _camera_type_str(cam)) == cam,
+			"current": _is_effective_current(cam) or (logical_current != null and logical_current == cam),
 		})
 	return {"data": {"cameras": out}}
 
--- a/plugin/addons/godot_ai/handlers/camera_handler.gd
+++ b/plugin/addons/godot_ai/handlers/camera_handler.gd
@@ -68,6 +68,8 @@ const _NODE_TRANSFORM_KEYS := [
 const _DAMPING_MARGIN_KEYS := ["left", "top", "right", "bottom"]
 const _CURRENT_SETTLE_ATTEMPTS := 8
 const _CURRENT_SETTLE_DELAY_MSEC := 10
+const _LOGICAL_CURRENT_2D_META := "godot_ai/logical_current_2d"
+const _LOGICAL_CURRENT_3D_META := "godot_ai/logical_current_3d"
 
 
 var _undo_redo: EditorUndoRedoManager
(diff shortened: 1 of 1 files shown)


dsarno:
## Review

The PR adds a "logical current camera" marker stored as metadata on the edited scene root, written from inside `_apply_make_current` / `_apply_clear_current`, and consulted by `get_camera("")` and `list_cameras`. The intent is to give MCP a deterministic answer for "which camera is current" that doesn't depend on the laggy `Viewport.get_camera_2d()` / `is_current()` state in headless macOS CI.

### Verdict: not okay as-is — several blockers.

---

### 🔴 Blocker 1 — This change does not actually fix the failing test

The flake the PR cites is `camera.test_configure_current_sibling_unmark_single_undo` (issue #301). Look at what that test actually asserts (`test_project/tests/test_camera.gd:243-244`):

```gdscript
assert_eq(second_node.is_current(), false, "After undo second should not be current. ...")
assert_eq(first_node.is_current(), true,  "Single undo should restore original current camera. ...")
```

It calls `Camera2D.is_current()` directly on the node — it never goes through `get_camera` or `list_cameras`. The metadata layer in this PR is invisible to `Camera2D.is_current()`, so the failing assertion will continue to fail in exactly the same way. The PR plasters over read-side symptoms while leaving the underlying viewport-lag race in `_apply_clear_current` untouched.

The PR description concedes that the unit tests never ran (`ModuleNotFoundError: websockets`), and no GDScript test was added or modified. There is no evidence the change makes the cited flake go away.

---

### 🔴 Blocker 2 — `list_cameras` can now report two cameras as current at once

The new line at `camera_handler.gd:833`:

```gdscript
"current": _is_effective_current(cam) or _logical_current_camera(scene_root, _camera_type_str(cam)) == cam,
```

This is an **OR** of two diverging sources of truth. During the exact race window the PR is trying to paper over (post-undo, viewport hasn't caught up):

- **First** (the camera that should be current): `_is_effective_current` = false, logical = First → reported `current=true` ✓
- **Second** (the displaced camera the viewport still points at): `_is_effective_current` = true, logical = First → reported `current=true` ✗

Both cameras get `current=true` in the list response. That breaks the per-class single-current invariant the rest of the handler maintains.

---

### 🔴 Blocker 3 — Inconsistent semantics across the three read paths

After this change, the three "is this camera current?" answers can disagree:

| Read path | Source of truth |
|---|---|
| `get_camera("")` | logical → viewport → `is_current` scan → first |
| `get_camera(camera_path="…")` | `_is_effective_current(node)` only |
| `list_cameras` | `_is_effective_current(node) or logical == node` |

`get_camera(camera_path="…")` at `camera_handler.gd:783` was not updated, so a per-path lookup of the same camera that `list_cameras` reports as `current=true` may report `current=false`, and vice versa. Pick one definition of "current" and use it everywhere, or commit to a clearly-named alternative (e.g. a separate `logical_current` field alongside `current`).

---

### 🟡 Concern 4 — Scene file pollution

`Node.set_meta()` is persisted to the `.tscn` on save. Once any MCP-touched scene gets saved, every such scene carries `metadata/godot_ai/logical_current_2d` / `_3d` keys forever, pointing at a path that may have moved or been deleted. The `_logical_current_camera` self-heal only fires when the meta is read in *this* handler — saved scenes opened by humans will quietly carry this dangling state, including in user repos that ship a Godot project. Inventing a sidecar metadata namespace on the user's scene root is a bigger commitment than it appears and warrants explicit discussion.

If you keep this approach, the metadata should live somewhere that isn't serialized — e.g. a dictionary on the handler instance keyed by scene-root `InstanceID`, cleared on scene-switch.

---

### 🟡 Concern 5 — Doesn't address the root cause

The actual race is in `_apply_clear_current` at `camera_handler.gd:319`: it polls `_is_current(cam)` (which checks `cam.is_current()`) but doesn't verify the viewport slot has switched, and doesn't have a `_displace_stale_camera_2d`-style escape hatch the way `_apply_make_current` does. PR #296 added that for the make path; the symmetric undo path needs the same kind of viewport-slot-aware settle, not a parallel metadata system.

A more targeted fix: in the undo branch where `_add_make_current_to_action` re-applies make_current to the previous current camera, the displaced camera (the one being undone-away-from) is still in the viewport slot on Camera2D. `_apply_make_current` on `prev_current` *should* take over, but the test failure shows the slot lags. Strengthening that path (toggle `enabled`, or explicit clear on the displaced camera before the make) would fix the actual `is_current()` answer the test reads.

---

### 🟡 Concern 6 — Style nits

- `camera_handler.gd:181` has two blank lines where one is conventional in this file.
- `_logical_current_camera` returns `Node` but every caller treats it as a `Camera2D`/`Camera3D`. Using `_is_camera()` as a guard is fine, but the type hint could be tightened or callers could narrow.
- The metadata key namespace `godot_ai/...` is fine, but document it in the existing constants block comment so future readers don't grep blindly.

---

### Recommendation

**Request changes.** The PR ships a workaround that doesn't actually move the failing assertion, introduces an OR-style bug in `list_cameras`, leaves `get_camera(path=…)` inconsistent, and pollutes saved scene files with sidecar metadata. The right fix is in `_apply_clear_current` / the undo make-current path itself — same family as the #296 fix, applied to the symmetric viewport-slot-displacement case the undo triggers.

If the team does want a logical-current cache as defense-in-depth, it should be (a) handler-local and not serialized, (b) the *single* source of truth for the `current` field across all three read paths, and (c) accompanied by a GDScript test that loops the exact #301 scenario in headless mode and asserts `is_current()` post-undo.

---

dsarno:
Closing — superseded by #306, which landed all of this PR's intent (logical-current bookkeeping for deterministic MCP camera reads) plus the review feedback this branch left open: OR-semantics replaced with `_resolve_current` (single source of truth), bookkeeping moved off `Node.set_meta` onto a handler-instance dict to avoid `.tscn` pollution, root-cause `_force_clear_other_currents` / `_is_clear_settled` strengthening of `_apply_make_current` / `_apply_clear_current`, and three regression tests in `test_camera.gd`. Branch is also `mergeable_state: dirty` against `beta` now that #306 has merged.

---

Issue #244, opened 2026-04-28
Broader plugin.gd untyping for self-update safety (defense-in-depth follow-up to #242)


dsarno:
## Summary

Defense-in-depth follow-up to [#242](https://github.com/hi-godot/godot-ai/issues/242) / [#243](https://github.com/hi-godot/godot-ai/pull/243).

[#243](https://github.com/hi-godot/godot-ai/pull/243) untyped two specific `plugin.gd` fields (`_game_log_buffer`, `_editor_log_buffer`) that triggered a self-update parse hazard in v2.1.1 → v2.1.2. The remaining typed `class_name` fields in `plugin.gd` are still hazards if any future release refactors them — the bug class is "any future release that changes any class_name's inheritance, or that adds a new class_name file referenced by `plugin.gd`'s parse, breaks self-update for users on the previous release."

## The remaining typed fields

```gdscript
var _connection: Connection
var _dispatcher: McpDispatcher
var _log_buffer: McpLogBuffer
var _dock: McpDock
var _debugger_plugin: McpDebuggerPlugin
```

None of these have changed inheritance recently, but the same bug fires the moment one of them does — and we won't notice in review because the typed declaration *looks* idiomatic.

## Decision needed

Two paths:

**(a) Untype them all, matching the `_editor_logger` / `_game_log_buffer` / `_editor_log_buffer` pattern.** Establishes "plugin.gd does NOT type-bind to plugin classes" as a clean architectural rule. Loses static type checks on those fields (minor — the typed callees still enforce types at the runtime call site). Loses IDE autocomplete on `_dispatcher.foo` etc. — moderate UX hit.

**(b) Add a CI / lint check that flags any new typed `class_name` field in `plugin.gd`, with an inline explanation of the self-update hazard.** Keeps the existing typed fields as-is. Newly-added fields would have to opt out of the typed pattern explicitly. Probably a Python AST scan in `tests/unit/test_editor_focus_refocus.py` style, generalizing the test added in #243.

(b) is less invasive but only catches *new* additions. (a) is the structurally correct answer.

## Why this isn't on the critical path

[...]

Issue #297, opened 2026-05-03
Architecture audit: god classes, vestigial Runtime protocol, silent path bug, and other structural debt


dsarno:
## Summary

Fresh-eyes architecture audit of the Godot AI plugin stack: Python server, FastMCP tools/handlers, and the GDScript editor plugin. The issue is now both diagnosis and execution roadmap so the work can be split into PRs cleanly.

Audit branch: `claude/audit-plugin-architecture-XYtQY` (no code changes, clean tree).

**Severity legend:** **P0** = correctness bug or data loss, **P1** = latent reliability hazard or silent failure mode, **P2** = structural / maintainability.

## Progress log

- [x] **PR 0** — `beta` branch created from `main` on origin.
- [x] **PR 1** — Scene path ancestry guard. Landed in #298 (targets `beta`). Finding #4 / P0 closed.
- [x] **PR 2** — Update/config data-loss safeguards. Landed in #299 (targets `beta`). Findings #9, #10 / P0 closed. Includes the inner-restore-from-backup follow-up that surfaces `FAILED_MIXED` when `_install_zip_file`'s local restore fails.
- [x] **PR 3** — Lifecycle reliability. Landed in #300 (targets `beta`). Findings #6, #7, #11 / P1 closed; also added dispatcher malformed-handler diagnostics and stabilized the Camera2D current-state headless flake tracked in #301.
- [x] **PR 4** — Characterization tests. Landed in #304 (targets `beta`). Multi-session concurrent read+write (asyncio.gather), disconnect → reconnect → handshake → first command, deferred-response request-id threading + missing-reply timeout, ProjectHandler pause restoration on validation error, McpConnection pause-depth survives `_clear_on_disconnect`, lower-level `update_reload_runner` success path. Config-write-failure preservation left intact in #299's coverage (`test_atomic_write_preserves_destination_when_swap_fails` plus the structural pins in `test_audit_data_loss_safeguards.py`); full interactive self-update smoke remains at `script/local-self-update-smoke`.
[...]

PR #309, opened 2026-05-04
Close third self-update parse hazard: dotted access on Mcp* class names

dsarno:
## Summary

A user reported the Godot warning:

> Unable to load addon script from path: 'res://addons/godot_ai/plugin.gd'. This might be due to a code error in that script. Disabling the addon at 'res://addons/godot_ai/plugin.cfg' to prevent further errors.

We had no log lines from the user's session, but the warning matches a known failure mode that I traced to a third syntactic form of the self-update parse hazard already covered (in part) by #242 / #244 / #294.

Those issues covered two forms — `var x: McpFoo` (typed-var) and `McpFoo.new()` (constructor). `plugin.gd` had ~60 references in a third form that the existing policy and lint did not cover:

```gdscript
static var _resolved_ws_port := McpClientConfigurator.DEFAULT_WS_PORT  # static-var initializer (worst case)
var _spawn_state: String = McpSpawnState.OK                            # instance-var initializer
McpClientConfigurator.http_port()                                      # static-method access
McpWindowsPortReservation.netsh_query_count()                          # static-method access
```

All three forms resolve through Godot's global `class_name` registry at parse time. The static-var initializer is the worst case because it runs at script-load — a failed registry lookup aborts the parse before `_enter_tree` runs and surfaces as exactly the warning the user reported, with the addon auto-disabled until manual re-enable.

## Changes

- **`plugin/addons/godot_ai/plugin.gd`** — add three preload aliases at the top of the file (`ClientConfigurator`, `SpawnState`, `WindowsPortReservation`) following the existing `Connection` / `Dispatcher` / `LogBuffer` pattern; replace all 58 dotted references on non-comment lines with the aliases. Doc comments referencing `McpSpawnState.*`/`McpConnection`/etc. by their registered class_name are left intact (prose is not parsed as identifiers and we want future readers to be able to grep for the registered name).
- **`plugin/addons/godot_ai/plugin.gd`** — expand the self-update parse-hazard policy comment to call out the third form (constants and static methods) and explicitly name static-var initializers as the worst case, with the literal warning text users see when this fires.
- **`tests/unit/test_plugin_self_update_safety.py`** — new test `test_plugin_gd_does_not_access_constants_or_static_methods_via_class_name` that scans `plugin.gd` (with comments stripped) for `McpFoo.<member>` access against any registered `class_name McpFoo`. Reuses the existing `_strip_gdscript_comments` and `_registered_mcp_class_names` helpers. Skips `.new(` (already covered by the constructor test) so offender lists stay focused. Strengthen `test_plugin_gd_documents_the_untyped_policy` to require the static-var-initializer mention in the comment so the next contributor can't strip it.
- **`tests/unit/test_self_update_orphan_recovery.py`** — update one literal-string assertion from `_spawn_state != McpSpawnState.OK` to `_spawn_state != SpawnState.OK` (the new alias). The test was a structural pin on plugin.gd source and follows the rename.

## Why not also fix `mcp_dock.gd` etc.

The hazard policy specifically targets the addon's entry script — the file `plugin.cfg` points to via `script="plugin.gd"`. A parse failure there auto-disables the addon and surfaces the user-visible warning. Other scripts in the addon tree (e.g. `mcp_dock.gd`) parse later, and a failure there produces a different downstream symptom outside this PR's scope.

## Test plan

- [x] `pytest -v` — 723 passed (including the new structural test)
- [x] `script/ci-check-gdscript` — All GDScript files OK
- [x] Lint smoke: hand-introducing one `McpClientConfigurator.DEFAULT_WS_PORT` reference into `plugin.gd` reproduces the failure with the expected paste-over-ready offender list — confirmed before commit.
- [ ] Live editor smoke: I was not able to reproduce the user's warning locally on Godot 4.6.2 (the project parses cleanly), so the live-editor smoke is the maintainer's call. Worth running the self-update interactive smoke (`script/local-self-update-smoke`) after merging since that's the path most likely to exercise the hazard window in the wild.

https://claude.ai/code/session_01Jy6tv13XJ4CwxSTCDMQiji

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jy6tv13XJ4CwxSTCDMQiji)_

### DIFF
--- a/plugin/addons/godot_ai/plugin.gd
+++ b/plugin/addons/godot_ai/plugin.gd
@@ -39,6 +39,9 @@ const GameLogBuffer := preload("res://addons/godot_ai/utils/game_log_buffer.gd")
 const EditorLogBuffer := preload("res://addons/godot_ai/utils/editor_log_buffer.gd")
 const Dock := preload("res://addons/godot_ai/mcp_dock.gd")
 const DebuggerPlugin := preload("res://addons/godot_ai/debugger/mcp_debugger_plugin.gd")
+const ClientConfigurator := preload("res://addons/godot_ai/client_configurator.gd")
+const SpawnState := preload("res://addons/godot_ai/utils/mcp_spawn_state.gd")
+const WindowsPortReservation := preload("res://addons/godot_ai/utils/windows_port_reservation.gd")
 
 ## Handlers — preloaded as consts instead of registered via `class_name` so
 ## they don't pollute the project-wide global scope. A user project that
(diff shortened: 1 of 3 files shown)


dsarno:
Superseded by #312, which adapts this fix for the current `beta` stack after the #297 refactors (#307/#308/#310).

This PR is still conceptually relevant, but its patch targets `main` and references pre-beta structure such as `McpSpawnState`, so it should not be merged as-is. Closing this stale version in favor of the beta-targeted replacement.

PR #336, opened 2026-05-05
Validate scene_save for never-saved scenes

### LINKED ISSUE #303: scene_save returns generic INTERNAL_ERROR for never-saved scenes
### Summary

Calling `scene_save` when the current edited scene has never been saved to disk returns a generic internal error:

```text
INTERNAL_ERROR: Failed to save scene: Can't create
```

For MCP clients this is not actionable. The underlying condition is that the edited scene has no `scene_file_path`, so the caller needs to provide a save path via `scene_save_as` / `scene_manage(op="save_as")`.

### Environment

Live validation:

- Godot: 4.6.2-stable (official), Windows
- Protocol: 1
- Observed session: plugin 2.3.1, server 2.3.2, `server_launch_mode: dev_venv`

Source check: `v2.3.2` uses the same `scene_save` handler path for this behavior.

### Steps to reproduce

1. Open a Godot project.
2. Create or switch to an in-memory scene with a root node, but do not save it to a `.tscn` / `.scn` path.
3. Call `scene_save` with an explicit `session_id`.

### Actual result

The call returns promptly with a generic internal error:

```text
INTERNAL_ERROR: Failed to save scene: Can't create
```

Live repro timings:

```text
scene_save on unsaved scene: 0.114s, INTERNAL_ERROR: Failed to save scene: Can't create
scene_manage save_as res://_codex_unsaved_probe.tscn: 0.639s ok
scene_save after save_as: 0.579s ok, path res://_codex_unsaved_probe.tscn
```

### Expected result

`scene_save` should detect this condition before calling Godot's pathless save API and return a structured, actionable error, for example:

```text
SAVE_PATH_REQUIRED: Current scene has never been saved; call scene_save_as with a res://...tscn path.
```

Alternatively, `scene_save` could accept an optional `path` parameter and delegate to the save-as path when the current scene has not been saved before.

### Source rationale

The plugin currently calls `EditorInterface.save_scene()` directly in `SceneHandler.save_scene`. Godot returns `ERR_CANT_CREATE` when the edited scene root has an empty scene file path. The plugin then wraps that as `INTERNAL_ERROR`, losing the actionable cause.

A small preflight like `scene_root.scene_file_path.is_empty()` would let the plugin return a clearer MCP error before invoking `EditorInterface.save_scene()`.

dsarno:
## Summary
- return INVALID_PARAMS before calling Godot's pathless scene save when the edited scene has no file path
- keep scene_save_as as the supported save path for never-saved scenes
- add focused scene handler coverage for unsaved save, saved save, and save_as behavior

## Verification
- script/ci-check-gdscript
- live MCP test_run suite=scene: 24 passed, 0 failed

Fixes #303

### DIFF
--- a/plugin/addons/godot_ai/handlers/scene_handler.gd
+++ b/plugin/addons/godot_ai/handlers/scene_handler.gd
@@ -4,6 +4,8 @@ extends RefCounted
 ## Handles scene tree reading and node search.
 
 var _connection: McpConnection
+var _save_scene_callable: Callable = Callable()
+var _save_scene_as_callable: Callable = Callable()
 
 
 func _init(connection: McpConnection = null) -> void:
@@ -165,9 +167,16 @@ func save_scene(_params: Dictionary) -> Dictionary:
 	if scene_root == null:
 		return McpErrorCodes.make(McpErrorCodes.EDITOR_NOT_READY, "No scene open")
 
+	var path := scene_root.scene_file_path
+	if path.is_empty():
+		return McpErrorCodes.make(
+			McpErrorCodes.INVALID_PARAMS,
+			"Current scene has never been saved; call scene_save_as or scene_manage(op='save_as') with a res://...tscn path."
+		)
+
 	if _connection:
 		_connection.pause_processing = true
-	var err := EditorInterface.save_scene()
+	var err := _save_current_scene()
 	if _connection:
 		_connection.pause_processing = false
 
@@ -176,7 +185,7 @@ func save_scene(_params: Dictionary) -> Dictionary:
 
 	return {
 		"data": {
-			"path": scene_root.scene_file_path,
+			"path": path,
 			"undoable": false,
 			"reason": "File save cannot be undone via editor undo",
 		}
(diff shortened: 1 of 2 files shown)


dsarno:
## Copilot review triage

**6 findings, 2 agreed and applied** in 279ab85; 4 disagreed (incorrect premise).

### Agreed

1. **Recovery hint pointed at non-existent `scene_save_as` MCP tool.** Verified at `src/godot_ai/tools/scene.py:84-103` — only `scene_save` and `scene_manage(op="save_as")` are registered. Dropped the misleading `scene_save_as` mention from the error message and the test assertion.
2. **Hint hardcoded `.tscn` even though `.scn` is also accepted.** `save_scene_as` at `plugin/addons/godot_ai/handlers/scene_handler.gd:195` accepts both extensions. Broadened the wording so callers wanting a binary `.scn` aren't told their extension is unsupported.

### Disagreed

The four remaining comments — all variants of "`Node.scene_file_path` is read-only, the test won't reach the handler" — are incorrect. `scene_file_path` is a read-write property in Godot 4.x (it has a setter on the C++ side; GDScript can both read and write it). The test assignments work as written and the live test_run already passes green; the reviewer appears to have confused this with read-only properties on derived types.

`script/ci-check-gdscript`: clean. Live test_run on suite=scene continues to pass.

---

PR #339, opened 2026-05-05
Disconnect debugger screenshot timer on _clear_pending (#297 blind spot)

dsarno:
## Summary

Base: `beta`. Closes the "debugger timer leak on screenshot success" blind spot from #297 (deferred-outside-audit list).

`McpDebuggerPlugin._clear_pending` previously only erased the dict entry on screenshot success/error, leaving the `SceneTreeTimer` and the bound `_on_timeout(request_id)` lambda alive until the timer naturally fired — up to `timeout_sec` (8s default) of dead state per request, with the captured `request_id` String pinned in memory the whole time.

The fix:
- `_send_take_screenshot` now stores the bound timeout callable into `_pending[rid]["timeout_callable"]` alongside the timer.
- `_clear_pending` disconnects `timer.timeout` from that callable before erasing the dict entry, so the lambda + captured request_id can be GC'd immediately and the SceneTreeTimer is reaped on its next tick.
- `_on_timeout` still bypasses `_clear_pending` and erases directly — by the time it runs the timer has already fired and self-disconnected, so an explicit disconnect is moot.

This is the lone behavioral change. Comment weight on `_pending` was tightened to a 3-line docstring matching the file's tone.

## Verification

- `ruff check src/ tests/` — clean
- `pytest -q` — 722 passed
- `script/ci-check-gdscript` — clean
- Live MCP `test_run` against a headless Godot 4.6.2 editor (`GODOT_AI_ALLOW_HEADLESS=1`):
  - `suite=editor`: 69/82 passed, 0 failed (13 skipped — env-gated)
  - Full suite: 1027/1041 passed, 0 failed (14 skipped — env-gated)
  - New `editor.test_debugger_plugin_clear_pending_disconnects_timer` runs and passes (3 assertions)

## /simplify pass

Three review agents reviewed the diff:
- **Applied (quality)**: tightened `_clear_pending` from `Variant` + `is X` checks to typed locals (`Dictionary`, `SceneTreeTimer`, `Callable`) using `_pending.get(request_id, {})` + `pending.get("timeout_callable", Callable())` defaults. The `is_connected(cb)` check is the only load-bearing guard.
- **Applied (quality)**: trimmed the duplicate inner comment in `_clear_pending` and the 6-line docstring on `_pending` down to a 3-line note matching the file's tone.
- **Applied (quality)**: dropped `fired[0]` capture-array assertion from the new test — disconnect is synchronous and the timer has 60s of headroom, so a "did not fire" assertion is tautological.
[...]

PR #337, opened 2026-05-05
Accept scene-absolute target_path in animation presets (closes #328)

dsarno:
## Summary

Animation preset tools (`preset_fade` / `slide` / `shake` / `pulse`) historically required `target_path` to be relative to the AnimationPlayer's `root_node`, while every other scene tool takes scene-absolute paths like `/Main/Foo`. Agents repeatedly tripped on the inconsistency and got back an unhelpful "Target node not found" error that didn't mention either convention.

This PR teaches `_resolve_preset_target` to accept **both** shapes:

- **Scene-absolute** (`/Main/World/Cube`) — resolved via `McpScenePath.resolve`, then converted to a `root_node`-relative path via `root_node.get_path_to(target)` so the derived track resolves correctly at playback. Targets outside the player's `root_node` subtree are rejected with a clear error.
- **Player-root-relative** (`World/Cube`) — historical behavior, unchanged.

The not-found error now names both supported conventions so callers can self-correct without spelunking docs:

```
Target node not found at 'Foo' (resolved relative to AnimationPlayer's root_node '/Main').
Pass a path relative to root_node (e.g. "World/Cube") or a scene-absolute path (e.g. "/Main/World/Cube").
```

Also extracts `_player_root_node` — the same `is_inside_tree → root_node NodePath → get_parent` walk was duplicated three times in `animation_handler.gd` (`validate_animation`, `_resolve_track_prop_context`, the new resolver).

## Verification

- `ruff check src/ tests/` — clean
- `pytest -q` — 762 passed
- `script/ci-check-gdscript` — clean
- Live MCP `test_run suite=animation` against a headless Godot 4.6.2 editor — **81/81 passed, 0 failed**
- Live MCP `test_run` (full suite) — 1141/1157 passed, 0 failed (16 skipped — environment-gated)
- Live smoke against a real editor:
  - `preset_pulse` with `target_path: "/Main/SmokeTarget"` → succeeds, `animation_manage validate` reports `broken_count=0`
  - `preset_fade` with relative `target_path: "SmokeTarget"` → succeeds (regression check)
  - `preset_pulse` with bogus relative path → improved error names both conventions
  - `preset_pulse` with absolute path outside player's `root_node` subtree → clear rejection citing `root_node` containment requirement

## Test plan

- [x] All four presets accept scene-absolute `target_path` and store track paths relative to `root_node`
- [x] Relative `target_path` continues to work (no regression)
- [x] `target == root_node` edge case yields `.` track path (the form Godot expects)
[...]

Issue #328, opened 2026-05-05
Animation preset target_path convention is easy to misuse


dsarno:
## Summary

Animation preset tools use `target_path` relative to the AnimationPlayer root, unlike many other scene tools that accept scene-absolute paths. This is defensible but easy to misuse.

## Environment

- Repo: `hi-godot/godot-ai`
- Branch/head: `beta` at `3890bbd`
- Godot: `4.6.2-stable`
- Plugin/server: `2.3.2`
- Disposable project: `/private/tmp/godot-ai-megasmoke-stress-20260504194400`

## Repro

Given:

- AnimationPlayer at `/MegaSmokeStress/StressAnimator`
- Target node at `/MegaSmokeStress/World/StressCube`

Call:

```json
{
  "op": "preset_pulse",
  "params": {
    "player_path": "/MegaSmokeStress/StressAnimator",
    "target_path": "/MegaSmokeStress/World/StressCube",
    "animation_name": "cube_pulse",
    "overwrite": true
  }
}
```

## Observed

The tool returned:

```text
INVALID_PARAMS: Target node not found at '/MegaSmokeStress/World/StressCube' (resolved against player's root_node)
```

Retrying with:

```text
target_path="World/StressCube"
```

succeeded, and `animation_manage(op="validate")` returned `broken_count=0`.

## Expected

At minimum, docs/tool descriptions should make this path convention explicit. Alternatively, accept scene-absolute paths and internally convert to the player-root-relative track path.

## Severity

Low. This is API ergonomics and consistency, not a correctness blocker.

dsarno:
Fixed in #337 (squash-merged into `beta` as 88e2ebe). All four preset tools (`fade`/`slide`/`shake`/`pulse`) now accept scene-absolute `target_path` and convert to a `root_node`-relative track path via `get_path_to`. Targets outside the player's `root_node` subtree become `..`-prefixed track paths (matching what the relative form already accepts) instead of being rejected. Closing now since `beta` doesn't auto-close — will roll out to `main` on the next release cut.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #316, opened 2026-05-04
flake: Camera2D current undo test fails again after #311 (4th recurrence)


dsarno:
## Summary

`camera.test_configure_current_sibling_unmark_single_undo` continues to flake on `Godot tests / macOS` after the latest stabilization PR #311 (commit `9517b04`). The flake re-surfaced unprompted on PR #315's CI run for commit `f4d7f3f` (a doc-comment-only change in `mcp_dock.gd`, no camera code touched). A re-run of just the failing job passed without any code change.

This is the **fourth** distinct recurrence after #311 and represents at least the **eleventh** attempted fix on this single test.

## Failing example

- PR: #315
- Run: https://github.com/hi-godot/godot-ai/actions/runs/25338403512
- Failed job: `Godot tests / macOS` (job `74289467689`)
- Step: `Run handler tests`
- Re-run of the same SHA passed (https://github.com/hi-godot/godot-ai/actions/runs/25338403512 after `gh run rerun --failed`).

Failure line:

```text
Godot tests: 1119/1137 passed, 1 failed
  FAIL: camera.test_configure_current_sibling_unmark_single_undo: Expected true
```

## What's different about this recurrence

Previous recurrences (#140, #256, #278, #301) all failed on the **post-undo** assertions at `test_camera.gd:300-303`, which carry custom diagnostic messages. Their failure output reported viewport state and tree membership.

**This recurrence fails on a pre-undo assertion** at `test_camera.gd:283` or `:284`:

```gdscript
# Flip current to second via configure.
var result := _handler.configure({
    "camera_path": second.data.path,
    "properties": {"current": true},
})
assert_has_key(result, "data")
assert_true(_wait_for_camera_current(second_node, true))   # ← :283 — no msg
assert_true(_wait_for_camera_current(first_node, false))   # ← :284 — no msg
```

Neither assertion has a custom message, so CI can only emit `"Expected true"` — there is no information about which camera, which viewport state, or how many of the 20 × 10ms polls elapsed. **This is a test-hygiene gap that has been masking diagnostic information for at least one recurrence.** Even if the underlying race is unfixable in this PR, fixing the test messages would make every future failure self-explaining.

## Failure-mode shift implication

[...]

PR #340, opened 2026-05-05
Flush CliFinder cache on force-refresh (#297 deferred)

dsarno:
## Summary

Base: `beta`. Closes the "CLI finder cache never invalidates" deferred audit finding from #297 — fits the same pattern as #338 (#12 contract test) and #339 (debugger timer leak): pick one finding, do it well, no bundling. Triage: the only other open issues (#303, #328, #316) already have active PRs (#336, #337, #335 merged for #316's diagnostic gap).

### What it does

`McpCliFinder.find()` is the three-tier CLI lookup (well-known dirs → `bash -lc 'command -v'` → `which`/`where`) used to detect installed clients (claude, codex, gemini, …). Once a binary was resolved or missed, the result was pinned for the editor session. So if a user installed `claude` in a terminal mid-session and clicked the dock's Refresh button, the row kept reporting "not installed" until they restarted Godot.

The fix wires `ClientConfigurator.invalidate_cli_cache()` (a new no-arg facade over `CliFinder.invalidate()`) into `_request_client_status_refresh(force)` so **every** explicit-user-action callsite shares one flush site:

- `_on_refresh_clients_pressed` — manual Refresh button
- `_on_open_clients_window` — popup open
- `_refresh_all_client_statuses` — compat wrapper marked "treat as manual refresh"

Focus-in cooldown refresh (`force=false`) deliberately keeps the cache — focus fires dozens of times per session and would otherwise re-fork `which` / `bash -lc` each time, defeating the cache. Per-CLI helpers like `invalidate_uvx_cli_cache` stay alongside for event-driven cases (`_on_install_uv` knows the binary name).

The flush sits after every early-return guard (`_server_blocks_client_health`, `_self_update_in_progress`, in-flight worker, cooldown, filesystem busy) and before `Thread.start()`, so it can't race the worker. The `has_worker_alive` early-return prevents flushing while a thread reads `_cache`/`_searched`. Filesystem-busy / cooldown paths preserve `_pending_force`, so the deferred dispatcher eventually flushes once the gate opens.

### Verification

- `pytest -q` — 763 passed
- `script/ci-check-gdscript` — clean
- Live MCP `test_run suite=clients` against a headless Godot 4.6.2 editor — 82/82 passed, 0 failed
  - New `clients.test_invalidate_cli_cache_clears_all_entries` runs with 4 asserts (verbose results confirmed)
- Live MCP full `test_run` — 1138/1154 passed, 0 failed (16 skipped — environment-gated)

### /simplify pass

Three review agents reviewed the diff:

- **Applied (reuse, substantive)**: original draft only flushed at `_on_refresh_clients_pressed`. Two other force=true callsites (`_on_open_clients_window`, `_refresh_all_client_statuses` — both documented as user-action equivalents) suffered the same stale-miss bug. Folded the flush into `_request_client_status_refresh(force)` so every force=true entry point invalidates from one place; the `force` arg is already exactly the "user-driven, bypass cooldown" signal.
- **Applied (quality)**: dropped narration comment from `_on_refresh_clients_pressed` (now a one-line delegate). Trimmed Python contract test docstring + assertion messages from paragraph-length to one-sentence each. Trimmed inline GDScript test docstring; assertion messages now only carry the load-bearing one (`_searched` not dropped → next find() short-circuits on stale negative).
- **Skipped (efficiency)**: agent found nothing to fix — manual refresh is user-initiated, the ~5-entry `Dictionary.clear()` is in the noise behind the worker's `OS.execute` round-trips, and gating on `_searched.size() > 0` would just add a branch for a nanosecond's saving.

### /review pass

Independent correctness review (verdict: SHIP) confirmed:

- Concurrency: flush sits before `Thread.start()`; `has_worker_alive` early-return prevents flushing while a worker holds the cache; deferred paths preserve `_pending_force`.
- All three force=true callsites are explicit-user / external-API paths — none should opt out.
- Test reaching into `_cache` / `_searched` underscored statics is acceptable: same-plugin internal class, asserting on observable state is stronger than a `cache_size()` shim that would only exist for tests.
- Comment weight on the new gate is justified — the WHY (focus-in fork cost) and relationship to per-CLI invalidator are non-obvious.
- Test hygiene: 4 load-bearing asserts, no zero-assertion path, no `assert_has_key` without value check, no bare `return`.
- **Applied (review nit)**: changed Python lock-test from literal-string match (`"if force:\n\t\tClientConfigurator..."`) to regex match (`r"if force:\s+ClientConfigurator\.invalidate_cli_cache\(\)"`) so a future GDScript reformat that touches indentation doesn't false-fail the contract.

## Out of scope

This is the **CLI finder cache invalidation slice only.** Other deferred-outside-audit items in #297 remain open: animation_handler split #13, structured `ci-check-gdscript` exit, `setup-dev.ps1` uv parity (likely already done — both scripts have uv handling now, worth a verification PR), structured dispatcher diagnostics, broader self-update preload-alias depth-2+ work. Each is a separate PR — they don't bundle naturally because they touch different layers and concerns (per #338 / #339's framing).

Closes audit deferred-item "CLI finder cache invalidation" from #297.

---
_Generated by [Claude Code](https://claude.ai/code/session_01N2ZL6fnxqNszddJPm8axpr)_

### DIFF
--- a/plugin/addons/godot_ai/client_configurator.gd
+++ b/plugin/addons/godot_ai/client_configurator.gd
@@ -490,6 +490,16 @@ static func invalidate_uvx_cli_cache() -> void:
 		CliFinder.invalidate(name)
 
 
+## Drop the entire `CliFinder` cache. Called by the dock's manual Refresh
+## button so a freshly-installed CLI (claude, codex, gemini, …) gets
+## detected without an editor restart. Per-CLI invalidation
+## (`invalidate_uvx_cli_cache`) is preferred when the dock knows which
+## binary changed; this catch-all fits the user-clicked-Refresh case
+## where any CLI may have been installed since the last sweep.
+static func invalidate_cli_cache() -> void:
+	CliFinder.invalidate()
+
+
 static var _uv_version_cache: String = ""
 static var _uv_version_searched: bool = false
 
--- a/plugin/addons/godot_ai/client_configurator.gd
+++ b/plugin/addons/godot_ai/client_configurator.gd
@@ -490,12 +490,19 @@ static func invalidate_uvx_cli_cache() -> void:
 		CliFinder.invalidate(name)
 
 
-## Drop the entire `CliFinder` cache. Called by the dock's manual Refresh
-## button so a freshly-installed CLI (claude, codex, gemini, …) gets
-## detected without an editor restart. Per-CLI invalidation
-## (`invalidate_uvx_cli_cache`) is preferred when the dock knows which
-## binary changed; this catch-all fits the user-clicked-Refresh case
-## where any CLI may have been installed since the last sweep.
+## Drop the entire `CliFinder` cache. Called from any explicit-user-action
+## refresh path (`force=true` in `_request_client_status_refresh` — manual
+## Refresh button, popup-open, compat wrapper, future external API) so a
+## freshly-installed CLI (claude, codex, gemini, …) gets detected without
+## an editor restart. Per-CLI invalidation (`invalidate_uvx_cli_cache`) is
+## preferred when the dock knows which binary changed; this catch-all
+## handles the "any CLI may have been installed since the last sweep" case.
+##
+## Thread safety: `CliFinder.invalidate()` guards `_cache` / `_searched`
+## with a mutex so it can race safely against worker threads calling
+## `find()` from `_run_client_action_worker`. The mutex is held only
+## across the dictionary clear, never across `OS.execute`, so this call
+## can never block the main thread on a subprocess.
 static func invalidate_cli_cache() -> void:
 	CliFinder.invalidate()
 
--- a/plugin/addons/godot_ai/clients/_cli_finder.gd
+++ b/plugin/addons/godot_ai/clients/_cli_finder.gd
@@ -8,8 +8,19 @@ extends RefCounted
 ##   2. Login shell lookup (`bash -lc 'command -v <exe>'`) — picks up .zshrc / .bashrc
 ##   3. Plain `which` / `where` against the inherited PATH
 ## Caches per-exe so repeated dock refreshes don't fork a shell every frame.
+##
+## Thread safety: `find()` runs on action-worker threads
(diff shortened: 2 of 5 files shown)


dsarno:
## Copilot review triage

Both findings agreed and applied in d84c9ea.

### #1 — `_cache` / `_searched` mutex (mcp_dock.gd:2078)

Valid: `McpCliStrategy.configure/check_status/remove` calls `CliFinder.find()` from action-worker threads spawned in `_run_client_action_worker` (mcp_dock.gd:1431). The PR's "flush sits before `Thread.start()`" defense covers the dock's *refresh* worker but not those per-button *action* workers.

Added `static var _mutex: Mutex` in `_cli_finder.gd` with critical sections deliberately kept tight: the slow `_resolve()` path (FileAccess + `bash -lc` / `which`, 100ms–1s) runs **outside** the lock. So a main-thread `invalidate()` only ever waits on a worker's microsecond-scale dict access, never on a subprocess fork — the off-main-thread CLI-lookup design is preserved. Trade-off: two workers racing the same exe both call `_resolve()` and write back the same answer. Wasted work in a corner case, no incorrect state.

### #2 — `invalidate_cli_cache()` doc wording (client_configurator.gd:498)

Valid, trivial: doc claimed "Called by the dock's manual Refresh button" but the new behavior fires on every `force=true` path (popup-open, compat wrapper, future external API). Updated wording and added a thread-safety note pointing at the new mutex contract.

### Test

New `test_cli_finder_cache_is_mutex_guarded` in `test_editor_focus_refocus.py` is a static-source contract test asserting:

1. `static var _mutex: Mutex = Mutex.new()` exists
2. `invalidate()` brackets its dict mutations with `_mutex.lock/unlock`
3. `_find_one()` releases the mutex **before** `_resolve()` is called — catches the most likely regression (someone "simplifying" the lock into one span enclosing the subprocess fork would re-introduce the main-thread freeze the tight critical sections exist to prevent)

`ruff check src/ tests/`: clean. `script/ci-check-gdscript`: clean. `pytest`: 764 passed (763 + 1 new).

---

Issue #343, opened 2026-05-05
Architecture audit v2: post-#297 soft spots — auth/origin gap, path-traversal, errno portability, and structural carry-overs


dsarno:
## Summary

Fresh-eyes architectural audit on top of the closed audit umbrella (#297). Baseline: `origin/beta` at `e6cb4a4`. Every claim below was verified against that ref before filing; four agent-suggested items that didn't reproduce are listed at the end so reviewers don't re-walk them.

**Calibration:** baseline is *beta*. P0 = breaks beta users in the wild today; P1 = bites at GA or under adversarial conditions; P2 = structural debt acceptable to ship beta with but should be addressed before 1.0.

**Severity legend:** **P0** = correctness/data-loss, **P1** = latent reliability hazard or security surface, **P2** = structural / maintainability.

Audit branch: `claude/audit-architecture-debt-IyrA6` (no code changes; clean tree).

---

## Progress log

**Wave 1 (P1 hardening + P2 reliability sweeps) — ✅ complete.** All P1 findings and all reliability sweeps merged to beta:

- ✅ **#345** (Audit v2 #1 — WS auth/Origin) via **PR #375** — *merged*
- ✅ **#346** (Audit v2 #2 — duplicate-ID handshake reject) via **PR #374** — *merged (bundled with #349)*
- ✅ **#347** (Audit v2 #3 — path traversal) via **PR #369** — *merged*
- ✅ **#348** (Audit v2 #4 — errno portability) via **PR #373** — *merged*
- ✅ **#349** (Audit v2 #5 — pending-Future leak) via **PR #374** — *merged (bundled with #346)*
- ✅ **#350** (Audit v2 #6 — drop SessionRegistry RLock) via **PR #370** — *merged*
- ✅ **#351** (Audit v2 #7 — Pydantic-validate WS events) via **PR #378** — *merged*
- ✅ **#352** (Audit v2 #8 — n=1 graceful failover) via **PR #379** — *merged*
- ✅ **#353** (Audit v2 #9 — filesystem_changed watchdog) via **PR #381** — *merged*
- ✅ **#354** (Audit v2 #10 — surface FAILED_MIXED in dock + editor_state) via **PR #382** — *merged*
- ✅ **#355** (Audit v2 #11 — `/godot-ai/status` probe) — *closed as addressed by #345 / PR #375*
- ✅ **#356** (Audit v2 #12 — packet-drain cap) via **PR #383** — *merged*

[...]

PR #366, opened 2026-05-05
Harden WebSocket transport: errno portability, Future leak, duplicate-ID reject

dsarno:
## Summary

Bundles three small, surgical hardening fixes from the audit-v2 umbrella (#343), all in `src/godot_ai/transport/websocket.py`. Each fix is independently scoped and pinned by a regression test; bundled because they're tightly themed (transport-layer hardening) and trivially reviewable side-by-side.

Closes findings **#2**, **#4**, and **#5** of #343.

### #4 (P1) — `errno.EADDRINUSE` portability

`src/godot_ai/transport/websocket.py:46` previously checked `e.errno == 48`. That number is macOS-only; Linux uses 98 and Windows 10048, so on those platforms the friendly "port already in use" branch never fired and the OSError propagated, taking the WS server lifespan down with a generic traceback. Replaced with `errno.EADDRINUSE`.

### #5 (P2) — `send_command` pending-Future leak

`send_command` registered `self._pending[request.request_id] = future` then awaited `ws.send(...)`. If `ws.send` raised (`ConnectionClosed` mid-send, transport error, cancellation), the pending entry was never popped and the dict grew unbounded under churn. Wrapped the send + wait in `try/finally: self._pending.pop(...)`; the response receiver still pops on the happy path so the finally is a no-op there.

### #2 (P1) — Duplicate-ID handshake hijack

`_handle_connection` registered the inbound peer with `session_id = handshake.session_id`, silently overwriting both the registry entry and `_connections[session_id]` when the same ID was already present. session_id format is `<slug>@<4hex>` — 16 bits of suffix is locally guessable, so any local peer could hijack an active session by sending a handshake with the same ID; subsequent MCP commands would route to the impostor.

Now `_handle_connection` checks `registry.get(handshake.session_id)` and rejects with WebSocket close code `4001` (RFC 6455 application-defined range) plus a structured warning log naming the existing peer's PID and project. Legitimate plugin reconnect after `editor_reload_plugin` first triggers `ConnectionClosed` → `unregister`, so the new connect still lands cleanly — pinned by `test_reconnect_after_clean_disconnect_succeeds`.

## Tests

New:
- `tests/unit/test_websocket_server_lifecycle.py` — parametrized errno coverage (macos 48, linux 98, windows 10048; one runs per CI host) plus a non-EADDRINUSE OSError propagation case.
- `tests/integration/test_websocket.py::TestDuplicateHandshake` — pins both the reject path and the legitimate-reconnect path.
- `tests/integration/test_websocket.py::TestPendingFutureCleanup` — pins both the timeout-pop and the send-raise pop behaviors (the latter monkey-patches `ws.send` to force a synthetic `ConnectionError` after the pending entry is registered).

Existing: 728 passed, 2 skipped (cross-platform errno parametrizations on the non-host OS).

## Live smoke

Spawned a real Godot 4.6.2 editor against `test_project/`, opened `main.tscn`, ran via the streamable-HTTP MCP transport:

- `editor_state` → `readiness=ready` ✓
- `test_run` → **1037/1040 passed, 0 failed** ✓
- Duplicate-ID handshake against the live session's own ID → server closed the WS with code **4001** and reason `'session id already registered'` ✓
- `editor_state` after the rejected hijack → still `readiness=ready` (original session unaffected) ✓
- Server log: `Rejecting duplicate handshake for session test-project@307e (existing pid=21071, project=/home/user/godot-ai/test_project/)` ✓

## What's NOT closed

- #1 (auth/Origin gap on the WS upgrade) — separate scope, deserves its own design discussion. This PR closes only the *takeover-of-existing-session* path; an unauthenticated local peer can still register a *new* session_id.
- #3 (path traversal in `script_create` / `filesystem_write_text`) — touches GDScript handlers; deserves its own PR with a regression test against `res://../etc/passwd.gd`.
- #7 (Pydantic-validate WS events) — bigger; needs new envelope models.
- #6 (drop SessionRegistry RLock) — already not in the code; no change needed.

## Test plan

- [x] `ruff format --check src/ tests/` clean
- [x] `pytest -q` — 728 passed
- [x] Live MCP `test_run` against Godot 4.6.2 editor — 1037/1040 passed, 0 failed
- [x] Live duplicate-ID handshake reject smoke — 4001 close received, original session unaffected
- [x] Logged server-side warning includes existing-peer pid + project_path

https://claude.ai/code/session_016ijmCD5S6QfwJGJcc5Wirp

---
_Generated by [Claude Code](https://claude.ai/code/session_016ijmCD5S6QfwJGJcc5Wirp)_

### DIFF
--- a/src/godot_ai/transport/websocket.py
+++ b/src/godot_ai/transport/websocket.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import errno
 import json
 import logging
 from typing import Any
@@ -17,6 +18,11 @@ from godot_ai.sessions.registry import Session, SessionRegistry
 logger = logging.getLogger(__name__)
 
 DEFAULT_PORT = 9500
+
+## RFC 6455 reserves 4000-4999 for application-defined close codes; we use
+## 4001 to flag a handshake rejected for duplicate session_id so a debugging
+## peer can distinguish it from a normal close.
+_CLOSE_CODE_DUPLICATE_SESSION = 4001
 
 
 class GodotWebSocketServer:
@@ -39,7 +45,10 @@ class GodotWebSocketServer:
             ):
                 await asyncio.Future()  # run forever
         except OSError as e:
-            if e.errno == 48:  # Address already in use
+            ## EADDRINUSE differs by platform (macOS 48, Linux 98, Windows
+            ## 10048) — let the stdlib name resolve it so the friendly
+            ## branch fires everywhere.
+            if e.errno == errno.EADDRINUSE:
                 logger.warning(
                     "WebSocket port %d already in use — another server instance may be running. "
                     "MCP tools will work but the Godot plugin won't connect to this instance.",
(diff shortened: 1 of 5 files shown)


dsarno:
Closing — landed on the wrong branch. Audit v2 (#343) is baselined on `beta`; opening a fresh PR against `beta` with the same three fixes (rebased clean onto `aac5483`).

---

Issue #355, opened 2026-05-05
[Audit v2 #11 — P2] /godot-ai/status HTTP endpoint is unauthenticated probe


dsarno:
Audit v2 finding #11 — see umbrella #343.

**File:** `src/godot_ai/server.py:208-219`

Comment-acknowledged. Combined with #1, a browser tab can probe-and-confirm the server is running before mounting a DNS-rebinding attack. Beta-acceptable; trim before broader exposure.

**Fix shape:** require an Origin allowlist or `?token=` matching the spawned-server PID-file secret. Or move it to a Unix-domain-socket / named-pipe channel that browser JS can't reach.

**Blocked by:** finding #1 (shared fix surface — best done together)

dsarno:
Closing — addressed by #345 / PR #375.

PR #375's `LocalhostOnlyHTTPMiddleware` is wired as the outermost wrap on every HTTP transport (`http`, `streamable-http`, `sse`), so `/godot-ai/status` is now guarded by the same Origin/Host allowlist as the FastMCP endpoints. Live smoke in the PR explicitly exercises `Host: attacker.example.com` against `/godot-ai/status` returning 403, plus the loopback-Host + browser-Origin DNS-rebinding fingerprint.

The remaining "PID-file token" path from this issue's fix shape is still available if the threat model later broadens to multi-user shared-loopback or container-shared 127.0.0.1 — but for the named beta-acceptable scope, the middleware is the fix.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #11, opened 2026-04-15
List parameters arrive as JSON strings when called from Claude Code MCP client


dsarno:
## Summary

Any MCP tool whose input schema declares a required `list` parameter fails with a pydantic `list_type` validation error when invoked from Claude Code's MCP client — the `commands`/`paths`/etc. argument arrives at the server as a JSON string instead of a parsed list.

The server tools and handlers are correctly typed; the problem is in the wire-level bridge between Claude Code and FastMCP input validation. `pytest` integration tests (which drive the same tools via FastMCP's in-process Python client) pass.

## Reproduction

From inside a Claude Code session with this MCP server connected:

```
batch_execute commands=[{"command": "create_node", "params": {"type": "Node3D", "name": "X", "parent_path": "/Main"}}]
```

Result:

```
1 validation error for call[batch_execute]
commands
  Input should be a valid list [type=list_type,
    input_value='[{"command": "create_nod...arent_path": "/Main"}}]',
    input_type=str]
```

Same error on `editor_selection_set paths=["/Main/Camera3D"]` — different tool, same systemic cause. Note `input_type=str`: pydantic sees a JSON-encoded string, not a parsed list.

## Affected tools

Every tool with a required `list[...]` parameter, including at minimum:

- `batch_execute` — `commands: list[dict]`
- `editor_selection_set` — `paths: list[str]`
- `filesystem_reimport` — `paths: list[str]`
- `performance_monitors_get` — `monitors: list[str] | None` (likely affected when supplied)

## What works

- `pytest -v` — 310/310 passing, including `TestBatchExecuteTool`. FastMCP's Python `Client` serializes params correctly.
- Tools taking `value: str | int | float | bool | dict | list | None` (union) — pydantic accepts the string form and the handler accidentally tolerates it in some cases. Not a workaround; just masks the symptom for union-typed params.

## Suggested fix (server-side workaround)

Since the root cause is outside this repo, the pragmatic unblock is accepting `list | str` at the tool boundary and JSON-parsing strings before delegating to the handler. Either:

1. Preprocess in each affected tool wrapper (`json.loads` when `isinstance(commands, str)`).
2. A shared pydantic `BeforeValidator` that coerces JSON strings to lists for known list params.

Option 2 keeps the tool bodies clean. Should also apply to `dict`-typed params for robustness, but I haven't seen dict stringification in practice (only list).

## Priority

[...]

PR #369, opened 2026-05-05
[Audit v2 #3] Reject res:// path traversal in script + filesystem write tools

### LINKED ISSUE #347: [Audit v2 #3 — P1] script_create / filesystem_write_text accept ../traversal
Audit v2 finding #3 — see umbrella #343.

**Files:** `plugin/addons/godot_ai/handlers/script_handler.gd:31, 122, 157, 289`, `handlers/filesystem_handler.gd:13, 43, 90` (19 sites total)

Path validation is only `path.begins_with("res://")`. `res://addons/../../../etc/passwd.gd` passes; whether Godot rejects it depends on platform/version.

LLM-driven path generation (prompt injection, agent typos, untrusted issue/PR text in context) makes this a real escape vector for the only write tools that produce arbitrary disk content.

**Fix shape:** after the prefix check, `ProjectSettings.globalize_path(path).simplify_path()` and verify it's still under `globalize_path("res://").simplify_path()`. Reject any path containing `..`. Add regression test with `res://../etc/passwd.gd`.

**Blocked by:** none
**Blocks:** none

dsarno:
## Summary

Closes #347 — audit-v2 P1 finding #3 from umbrella #343.

`script_handler.gd` and `filesystem_handler.gd` validated agent-supplied
paths only with `path.begins_with("res://")`. That accepts payloads like
`res://../etc/passwd.gd`; whether Godot rejects them depends on platform/
version. Combined with LLM-driven path generation (prompt injection,
agent typos, untrusted issue/PR text in context), this is a real escape
vector for the only write tools that produce arbitrary disk content
(`script_create`, `filesystem_write_text`, `patch_script`) and for the
matching reads (info-disclosure surface).

## Fix

Adds `plugin/addons/godot_ai/utils/path_validator.gd` →
`McpPathValidator.validate_resource_path(path)` with the four layers
called for in the issue's fix shape:

1. non-empty
2. `path.begins_with("res://")`
3. no `..` substring (cheap defence-in-depth)
4. `ProjectSettings.globalize_path(path).simplify_path()` must still
   resolve under `globalize_path("res://").simplify_path()` (boundary
   verification with trailing-slash handling so `/proj_evil/...` can't
   pretend to be inside `/proj`)

Migrates every prefix-check call site named in #347:

- `script_handler.gd` lines 31 / 122 / 157 / 289 →
  `create_script`, `read_script`, `patch_script`, `find_symbols`
- `filesystem_handler.gd` lines 13 / 43 / 90 →
  `read_file`, `write_file`, `reimport` (per-path inside the loop)

No bare `begins_with("res://")` remains in either handler.

## Tests

- `test_project/tests/test_path_validator.gd` — McpTestSuite covering
  happy path, prefix variants (`user://` rejected), and four traversal
  shapes including the exact `res://../etc/passwd.gd` payload from the
  audit body.
- `test_project/tests/test_script.gd` / `test_filesystem.gd` — per-
  handler regression tests asserting `INVALID_PARAMS` on the traversal
  payload at every entry point. The two write tests also assert no
  file landed on disk outside the project root.
- `tests/unit/test_path_traversal_guard.py` — source-structure pytest
  (5 tests) pinning the validator's existence, all four required check
  layers, and the absence of bare `begins_with("res://")` patterns in
  the affected handlers. Drift here would silently weaken the security
  boundary without any single behavioural test catching it.

## Test plan

- [x] `ruff format` applied to the new test file (other files have
      pre-existing format drift on beta — not touched)
- [x] `pytest -q` — 772 passed (5 new)
- [x] `script/ci-check-gdscript` — all GDScript files OK (the new
      validator + test suite parse cleanly)
- [ ] Live editor `test_run` via MCP — not feasible from this sandboxed
      CLI session (the running godot is `--headless` without the
      `GODOT_AI_ALLOW_HEADLESS=1` override, so the plugin returns
      early). The Godot CI matrix (Linux + macOS + Windows) will
      exercise the new `path_validator` suite and the new handler
      regression tests on PR.

## Scope

Per the audit prompt's "One audit finding == one PR" rule, only the
seven sites named in #347 are migrated. Other handlers
(`scene_handler.gd`, `theme_handler.gd`, `resource_handler.gd`,
`autoload_handler.gd`, `material_handler.gd`, `audio_handler.gd`,
`ui_handler.gd`, `node_handler.gd`) still carry the bare
`begins_with("res://")` pattern; if a future audit finding promotes
them, the validator is already there to delegate to.

Blocks #355 (audit-v2 #11) only inasmuch as #355 references finding #1
(WebSocket auth/Origin) — independent of this PR.

Part of #343.

https://claude.ai/code/session_01LYwesbn3B7LzLvcV2mPQTV

---
_Generated by [Claude Code](https://claude.ai/code/session_01LYwesbn3B7LzLvcV2mPQTV)_

### DIFF
--- a/plugin/addons/godot_ai/handlers/filesystem_handler.gd
+++ b/plugin/addons/godot_ai/handlers/filesystem_handler.gd
@@ -7,11 +7,9 @@ extends RefCounted
 func read_file(params: Dictionary) -> Dictionary:
 	var path: String = params.get("path", "")
 
-	if path.is_empty():
-		return McpErrorCodes.make(McpErrorCodes.INVALID_PARAMS, "Missing required param: path")
-
-	if not path.begins_with("res://"):
-		return McpErrorCodes.make(McpErrorCodes.INVALID_PARAMS, "Path must start with res://")
+	var path_err := McpPathValidator.validate_resource_path(path)
+	if not path_err.is_empty():
+		return McpErrorCodes.make(McpErrorCodes.INVALID_PARAMS, path_err)
 
 	if not FileAccess.file_exists(path):
 		return McpErrorCodes.make(McpErrorCodes.INVALID_PARAMS, "File not found: %s" % path)
@@ -37,11 +35,9 @@ func write_file(params: Dictionary) -> Dictionary:
 	var path: String = params.get("path", "")
 	var content: String = params.get("content", "")
 
-	if path.is_empty():
-		return McpErrorCodes.make(McpErrorCodes.INVALID_PARAMS, "Missing required param: path")
-
-	if not path.begins_with("res://"):
-		return McpErrorCodes.make(McpErrorCodes.INVALID_PARAMS, "Path must start with res://")
+	var path_err := McpPathValidator.validate_resource_path(path)
+	if not path_err.is_empty():
+		return McpErrorCodes.make(McpErrorCodes.INVALID_PARAMS, path_err)
 
 	# Ensure parent directory exists
 	var dir_path := path.get_base_dir()
@@ -87,8 +83,9 @@ func reimport(params: Dictionary) -> Dictionary:
 
 	for path_variant in paths:
 		var path: String = str(path_variant)
-		if not path.begins_with("res://"):
-			not_found.append("%s (must start with res://)" % path)
+		var path_err := McpPathValidator.validate_resource_path(path)
+		if not path_err.is_empty():
+			not_found.append("%s (%s)" % [path, path_err])
 			continue
 		if not FileAccess.file_exists(path):
 			not_found.append("%s (file does not exist)" % path)
--- a/plugin/addons/godot_ai/handlers/script_handler.gd
+++ b/plugin/addons/godot_ai/handlers/script_handler.gd
@@ -25,11 +25,9 @@ func create_script(params: Dictionary) -> Dictionary:
 	var path: String = params.get("path", "")
 	var content: String = params.get("content", "")
 
-	if path.is_empty():
-		return McpErrorCodes.make(McpErrorCodes.INVALID_PARAMS, "Missing required param: path")
-
-	if not path.begins_with("res://"):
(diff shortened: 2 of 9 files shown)


dsarno:
Addressed all three Copilot review findings in `bcd6b0d`:

1. **`test_path_validator.gd` boundary test misnamed** → renamed to `test_well_formed_nested_path_passes_boundary_check` with a comment explaining why no rejection-side input exists for that layer (the substring guard catches every `..` payload before the boundary check is reachable; the boundary check is defence-in-depth for any future encoding bypass).
2. **`test_filesystem.gd` / `test_script.gd` write-traversal disk assertions could false-positive on Unix** (where `/etc/passwd` already exists) → switched both to a synthetic target name (`__mcp_traversal_test_target__`) that won't exist in any clean tree, so the assertion now reliably means "we did not create this file".
3. **`test_path_traversal_guard.py` missed the non-empty layer** → added the `is_empty()` assertion and updated the docstring to drop the misleading "four layers" framing now that all five guards are pinned.

---

dsarno:
CI status (commit `bcd6b0d`): **9/11 green**, only `Godot tests / macOS` and `Godot tests / Windows` failed.

Linux Godot tests passed at 1m17s with the new validator + 18 new tests. The macOS+Windows failures *look* like the known `camera.test_configure_current_sibling_unmark_single_undo` flake tracked in #316 (4th recurrence on macOS, 11 prior fix attempts) — and my changes have no plausible mechanism to fail only on those two platforms (`ProjectSettings.globalize_path` is string substitution; `simplify_path` normalises `\` → `/` and strips trailing slashes consistently across OSes).

I don't have GitHub Actions log access from this environment (the public API rate-limits unauth, the HTML page requires sign-in), so I can't pull the specific failure messages to confirm. **Could you share the failing test name + assertion text from the macOS and Windows job logs?** If it's #316 and friends, that's the audit-v2 prompt's "ONE retry acceptable" exception. If it's something my change actually triggered, the message will pinpoint the test and I'll fix the cause.

---

PR #472, opened 2026-05-27
Fix self-update ZIP backslash test on Windows

### LINKED ISSUE #469: Self-update ZIP backslash rejection test fails on Windows
## Summary

`test_backslash_in_entry_name_is_rejected` fails on Windows during the full Python test suite.

This appears related to the ZIP hardening added in #428.

## Repro

On Windows:

```powershell
$env:HOME=$env:USERPROFILE
uv run pytest -v tests/unit/test_self_update_fixture_extract.py
```

## Failing test

```text
tests/unit/test_self_update_fixture_extract.py::test_backslash_in_entry_name_is_rejected
```

## Observed behavior

The test writes a ZIP member named:

```text
addons/godot_ai/sub\windows-style.gd
```

It expects `extract_addon_from_zip()` to raise:

```text
ValueError: contains backslash
```

On this Windows run, the expected exception is not raised.

## Expected behavior

The test should pass consistently on Windows, or the fixture should be adjusted so the backslash case is tested in a way that survives Windows/zipfile path handling.

Clubhouse1661:
## Summary

Fixes #469.

- Update the self-update ZIP fixture test so it creates a raw backslash archive entry on Windows.
- Check `ZipInfo.orig_filename` as well as `filename`, since Python normalizes `filename` on Windows.

## Tests

- [x] `uv run pytest tests/unit/test_self_update_fixture_extract.py -v`
- [x] `uv run ruff check tests/unit/test_self_update_fixture_extract.py tests/integration/_self_update_fixture.py`

### DIFF
--- a/tests/integration/_self_update_fixture.py
+++ b/tests/integration/_self_update_fixture.py
@@ -279,13 +279,14 @@ def extract_addon_from_zip(zip_path: Path, target_addon: Path) -> None:
         for info in zf.infolist():
             if not info.filename.startswith("addons/godot_ai/") or info.is_dir():
                 continue
+            raw_filename = getattr(info, "orig_filename", info.filename)
             # ZIP spec uses POSIX-style forward-slash separators; parsing the
             # member name as PurePosixPath stops a Windows host from
             # reinterpreting an entry like "addons/godot_ai/C:evil.txt" as a
             # drive component (which would make `target_addon / rel` discard
             # the prefix and write outside the fixture).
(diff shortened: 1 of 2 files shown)


dsarno:
Merged — thanks for the quick follow-up! The added validation in `_write_zip_with_raw_member_name` makes the fixture fail loudly if the byte-level injection ever stops working, which is exactly the right call. Appreciate the clean, focused fix on the Windows backslash guard.

---

PR #491, opened 2026-05-30
Fix opaque 10s game_eval timeout on compile/runtime errors (#490)

### LINKED ISSUE #490: game_eval: opaque 10s INTERNAL_ERROR on compile/runtime errors (separate from #487 hung-await)
# game_eval: opaque timeout on compile / runtime errors — findings

**Status:** root cause confirmed; robust fix NOT shipped (hit deep engine-quirk
walls). Separate from PR #488 (which fixed the *hung-await* case). This is a
**pre-existing** bug, independent of #487/#488.

## The bug

`editor_manage / game_eval` with GDScript that has a **syntax/parse error** or
that throws a **runtime error** returns an opaque `INTERNAL_ERROR` after a
~10–15s timeout, instead of a fast, actionable message. This is likely a
meaningful slice of the F-009 ~19–20% `game_eval` failure rate (agent-written
eval code frequently has typos / null derefs).

## Root cause (verified)

In a debug build (play-in-editor), a GDScript **script error aborts the
executing function and unwinds its whole call stack**:

- **Parse error:** `GDScript.reload()` on bad source aborts the *caller*.
  Proven with an `Engine.set_meta()` side-channel probe: execution reaches the
  line immediately before `reload()` (`"B_before_reload"`) but never the line
  after — `reload()` never returns its error code; it aborts.
- **Runtime error** (e.g. calling a method on `null`): aborts `execute()` and
  unwinds through `_handle_eval`. Same side-channel probe froze at
  `"A_before_deref"`.

`_handle_eval` calls `reload()` / `execute()` synchronously, so the abort kills
it **before** it sends `mcp:eval_response` / `mcp:eval_error`. The request then
rides out to the editor-side 10s fallback timer in
`mcp_debugger_plugin.gd::request_game_eval` → generic `INTERNAL_ERROR`.

The existing `if err != OK:` compile-error branch in `_handle_eval` is
effectively **dead code** in the game context — `reload()` aborts before it can
return a non-OK code. (It works in the *editor* process, which is the context
the handler unit tests run in — which is why this was never caught.)

Reproduced **identically on pre-#487 code** → pre-existing, not from #487/#488.

## What IS available (the levers)

- **Runtime-error text is recoverable in-process.** Runtime errors flow through
  the game-process `Logger` (`runtime/loggers/game_logger.gd::_log_error`),
  including the resolved user line. Confirmed capture (via a temporary
  `Engine.set_meta` from `_log_error`, read back by a later successful eval):
  `"Invalid call. Nonexistent function 'kaboom_method' in base 'Nil'.
  (gdscript://…gd:9 @ _run)"`.
- **Parse-error text is NOT recoverable game-side.** Parse/compile errors go
  straight to the script debugger, bypassing `OS` loggers — `_log_error` does
  **not** fire for them (confirmed: the capture counter stayed at 1 across a
  parse-error eval). The text IS visible in the editor's Output / Debugger
  panel (Godot's built-in remote-debugger error stream).
- **Synchronous `EngineDebugger.send_message` from `_handle_eval` is reliable**
  (normal eval replies prove it).
- **`_handle_eval`'s poll loop (`await tree.process_frame`) is reliable for
  reload-SUCCESS cases** — #487's 8s deadline fires; normal evals complete.

## What does NOT work (dead ends — all tested live)

1. **`game_helper._process` does not run reliably** for this autoload in
   play-in-editor. The existing logger drain lives in `_process`, so
   `logs_read(source="game")` is itself broken/flaky — only the startup
   registration line ever surfaces; eval-context `print`/`push_error` never
   drain. *(Possible separate pre-existing bug worth its own look.)*
2. **A watchdog coroutine started from `_ready`** (an `await process_frame`
   loop): the loop *iterates* (heartbeat counter advanced 2000+), but reads
   `_inflight_evals.size() == 0` while a separate read on the **same instance**
   (verified identical `get_instance_id()`) sees 2 entries. It observes stale /
   inconsistent instance state. Unreliable.
3. **Two coroutines awaiting the same `process_frame` signal, one aborting:**
   the abort breaks the shared signal emission and kills the other coroutine
   (the poll loop) — the request then hangs to the 10s fallback.
4. **Decoupling the risky work into a coroutine that yields on a *separate*
   timer first** (`await create_timer(0.01).timeout`, then `reload()`): normal
   evals work, but a **parse error poisons all subsequent evals** (they hang
   until game restart) — a regression worse than the original bug. The abort
   during the timer-signal resume appears to corrupt message-capture / timer
   processing. **Do not pursue this shape.**

Note: the original #487 structure (reload **inline** in `_handle_eval`) does
**not** poison subsequent evals — a parse error aborts `_handle_eval` cleanly
and the next eval works. Keep reload inline in any future fix.

## Remaining viable approach (designed, NOT built): editor-side beacon protocol

Uses only the two mechanisms confirmed reliable — synchronous sends + the poll
loop (for reload-success) — and keeps `reload()` inline (no poisoning):

- **Game `_handle_eval`**, at the very top, before `reload()`: send
  `mcp:eval_ack [request_id]` synchronously. Then `reload()` inline as today.
  - Parse error / immediate runtime error → `_handle_eval` aborts cleanly after
    the ack (no `alive` beacons follow).
  - Reload success → the poll loop runs; it sends throttled
    `mcp:eval_alive [request_id]` beacons (~every 250ms) and, each frame, checks
    `_logger.last_error_ms()/last_error_text()` to detect + report a **post-await
    runtime error** with its real text.
- **Editor side** (`mcp_debugger_plugin.gd` — a normal, reliable process):
  track `ack` + `last_alive` per request. If acked but **no `alive` within
  ~0.5s** → the game-side function aborted at the synchronous top → reply fast
  with *"eval failed to compile or raised an immediate runtime error — see
  Godot's Output / Debugger panel."* Post-await runtime errors arrive already
  reported (with text) from the poll loop.

**Coverage:** parse + immediate-runtime → fast, generic ("check Output panel");
post-await runtime → fast + real text + line; hung await → #487's 8s; normal →
unchanged.

**Risks:** unbuilt; this environment has produced repeated surprises, so expect
more; beacon throttling + editor-side timing need care; multi-in-flight
attribution should be simplified to the single-in-flight common case.

## To re-pursue

- Re-add `game_logger.gd::last_error_ms()/last_error_text()` (mutex-guarded;
  set in `_log_error`) — needed for the post-await runtime-text path. (These
  were prototyped and reverted.)
- Reliable iteration requires **relaunching the editor** after each game-side
  edit — the running editor does not pick up symlinked-plugin source changes
  for the spawned game (its script cache is stale until relaunch).
- Test inputs:
  - `return 1 +` — parse error
  - `var x = null; return x.foo()` — immediate runtime error
  - `await get_tree().create_timer(0.5).timeout; return broken.thing()` —
    post-await runtime error
  - `push_error("x"); return 5` — **non-fatal**: must still return `5`, not an
    error (guards against over-eager error attribution)
  - `await get_tree().create_timer(999).timeout` — hung await (#487, 8s)
  - `return 1 + 1` — normal

dsarno:
Fixes the opaque ~10s `INTERNAL_ERROR` that `game_eval` returned for GDScript **compile** and **runtime** errors, replacing it with fast, specific, actionable codes. Resolves #490.

**Rebased onto #488** (the hung-await deadline guard, now on main) and integrated with it — see "Composes with #488" below. This PR owns the **compile + runtime** error cases; #488 owns the **hung-await** case. They're complementary.

## Root cause
In a debug build a GDScript error aborts the executing function: `GDScript.reload()` on bad source aborts the caller, and a runtime error aborts `await execute()`. So the game-side `_handle_eval` dies before replying, and the request rides out to the editor-side ~10s timeout → generic `INTERNAL_ERROR`.

Compounding it (verified live): when the play-in-editor game is **backgrounded** — the normal state during agent/MCP-driven evals — its **idle loop freezes**. `_process`, `await get_tree().process_frame`, and game-side `SceneTreeTimer`s stop ticking (`await process_frame` ×15 hung past 10s; a `print()` from a successful eval never drained). Only the debugger capture callback (`_on_debug_message`) stays reliable, and the editor's own loop keeps ticking — so error detection has to be **editor-driven**.

## Fix — three fast signals (reload() kept inline; moving it behind a timer/await poisons subsequent evals)
- **Compile error → `EVAL_COMPILE_ERROR`.** The game sends `mcp:eval_compiled` immediately after `reload()` succeeds. A parse error aborts before that beacon, so an editor-side 3s grace timer fires the error. (Parse text isn't capturable game-side; it's in the editor Output/Debugger panel.)
- **Runtime error → `EVAL_RUNTIME_ERROR` with the real message + resolved line.** `game_logger` counts `ERROR_TYPE_SCRIPT` (type 2) errors and stores the latest text. Once an eval compiles, the editor polls the game every 0.35s with `mcp:eval_check`; the game answers from its debugger capture callback (reliable even when `_process` / the poll loop is frozen) and reports the error when the script-error counter advanced past the pre-eval baseline. A `_process` path is kept as a focused-game fast path. Gating on type 2 keeps `push_error("x"); return 5` returning **5**.
- **Genuine hang →** #488's 8s game-side guard when focused; the editor's 10s fallback timer (refined message) when backgrounded.

## Composes with #488
#488 rewrote `_handle_eval` to drive `execute()` as a fire-and-forget coroutine and poll a deadline via `await tree.process_frame`. Crucially that poll **also freezes when the game is backgrounded** (same root cause), so #488's 8s guard only fires when focused. This PR layers the editor-driven probe on top: the probe reads the logger's script-error counter independently of `_handle_eval`'s coroutine, so runtime errors are reported even while the poll loop is frozen. Timeout ordering preserved: **grace 3s < game 8s < editor 10s < dispatcher 15s**. The merged `_handle_eval` also disambiguates its 8s timeout branch (runtime error → real text vs genuine hang → "Eval exceeded 8s").

## Verified live (backgrounded play-in-editor game; re-verified against the merged code)
| eval | result |
|---|---|
| `return 1 + 1` | `2` |
| `return 1 +` | `EVAL_COMPILE_ERROR` (fast) |
| `var x = null; return x.foo()` | `EVAL_RUNTIME_ERROR` (fast, real text + line) |
| `push_error("x"); return 5` | `5` (guard holds) |
| `await create_timer(999).timeout` | refined 10s hang message |

Known limitation (documented in the handler docstring): an eval that `await`s a timer/signal/frame only progresses while the game window is focused — a backgrounded game's frozen idle loop means such an eval reads as the hang timeout. This is inherent to play-in-editor occlusion, not specific to error handling.

## Tests
- **Python** (`tests/unit/test_game_eval.py`): new error codes exist + propagate via `GodotCommandError`; handler still uses `timeout=15.0`.
- **GDScript** (`test_project/tests/test_game_eval_errors.gd`):
  - *game logger*: counts type-2 (script) errors only — not push_error/warning — and stores the resolved text / backtrace frame.
  - *game helper*: `_try_report_eval_runtime_error` / `_handle_eval_check` fire on counter advance, respect the request-id filter, ignore push_error, and skip a finished eval node.
  - *editor debugger plugin*: `_on_eval_compiled` flips the `compiled` flag; `_on_eval_runtime_error` clears pending + replies `EVAL_RUNTIME_ERROR` with the real text; unknown-request paths don't crash; `_clear_pending` disconnects the grace + probe timers.
- Full suites green locally after the rebase: **1096 Python, 1361 GDScript** (0 failures).

### DIFF
--- a/plugin/addons/godot_ai/debugger/mcp_debugger_plugin.gd
+++ b/plugin/addons/godot_ai/debugger/mcp_debugger_plugin.gd
@@ -154,6 +154,9 @@ func _capture(message: String, data: Array, _session_id: int) -> bool:
 		"mcp:eval_error":
 			_on_eval_error(data)
 			return true
+		"mcp:eval_ack":
+			_on_eval_ack(data)
+			return true
 		"mcp:eval_compiled":
 			_on_eval_compiled(data)
 			return true
@@ -437,21 +440,12 @@ func _send_eval(
 			_log_buffer.log("[debug] !! eval timeout (%s)" % request_id)
 	timer.timeout.connect(timeout_callable)
 
-	## #490: arm the compile-grace timer. If mcp:eval_compiled hasn't set
-	## `compiled` by the time it fires, the eval source failed to parse.
+	## #490: arm the compile-grace timer. _on_eval_grace concludes a parse error
+	## only when the game acked the eval (it received the message and started
+	## reload()) but never sent mcp:eval_compiled — see there for why a missing
+	## ack must NOT be read as a compile error.
 	var grace: SceneTreeTimer = tree.create_timer(EVAL_COMPILE_GRACE_SEC)
-	var grace_callable := func() -> void:
-		var pending_entry = _pending.get(request_id)
-		if pending_entry == null or pending_entry.get("compiled", false):
-			return
-		_clear_pending(request_id)
-		var conn: McpConnection = pending_entry.connection
-		if conn == null or not is_instance_valid(conn):
-			return
-		_send_error(conn, request_id, ErrorCodes.EVAL_COMPILE_ERROR,
-			"Game eval failed to compile — likely a GDScript syntax/parse error. The parse error text is in the editor's Output/Debugger panel; it is not capturable from the running game. Check your eval code's syntax.")
-		if _log_buffer:
-			_log_buffer.log("[debug] !! eval compile error (%s)" % request_id)
+	var grace_callable := func() -> void: _on_eval_grace(request_id)
 	grace.timeout.connect(grace_callable)
 
 	_pending[request_id] = {
@@ -460,6 +454,7 @@ func _send_eval(
 		"timeout_callable": timeout_callable,
 		"grace_timer": grace,
 		"grace_callable": grace_callable,
+		"acked": false,
 		"compiled": false,
 	}
 
@@ -510,6 +505,47 @@ func _on_eval_error(data: Array) -> void:
 	_send_error(connection, request_id, ErrorCodes.INTERNAL_ERROR, message)
 	if _log_buffer:
 		_log_buffer.log("[debug] <- mcp:eval_error (%s): %s" % [request_id, message])
+
+
+## #490: the game sends this at the top of _handle_eval, BEFORE reload() (so it
+## survives a parse-error abort). It positively signals "the game received this
+## eval and started compiling it" — letting _on_eval_grace tell a real parse
+## error (acked, never compiled) apart from a message the game hasn't serviced
+## yet (never acked — main thread blocked by a long frame/load or a CPU-bound
+## prior eval).
+func _on_eval_ack(data: Array) -> void:
+	if data.is_empty():
+		return
+	var request_id: String = data[0]
+	var pending_entry = _pending.get(request_id)
+	if pending_entry == null:
+		return
+	pending_entry["acked"] = true
+	if _log_buffer:
+		_log_buffer.log("[debug] <- mcp:eval_ack (%s)" % request_id)
+
+
+## #490: compile-grace timer fired. Conclude a parse error ONLY when the game
+## acked the eval (started reload()) but never sent mcp:eval_compiled. If it
+## never acked, the game simply hasn't serviced the message yet — NOT a parse
+## error — so leave _pending intact and let the normal eval timeout handle it
+## rather than false-failing a valid eval and dropping its eventual real reply.
+func _on_eval_grace(request_id: String) -> void:
+	var pending_entry = _pending.get(request_id)
+	if pending_entry == null or pending_entry.get("compiled", false):
+		return
+	if not pending_entry.get("acked", false):
+		if _log_buffer:
+			_log_buffer.log("[debug] eval grace: no ack yet, deferring to timeout (%s)" % request_id)
+		return
+	_clear_pending(request_id)
+	var conn: McpConnection = pending_entry.connection
+	if conn == null or not is_instance_valid(conn):
+		return
+	_send_error(conn, request_id, ErrorCodes.EVAL_COMPILE_ERROR,
+		"Game eval failed to compile — likely a GDScript syntax/parse error. The parse error text is in the editor's Output/Debugger panel; it is not capturable from the running game. Check your eval code's syntax.")
+	if _log_buffer:
+		_log_buffer.log("[debug] !! eval compile error (%s)" % request_id)
 
 
 ## #490: the game sends this the instant reload() of the eval source
--- a/plugin/addons/godot_ai/debugger/mcp_debugger_plugin.gd
+++ b/plugin/addons/godot_ai/debugger/mcp_debugger_plugin.gd
@@ -30,6 +30,20 @@ const DEFAULT_TIMEOUT_SEC := 8.0
 ## silent black hole. On CI the game subprocess has been observed
 ## taking ~15s to boot + register.
 const GAME_READY_WAIT_SEC := 20.0
+## #490: how long to wait for the game's mcp:eval_compiled beacon before
+## concluding the eval source failed to compile. A parse error aborts the
+## game-side handler before it can reply, so without this we'd wait the
+## full eval timeout for a syntax mistake. reload() of valid source is
+## sub-millisecond, so 3s is comfortably clear of false positives.
+const EVAL_COMPILE_GRACE_SEC := 3.0
+## #490: once an eval compiles, the editor polls the game every this many
+## seconds with mcp:eval_check. A backgrounded play-in-editor game has a
+## frozen idle loop (no _process / SceneTreeTimer ticks) so it can't
+## self-report a runtime error that aborted the eval — but its debugger
+## capture callback still answers a probe. The editor's own loop keeps
+## ticking, so it drives the poll. 0.35s keeps detection well under a second
+## without flooding the channel; most evals reply before the first probe.
+const EVAL_PROBE_INTERVAL_SEC := 0.35
 
 var _log_buffer: McpLogBuffer
 var _game_log_buffer: McpGameLogBuffer
@@ -139,6 +153,12 @@ func _capture(message: String, data: Array, _session_id: int) -> bool:
 			return true
 		"mcp:eval_error":
 			_on_eval_error(data)
+			return true
+		"mcp:eval_compiled":
+			_on_eval_compiled(data)
+			return true
+		"mcp:eval_runtime_error":
+			_on_eval_runtime_error(data)
 			return true
 		"mcp:game_command_response":
 			_on_game_command_response(data)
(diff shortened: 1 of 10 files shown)


dsarno:
Addressed the Copilot review:

- **Editor-side handler coverage** (`mcp_debugger_plugin.gd:522`): added GDScript tests in `test_game_eval_errors.gd` for `_on_eval_compiled` (flips the `compiled` flag), `_on_eval_runtime_error` (clears pending + replies `EVAL_RUNTIME_ERROR` with the real text, via a stub `McpConnection`), the unknown-request no-crash paths, and `_clear_pending` disconnecting the grace + probe timers. Full GDScript suite green (1361, 0 failures).
- **`_clear_pending` freed-timer guard** (`:354`): no change needed — `SceneTreeTimer` is `RefCounted`, so the reference held in `_pending` keeps it alive after it emits `timeout`; `is_instance_valid()` isn't required here, and this matches the pre-existing pattern for the base `timer`. The new `test_clear_pending_disconnects_grace_and_probe_timers` exercises the path.
- **`_on_eval_error` → `INTERNAL_ERROR` for the game-side compile branch** (`:512`): acknowledged but out of scope for real usage. `game_eval` always runs in a debug build, where `reload()` *aborts* on a parse error (it never returns non-OK), so the `err != OK` branch is effectively dead there and the editor's 3s grace timer produces `EVAL_COMPILE_ERROR`. That branch only fires in the editor process (handler unit tests). Left as-is to avoid fragile message-shape matching on the wire.

dsarno:
Pushed a follow-up commit addressing the two P2 regressions in the runtime-error detector (thanks for the catch):

- **P2a (false-fail on unrelated errors):** the detector no longer keys off a process-wide script-error counter. Each eval's wrapper now has a uniquely named inner function (`_mcp_run_<n>`); the game logger records each script error's backtrace function names; an error is attributed to an eval only if that eval's function appears in the error's stack (past the eval's pre-run baseline seq). A background game error during an in-flight `await` no longer fails the eval. Covered by `test_try_report_ignores_unrelated_game_error`.
- **P2b (overlapping evals):** replaced the single `_eval_request_id`/`_eval_node` slot with a per-request `_inflight_evals` dict, preserving #488's concurrent-eval support; the per-eval token keeps siblings from cross-attributing. Covered by `test_try_report_isolates_overlapping_evals`.

Also dropped the per-frame `_process` runtime hook — the editor probe + #488's in-flight poll loop already cover detection (focused *and* backgrounded), so nothing extra runs on the game's main thread per frame.

Re-verified all 6 live cases against the merged code; full GDScript suite green (1362). The probe itself stays lightweight: editor-driven, fires only while an eval is in-flight past the compile beacon (fast evals get zero probes), and the game-side handler is a mutex-read + at most one message.

dsarno:
Pushed another follow-up for the compile-grace finding — good catch, agreed it was a real false-positive.

**Problem:** the 3s grace inferred a parse error purely from the *absence* of `mcp:eval_compiled`, conflating "source failed to parse" with "the game main thread hasn't serviced the message yet" (a long frame/load, or a CPU-bound prior eval — now possible with concurrent evals). A valid-but-slow-to-service eval was false-failed as `EVAL_COMPILE_ERROR`, and `_clear_pending` then dropped its eventual real `eval_compiled`/`eval_response` as an unknown `request_id`.

**Fix (the `eval_ack` the #490 design originally called for):** the game now sends `mcp:eval_ack` at the **top of `_handle_eval`, before `reload()`** (so it survives a parse-error abort). The editor fires `EVAL_COMPILE_ERROR` only when the eval was **acked but never compiled** — a positive compile-failure signal (game received it, started `reload()`, `reload()` aborted). No ack ⇒ not serviced yet ⇒ the grace defers to the normal timeout and leaves `_pending` intact, so the real reply is still delivered.

Refactored the grace lambda into `_on_eval_grace(request_id)` so the gate is unit-testable. New tests: `test_on_eval_ack_sets_flag`, `test_eval_grace_fires_compile_error_when_acked_but_not_compiled`, `test_eval_grace_defers_when_not_acked` (the fix — no false error, pending intact), `test_eval_grace_noop_when_already_compiled`.

Re-verified live: compile error still → fast `EVAL_COMPILE_ERROR` (confirms the ack flows end-to-end), and the other 5 cases unchanged. Full GDScript suite green (1366).

PR #494, opened 2026-05-30
Fix read-only audit findings (correctness, fragility, dead code, tests, docs)

dsarno:
## Summary

A read-only, adversarially-verified audit of the codebase surfaced **29** findings. This PR fixes **28** of them (one declined — see below). No behavior-defining conventions changed; these are bug fixes, robustness improvements, dead-code removal, comment corrections, and test-assertion tightening.

### Correctness
- **theme** `set_constant` / `set_font_size` now reject non-numeric values instead of silently coercing garbage to `0` (`int("abc")==0` never tripped the existing null guard).
- **signal_handler** type-checks `path`/`signal`/`target`/`method` before `.is_empty()`, so a non-string param returns `WRONG_TYPE` instead of crashing the dispatcher as an opaque "malformed result".
- **script** `find_symbols` now detects `static func` declarations.
- **server_lifecycle** `stop_server` no longer re-appends an already-tracked PID (duplicate kill candidates).

### Fragility
- **material** `apply_to_node` falls back to the in-memory material when the post-save reload returns `null` (was clearing the slot / crashing `get_class()`).
- **project/session** `godot://` resources route through `safe_payload(_sync)` like every other resource module.
- **ci-game-capture-smoke** cleanup calls `project_manage(op="stop")` — `project_stop` is not a named tool, so the game subprocess was never stopped.
- **serve-this-worktree** guards `--port` with no value under `set -u`.
- **test_cli_reload** no longer leaks `GODOT_AI_DEV_*` env vars past teardown.

### Efficiency
- ~~**client_configurator** `check_uv_version` / `_find_system_install` use bounded `McpCliExec` instead of blocking `OS.execute`.~~ **Withdrawn** — `McpCliExec` relies on `OS.execute_with_pipe`, whose capture is unreliable on Godot 4.3 (broke the 4.3 canary). Reverted; `client_configurator.gd` is unchanged in this PR. (Net: 2 findings declined — this and the `_cli_exec` pipe-drain.)

### Dead code
- Remove unused `_resolve_or_create_player` and `_snapshot_curve`, drop the no-op particle alias identity-map, prefix unused `scene_root` locals with `_`.

### Test quality
- Fix a vacuous readiness-gate assertion (asserted a command name never sent).
- Assert returned values (not just key presence) in logs/resource read tests.
- Assert color components in `test_set_param_color_dict`.
- Convert a bare `return` to `skip()` in `test_project`.
- Rename misnamed integration "pagination" tests to round-trip (real pagination is covered in `test_mcp_tools.py` + unit `test_pagination.py`).

### Docs / comments
- Correct stale telemetry milestone/emit-point comment and mark reserved record/milestone types; fix `plugin.gd` logger-build comment, batch history docstring, tools core-tool docstring; rename debugger `_session_id` (it is read in the stale-session guard).

### Repo hygiene
- Remove orphaned `test_telemetry_setting.gd.uid`.

### Declined (1)
`_cli_exec.gd` "drain pipes during the poll loop" — the suggested fix is unsafe: `_drain_pipe` uses `FileAccess.get_as_text()`, which reads to EOF and would **block until the process exits**, defeating the timeout. The bounded-output assumption is already documented at lines 18–20, so the code is left as-is.

## Verification
- `pytest` — **1096 passed, 4 skipped**
- GDScript suite via `test_run` — **0 failures on Godot 4.6.3 (1366 passed)** and **0 failures on Godot 4.3 (1332 passed)**; **0 parse errors** on both engines

### DIFF
--- a/plugin/addons/godot_ai/client_configurator.gd
+++ b/plugin/addons/godot_ai/client_configurator.gd
@@ -504,10 +504,10 @@ static var _uv_version_searched: bool = false
 
 ## Cached for the editor session. The dock's `_refresh_setup_status`
 ## (called via `call_deferred` from `_build_ui`) calls this on the
-## main thread in user mode, so a single cold `uvx --version` shell-out
-## adds ~80 ms to the dock's first paint on Linux and more on Windows.
-## Subsequent calls (focus-in refresh, manual Refresh clicks) reuse the
-## cached string.
+## main thread in user mode, so a single cold `OS.execute(uvx,
+## ["--version"])` adds ~80 ms to the dock's first paint on Linux and
+## more on Windows. Subsequent calls (focus-in refresh, manual Refresh
+## clicks) reuse the cached string.
 ##
 ## Invalidate via `invalidate_uv_version_cache()` when the user
 ## installs / reinstalls uv via the dock so the next refresh reflects
@@ -522,13 +522,9 @@ static func check_uv_version() -> String:
 		_uv_version_searched = true
 		_uv_version_cache = ""
 		return ""
-	## Bounded exec (not blocking OS.execute) so a hung/contended uvx can't
-	## wedge the main thread on the dock's first paint — same rationale as
-	## every other per-client CLI shell-out (#238/#239).
-	var result := McpCliExec.run(uvx, ["--version"], 5000)
-	var text := String(result.get("stdout", "")).strip_edges()
-	if int(result.get("exit_code", -1)) == 0 and not text.is_empty():
-		_uv_version_cache = text
+	var output: Array = []
+	if OS.execute(uvx, ["--version"], output, true) == 0 and output.size() > 0:
+		_uv_version_cache = output[0].strip_edges()
 	else:
 		_uv_version_cache = ""
 	_uv_version_searched = true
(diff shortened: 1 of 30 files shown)


dsarno:
**Follow-up (commit d092594): reverted finding #17.**

The first push failed the `Godot tests / Linux (Godot 4.3)` canary on `test_uvx_server_command_uses_exact_pin_not_tilde` ("0 assertions"). Root cause was the finding #17 change itself: routing `check_uv_version`/`_find_system_install` through `McpCliExec` uses `OS.execute_with_pipe`, whose output capture differs on Godot 4.3 (the documented reason the cli_exec capture tests are skipped on < 4.4). On the canary, `which godot-ai` captured empty → `get_server_command()` resolved to nothing → the test's assert loop never ran. 4.6 captured fine, so only the 4.3 job failed (and my local 4.3 smoke resolved via dev-venv first, masking it).

Finding #17 was a low-severity defensive tweak, so I reverted it rather than break the minimum-supported engine; `client_configurator.gd` is now byte-identical to its pre-audit baseline. The Efficiency item in the description is therefore withdrawn — net 27 findings applied, 2 declined.

PR #495, opened 2026-05-30
Reap orphaned servers when their owning editor dies

dsarno:
## Problem

The plugin spawns the MCP server as a **detached** child (`OS.create_process`) so it survives a plugin reload and can be re-adopted. The only teardown is the editor's clean `_exit_tree → stop_server`. A crash or hard-kill skips that, leaving the server squatting on ports 8000/9500 until a human or the next session's port reconciliation kills it — which is how a fresh session inherits a stale older-version server (the exact thing seen during the audit-PR smoke).

## Fix

Tie the server's life to its owner:

- When the plugin auto-spawns the server it sets **`GODOT_AI_OWNER_PID`** to the editor's PID (`server_lifecycle.gd`).
- The server runs a lightweight watchdog (`orphan_reaper.watch_owner`) that polls that PID and, once the owner editor is gone **and** no editor session is connected, shuts itself down via `SIGTERM` (the graceful path uvicorn already handles).
- The **no-sessions guard preserves adoption**: a different editor that adopted the server holds a live WebSocket session, so the watchdog leaves it running. That adopter's own clean exit (or later crash, by which point sessions are again zero) reaps it.

**Why an env var, not a `--owner-pid` flag, for the plugin→server channel:** an older server in a staggered user-mode upgrade would reject an unknown argparse flag and fail to spawn, whereas it silently ignores an unknown env var. The `--owner-pid` CLI flag still exists for tests/manual use; `main()` prefers it and falls back to the env var.

Servers started **without** an owner (CI's `ci-start-server`, manual `--reload` dev, `uvx` one-shots) never arm the watchdog and behave exactly as before.

## Verification

- **pytest: 1106 passed** — 6 new `orphan_reaper` unit tests (reap-when-dead, no-reap-while-alive, no-reap-when-adopted, reap-after-adopter-leaves, `pid_alive`), owner-pid plumbing tests (flag / env / malformed-env), and updated the 3 `create_server` stubs for the new kwarg.
- **Live end-to-end on Godot 4.6.3**: plugin spawned the worktree server; log showed `Orphan reaper armed for owner editor pid <N>`; `kill -9` of the editor (crash simulation, bypassing `_exit_tree`) reaped the server in ~1s and freed both ports.
- **0 GDScript parse errors** on Godot 4.6.3 and 4.3.

### DIFF
--- a/plugin/addons/godot_ai/utils/server_lifecycle.gd
+++ b/plugin/addons/godot_ai/utils/server_lifecycle.gd
@@ -521,6 +521,14 @@ func start_server() -> void:
 			OS.set_environment("PYTHONPATH", new_pp)
 			pythonpath_set = true
 
+	## Tell the spawned server which editor owns it so it can self-reap if we
+	## die without a clean stop_server (crash / hard-kill). Passed via env, not
+	## a CLI flag, so an older server (staggered user-mode upgrade) silently
+	## ignores an unknown var instead of failing argparse. Left set, not
+	## restored: it's this editor's own constant PID, so it's correct for every
+	## server this editor spawns and harmless to any other subprocess.
+	OS.set_environment("GODOT_AI_OWNER_PID", str(OS.get_process_id()))
+
 	_server_pid = OS.create_process(cmd, args)
 	var spawned_pid := int(_server_pid)
 
--- a/plugin/addons/godot_ai/utils/server_lifecycle.gd
+++ b/plugin/addons/godot_ai/utils/server_lifecycle.gd
@@ -386,6 +386,17 @@ func _inject_telemetry_env() -> bool:
 	return false
 
 
+## Set GODOT_AI_OWNER_PID to this editor's PID for the next OS.create_process,
+## so the spawned server can self-reap if this editor crashes. Returns true if
+## set (caller must unset right after spawning — keep it out of the persistent
+## editor env). No-op on Windows, where the server's reaper is disabled.
+func _set_owner_pid_env() -> bool:
+	if OS.get_name() == "Windows":
+		return false
+	OS.set_environment("GODOT_AI_OWNER_PID", str(OS.get_process_id()))
+	return true
+
+
 ## Branch table (recorded version is the "is this ours?" signal — uvx
 ## launcher PIDs go stale; #135/#137):
 ##   port free                                -> spawn fresh, record PID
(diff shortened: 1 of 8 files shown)


dsarno:
**Follow-up (cfa0846): found and fixed a Windows landmine while smoke-testing harder.**

The liveness probe used `os.kill(owner_pid, 0)`. POSIX: harmless existence check. **Windows: `os.kill` has no signal-0 — any non-CTRL signal calls `TerminateProcess`**, so the first poll would have *killed the owner editor* it was checking (then crashed the reaper task on the next poll). CI never surfaced it because `ci-start-server` starts the server with no owner pid, so the reaper is **never armed in CI on any platform** — my macOS live smoke was the only place it ran.

Fix is conservative: the reaper is now **gated off on Windows** (both `should_arm_reaper()` in Python and the plugin skipping `GODOT_AI_OWNER_PID`). Windows keeps its prior behavior — clean editor shutdown still stops the server; a hard-crash orphan lingers until the next session's port reconciliation, exactly as before. `pid_alive()` was also made non-destructive on Windows (OpenProcess + WaitForSingleObject, never TerminateProcess) as defense-in-depth. Re-verified the macOS reap still works after the gate.

**Coverage honesty:** the *armed* reaper has only run live on macOS. Linux is POSIX-identical (`os.kill(pid,0)`) so low-risk but not live-exercised; Windows is intentionally disabled pending validation on real Windows.

dsarno:
**Review fixes (6648027) — addresses codex's blocking finding + a high-effort self-review.**

- **Owner-env leak (codex blocking):** `GODOT_AI_OWNER_PID` is no longer left set in the editor process env. It's now scoped tightly around each `OS.create_process` (set → spawn → unset) via a shared `_set_owner_pid_env()` helper, in **both** `start_server` and the refresh-retry path (the latter previously relied on the leftover env — fragile). Verified live: the spawned server inherits the var, the editor env does **not** retain it, and the reaper still arms + reaps.
- **Adoption hand-off race (self-review, correctness):** the reaper now requires "owner dead AND zero sessions" to hold on **two samples a grace window apart**, so a transient zero-session blip (an adopter's WebSocket reconnecting across a plugin reload) can't SIGTERM the server out from under the live adopter. New unit test covers it.
- **pid_alive docstring:** corrected the "only delays a reap" claim — a recycled owner PID can cause a permanent missed reap (documented as degrading to the prior port-reconciliation behavior, not an unbounded leak).
- **Integration-test flake:** dedup the two ephemeral ports and poll for port release after exit (TIME_WAIT).

pytest 1111 passed; 0 GDScript parse errors on 4.6.3 + 4.3; reap re-verified live on macOS (server inherits env, editor doesn't, kill→reap in ~poll+grace).

Issue #421, opened 2026-05-11
Allow opting MCP server into LAN access for remote agents (e.g. --allow-host CIDR)


dsarno:
## Background

Both transports are hard-locked to loopback with no escape hatch:

1. **HTTP bind**: `fastmcp.settings.host` (default `127.0.0.1`) in `asgi.py`.
2. **WebSocket bind**: hardcoded `"127.0.0.1"` literal in [`transport/websocket.py:51`](https://github.com/hi-godot/godot-ai/blob/main/src/godot_ai/transport/websocket.py#L51).
3. **Host/Origin allowlist**: [`LocalhostOnlyHTTPMiddleware`](https://github.com/hi-godot/godot-ai/blob/main/src/godot_ai/transport/origin_guard.py) rejects any request whose `Host` header doesn't resolve to `127.0.0.1` / `localhost` / `[::1]` with HTTP 403. The same guard runs on WebSocket upgrade.

So even if you patched the bind to `0.0.0.0`, every remote request would 403 because its `Host` header is the LAN IP, not `localhost`. The guard is there for DNS-rebinding defense and shouldn't go away by default — but there's no way to opt in for users who want remote agents (other machines on a LAN / behind a VPN) doing the heavy lifting against an editor running on a workstation.

Today the only workaround is an SSH tunnel or Tailscale, which works but requires non-trivial setup and isn't discoverable from the dock.

## Proposal

Add an explicit, scoped opt-in — *not* a blanket `--insecure` flag. Something like:

```
python -m godot_ai --transport streamable-http --port 8000 \
  --allow-host 192.168.1.0/24
```

Semantics:
- `--allow-host` takes one or more CIDRs (or bare IPs); repeatable.
- When present:
  - Bind address widens to `0.0.0.0` (or `::`) for both HTTP and WebSocket.
  - `LocalhostOnlyHTTPMiddleware` widens its allowlist to accept any `Host` header whose IP literal falls inside one of the CIDRs (DNS names still require lookup → in-CIDR match, or just reject for simplicity).
  - The Sec-Fetch-Site / Origin checks stay on so browser-driven rebinding still fails.
- When absent: today's loopback-only behavior, byte-for-byte.

Plumbing notes:
- CLI flag in `__main__` argparse, threaded through `create_server` and `run_with_reload` into both the WebSocket server and the ASGI middleware.
[...]

PR #505, opened 2026-05-31
feat(transport): add --allow-host LAN opt-in for remote agents (#421)

dsarno:
## Summary

Implements the server-side core of #421. Both transports are hard-locked to loopback with no escape hatch, so a remote agent on a trusted LAN / VPN can't drive an editor running on a workstation without an SSH tunnel or Tailscale. This adds an explicit, **scoped** opt-in — *not* a blanket `--insecure` switch.

```
python -m godot_ai --transport streamable-http --port 8000 \
  --allow-host 192.168.1.0/24
```

`--allow-host` takes one or more CIDRs / bare IPs (repeatable, or comma-separated). **When present:**
- the HTTP bind (`fastmcp.settings.host`) and the WebSocket bind both widen off loopback to `0.0.0.0`, and
- the DNS-rebinding guard (`LocalhostOnlyHTTPMiddleware` + the WebSocket `process_request` hook) widens its **Host** allowlist to those networks **only**.

**When absent:** today's loopback-only behavior, byte-for-byte (`allowed_networks=None` everywhere).

## Why this is safe (rebinding defense survives the opt-in)

The opt-in widens **only the Host allowlist**. The `Origin` and `Sec-Fetch-Site` rules are deliberately left untouched:

- A **browser** on the LAN sends a non-loopback `Origin` (rejected) and a foreign `Sec-Fetch-Site` (rejected) — so the classic rebinding / liveness-oracle shapes still fail.
- A **native remote agent** (the heavy-lifting use case) sends neither header, so it passes once its Host IP falls inside an allowed CIDR.
- A **DNS name** (the shape a rebinding attack presents in the Host header) never parses to an IP literal, so it can't slip into an allowed network even if it resolves there.

A blanket `--bind 0.0.0.0` was rejected (per the issue) because it's too easy to misuse on untrusted Wi-Fi; CIDR scoping forces the user to name the network they trust, and a bad CIDR **fails loudly at startup** rather than silently exposing nothing (or worse).

## Changes
- **`transport/origin_guard.py`** — `parse_allow_hosts()` (CIDR/bare-IP/comma parsing, `strict=False`), CIDR-aware `is_allowed_host()`, and `allowed_networks` threaded through `evaluate_loopback()`, the ASGI middleware, and the WebSocket guard. Both transports share the one `evaluate_loopback` so they can't diverge.
- **`transport/websocket.py`** — `GodotWebSocketServer` gains `host` + `allowed_networks`.
- **`server.py`** — `create_server(allow_host_networks=...)`; `http_app()` installs the guard with the networks; the lifespan binds the WS server accordingly.
- **`__init__.py`** — `--allow-host` argparse, parse + validate, widen the HTTP bind only when set + only for HTTP transports.
- **`asgi.py`** — reload path carries the CIDRs to the uvicorn factory subprocess via `GODOT_AI_DEV_ALLOW_HOST` and widens the reload bind.

## Scope / follow-up
This is the **server-side core** — the functional, security-critical gate. The two remaining pieces from the issue are intentionally **deferred to a follow-up** (they're additive UI/UX, not security boundaries):
- the developer-mode-gated dock field ("Allow remote hosts (CIDR)") with a warning banner;
- surfacing the LAN URL in the configurator's `manual_command`.

## Testing
- **27 new tests** in `tests/unit/test_origin_guard.py` (parsing, CIDR host matching, loopback-still-passes, DNS-name-never-matches, native-LAN-agent passes, LAN-browser-Origin rejected, middleware opt-in on/off) and `tests/unit/test_cli_reload.py` (flag plumbing into `create_server` / `run_with_reload`, bind widening on/off, invalid-CIDR rejection, reload env).
- Updated existing CLI/runtime fakes for the new kwarg.
- **Full suite: 1135 passed, 2 skipped.** `ruff check` / `ruff format` clean.

https://claude.ai/code/session_01Jq5X4ivngAf1N6r5UX2BVw

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jq5X4ivngAf1N6r5UX2BVw)_

### DIFF
--- a/src/godot_ai/__init__.py
+++ b/src/godot_ai/__init__.py
@@ -86,6 +86,21 @@ def main(argv: Sequence[str] | None = None) -> None:
         ),
     )
     parser.add_argument(
+        "--allow-host",
+        action="append",
+        metavar="CIDR",
+        default=None,
+        help=(
+            "Expose the server to a named LAN range for remote agents (issue "
+            "#421). Takes a CIDR or bare IP (repeatable, or comma-separated), "
+            "e.g. --allow-host 192.168.1.0/24. When set, both transports bind "
+            "off loopback and the rebinding guard widens its Host allowlist to "
+            "these networks ONLY — browser Origin / Sec-Fetch-Site checks stay "
+            "on. Omit for the default loopback-only behavior. Prefer an SSH "
+            "tunnel / Tailscale on untrusted networks; only name ranges you trust."
+        ),
+    )
+    parser.add_argument(
         "--exclude-domains",
         default="",
         help=(
@@ -104,6 +119,23 @@ def main(argv: Sequence[str] | None = None) -> None:
         exclude_domains = parse_exclude_list(args.exclude_domains)
     except ValueError as exc:
         parser.error(str(exc))
+
+    ## #421: parse --allow-host CIDRs. A typo here fails loudly at startup
+    ## rather than silently binding loopback-only (or worse, wide open).
+    from godot_ai.transport.origin_guard import parse_allow_hosts
+
+    try:
+        allow_host_networks = parse_allow_hosts(args.allow_host or [])
+    except ValueError as exc:
+        parser.error(str(exc))
+
+    ## Widen the HTTP bind off loopback only when an allowlist is named. The
+    ## DNS-rebinding guard still gates every request by the CIDR(s); binding
+    ## 0.0.0.0 without the guard would be the footgun this flag avoids.
+    if allow_host_networks and args.transport in ("sse", "streamable-http"):
+        import fastmcp
+
+        fastmcp.settings.host = "0.0.0.0"  # noqa: S104
 
     from godot_ai.runtime_info import install_pid_file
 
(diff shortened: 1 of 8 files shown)


dsarno:
Thanks — both addressed in f60b56d (also merged into `beta-clean` for #506):

1. **WebSocket stays loopback-only.** You're right that widening it was wrong. The WS (9500) is the *local* Python-server↔Godot-editor bridge — the editor connects via `ws://127.0.0.1` and remote agents reach us over HTTP only — so `--allow-host` no longer touches the WS bind. This removes the LAN exposure of the unauthenticated plugin WS and the Windows IPv6-only breakage of the editor's IPv4 loopback connection. (This intentionally deviates from the original #421 "widen both HTTP and WebSocket" note — the WS doesn't need widening for the remote-HTTP-agent use case.)

2. **`bind_host_for_networks()` now prioritizes IPv4 reachability** instead of assuming `::` is dual-stack: any IPv4 network in the allowlist → bind `0.0.0.0` (reachable on every platform, incl. Windows v6-only); `::` only when the allowlist is exclusively IPv6. A mixed-family allowlist no longer silently drops IPv4 reachability.

Tests updated accordingly (`test_bind_host_for_networks_prioritizes_ipv4_reachability`); full suite green.

---

PR #520, opened 2026-06-03
Surface editor error context in log reads

mq1n:
Editor and game logs now preserve structured error details, read visible Debugger Errors rows through logs_read, and clear those rows from clear_logs. Compact responses remain the default; callers opt into rich metadata with include_details.

### DIFF
--- a/plugin/addons/godot_ai/handlers/editor_handler.gd
+++ b/plugin/addons/godot_ai/handlers/editor_handler.gd
@@ -14,6 +14,7 @@ var _debugger_plugin: McpDebuggerPlugin
 var _game_log_buffer: McpGameLogBuffer
 var _editor_log_buffer: McpEditorLogBuffer
 var _debugger_errors_root: Node
+var _debugger_search_root_cache: Node
 
 
 func _init(log_buffer: McpLogBuffer, connection: McpConnection = null, debugger_plugin: McpDebuggerPlugin = null, game_log_buffer: McpGameLogBuffer = null, editor_log_buffer: McpEditorLogBuffer = null, debugger_errors_root: Node = null) -> void:
@@ -196,12 +197,17 @@ func _get_all_logs(count: int, offset: int, include_details: bool) -> Dictionary
 
 
 func _entries_for_response(entries: Array[Dictionary], include_details: bool) -> Array[Dictionary]:
+	## Compact responses only drop the top-level "details" key, so a shallow
+	## copy is enough; the deep copy is reserved for the opt-in details path
+	## where nested dicts leave the buffer.
 	var out: Array[Dictionary] = []
 	for entry in entries:
-		var copy: Dictionary = entry.duplicate(true)
-		if not include_details:
+		if include_details:
+			out.append(entry.duplicate(true))
+		else:
+			var copy: Dictionary = entry.duplicate(false)
 			copy.erase("details")
-		out.append(copy)
+			out.append(copy)
 	return out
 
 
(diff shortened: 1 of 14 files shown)


plugin/addons/godot_ai/handlers/editor_handler.gd
```diff
@@ -736,13 +915,32 @@ func get_performance_monitors(params: Dictionary) -> Dictionary:
...
+	if root == null:
+		root = EditorInterface.get_base_control()
+	if root == null:
+		return 0
+	var trees: Array[Tree] = []
+	_collect_debugger_error_trees(root, trees)
+	var cleared := 0
+	for tree in trees:
+		cleared += _entries_from_debugger_error_tree(tree).size()
+		tree.clear()
```
dsarno: Calling `tree.clear()` directly desyncs the Debugger dock's Errors panel. Godot's own Clear button (`ScriptEditorDebugger::_clear_errors_list()`) does more than clear the tree:

```cpp
error_tree->clear();
error_count = 0;
warning_count = 0;
emit_signal("errors_cleared");
update_tabs();           // resets the "Errors (N)" tab badge
// ... also disables the Clear/Expand/Collapse buttons
```

Clearing only the `Tree` leaves `error_count`/`warning_count` non-zero, so the **"Errors (N)" tab badge stays stale**, the `errors_cleared` signal never fires, and the toolbar buttons remain enabled against an empty tree. There's no public API to invoke `_clear_errors_list()` from the plugin, so reaching into the `Tree` like this can't fully/safely reset the panel.

Two concerns compound here:
1. **Destructive + surprising:** a routine `clear_logs` (a log-*buffer* maintenance call) now wipes the user's visible Errors panel as a side effect. Consider gating the Errors-tab clear behind an explicit opt-in param rather than firing on every `clear_logs`.
2. **UI inconsistency:** if it stays, it shouldn't bypass the engine's counter/tab reset — at present the badge and buttons are left out of sync with the now-empty tree.

This needs a live-editor smoke (per AGENTS.md) with real Debugger errors present to observe the badge/button state after a clear.

---

test_project/tests/test_editor.gd
```diff
@@ -1006,6 +1111,58 @@ func test_get_logs_source_invalid_message_lists_editor() -> void:
...
+	warning_condition.set_text(1, "BUILTIN_SHADOWED")
+	var warning_source := tree.create_item(warning)
+	warning_source.set_text(0, "<GDScript Source>")
+	warning_source.set_text(1, "player.gd:21 @ GDScript::reload()")
+	warning_source.set_metadata(0, ["res://scripts/player.gd", 21])
+	var warning_frame := tree.create_item(warning)
+	warning_frame.set_text(0, "<Stack Trace>")
+	warning_frame.set_text(1, "res://scripts/player.gd:21 @ _ready()")
+	warning_frame.set_metadata(0, ["res://scripts/player.gd", 21])
+
```
dsarno, line 1231: This fixture models stack frames as **flat direct children** of the error item (a single `<Stack Trace>` row). In the real `ScriptEditorDebugger` error tree, there is one `<Stack Trace>` *header* item with the individual frames nested **underneath it** as grandchildren. Since `_details_from_debugger_error_item` only walks direct children (`item.get_first_child()` → `child.get_next()`), live `details.frames` will likely collapse to a single entry instead of the real per-frame list.

The compact path (`level`/`text`/`path`/`line`, sourced from the top item's own `metadata(0)`) is unaffected — this gap is confined to the opt-in `details` payload. Worth verifying against a real editor and, if confirmed, recursing into the `<Stack Trace>` subtree so the fixture and the extraction reflect Godot's actual nesting.

---

dsarno (commented):
## Eval summary

Solid, well-structured work — the reverse-engineering of Godot's debugger internals is faithful (I checked `script_editor_debugger.cpp` on both 4.6 and 4.3: error items really do carry `_is_warning`/`_is_error` meta and `metadata(0) = [source_file, source_line]` on both, so the data model holds across the supported 4.3+ range). All 15 CI checks are green and the Python suite passes locally (1142 passed). A few things to address before merge:

**Blocking**
- **`clear_logs` desyncs the Debugger Errors panel** — calling `tree.clear()` directly skips the engine's counter/tab/signal reset, leaving the "Errors (N)" badge stale and the toolbar buttons enabled against an empty tree. It also turns a log-buffer maintenance call into a destructive wipe of the user's visible Errors panel. Details inline. Consider opt-in gating.
- **No live-editor smoke.** AGENTS.md requires a live smoke for editor-UI features, and the two UI-fidelity issues here (the clear desync, and the stack-frame nesting below) are exactly the class that synthetic-tree GDScript tests can't catch.

**Should fix**
- **`details.frames` is tested against a structure that doesn't match the live tree** — the fixture flattens frames as direct children, but real Godot nests them under a single `<Stack Trace>` header. The compact path is fine; the opt-in `details.frames` likely collapses to one entry live. Inline.

**Copilot's two perf comments are both valid and worth acting on:**
- `_entries_for_response` deep-duplicates every entry even on the default `include_details=false` path, where the only mutation is `erase("details")`. Shallow-copy there; reserve the deep copy for the details branch.
- `_read_debugger_error_entries` walks the entire `get_base_control()` subtree on every `source="editor"`/`"all"` poll — and the clear path does the same walk. Since `logs_read` is a polling/tail tool, caching the located `Tree`(s) (with free-invalidation) or narrowing the search root is the right mitigation.

Nit: `plugin.gd:239` still constructs `EditorHandler` with 5 args, so the new `debugger_errors_root` param is test-injection only and production relies on the `get_base_control()` fallback — which is what makes both the clear path and Copilot's walk hit the full editor tree.

---

dsarno:
@mq1n were you still planning on fixing this Pr

dsarno:
Pushed the review fixes directly to your branch so we can land this — opt-in gating + engine-path reset for the Errors-tab clear, full multi-frame stack extraction, and the two perf items. Thanks for the contribution; happy to walk through any of it if you resurface.

PR #556, opened 2026-06-14
Add GDScript write diagnostics

Clubhouse1661:
## Summary

This PR adds inline GDScript diagnostics to `script_create` and `script_patch`.

Before this change, these tools could successfully write invalid `.gd` files and return `committed: true`; the caller only found out later by running the project or manually reading editor logs. Now each write validates the GDScript content immediately and includes diagnostics in the tool response.

This gives agents a faster feedback loop: write a script, see whether it parses, and fix it before moving on.

## What Changed

- Added a shared diagnostics capture helper.
- `script_create` and `script_patch` now validate `.gd` content with an ephemeral `GDScript` instance.
- Responses now include:
  - `diagnostics`
  - `diagnostics_scope`
  - `diagnostics_status`
  - `diagnostics_detail`
- Diagnostics are scoped to the file being written.
- If the editor log capture window overflows, the response reports `diagnostics_status: "partial"`.
- If Godot reports only the reload error code and does not expose structured log details for the ephemeral script, the response still includes a fallback diagnostic.

## Important Note

On current Godot versions tested here, ephemeral `GDScript.reload()` parse errors do not flow through the editor `Logger` hook into `McpEditorLogBuffer`. Because of that, invalid `.gd` writes currently return `diagnostics_detail: "fallback"`.

That fallback still proves the written content failed validation, but its line number is only a best-effort hint and is marked with `details.fallback_line: true`.

The path to exact line numbers is a follow-up: capture the editor’s real import/scan parse errors after `update_file()` settles, especially for `script_create`, which already has a deferred response path.

## Tests

- Added tests for diagnostics capture path normalization and overflow behavior.
- Added script handler tests for clean writes, invalid writes, fallback diagnostics, and the live logger-backed canary.
- Updated source contract tests for the new `ScriptHandler` constructor wiring.
- Added logger cleanup backstop so the live logger test cannot leak a logger into later suites.

Verified locally:

- Godot import passed
- `test_run suite="script"` passed
- `test_run suite="editor"` passed
- `tests/unit/test_script_create_import_settle.py` passed
- Python unit suite passed earlier on this branch
- `git diff --check` clean

### DIFF
--- a/plugin/addons/godot_ai/handlers/script_handler.gd
+++ b/plugin/addons/godot_ai/handlers/script_handler.gd
@@ -2,11 +2,13 @@
 extends RefCounted
 
 const ErrorCodes := preload("res://addons/godot_ai/utils/error_codes.gd")
+const DiagnosticsCapture := preload("res://addons/godot_ai/utils/diagnostics_capture.gd")
 
 ## Handles script creation, reading, attaching, detaching, and symbol inspection.
 
 var _undo_redo: EditorUndoRedoManager
 var _connection: McpConnection
+var _editor_log_buffer: McpEditorLogBuffer
 
 # Bounded settle window for `ResourceLoader.exists(path)` after `scan()` so
 # that an agent calling create_script -> attach_script back-to-back doesn't
@@ -18,9 +20,10 @@ const _IMPORT_SETTLE_MAX_FRAMES := 300
 const _IMPORT_SETTLE_MAX_MSEC := 3500
 
 
-func _init(undo_redo: EditorUndoRedoManager, connection: McpConnection = null) -> void:
+func _init(undo_redo: EditorUndoRedoManager, connection: McpConnection = null, editor_log_buffer: McpEditorLogBuffer = null) -> void:
 	_undo_redo = undo_redo
 	_connection = connection
+	_editor_log_buffer = editor_log_buffer
 
 
 func create_script(params: Dictionary) -> Dictionary:
@@ -50,6 +53,17 @@ func create_script(params: Dictionary) -> Dictionary:
 	file.store_string(content)
 	file.close()
 
+	var data := {
+		"path": path,
+		"size": content.length(),
+		"committed": true,
+		"import_settled": existed_before,
+		"import_settle": "already_known" if existed_before else "not_waited",
+		"undoable": false,
+		"reason": "File system operations cannot be undone via editor undo",
+	}
+	_attach_gdscript_diagnostics(data, path, content)
+
 	# Register just this file with the editor instead of a full recursive
 	# scan(). A scan() per write stacks `update_scripts_classes` /
 	# `update_script_paths_documentation` WorkerThreadPool tasks under concurrent
@@ -61,15 +75,6 @@ func create_script(params: Dictionary) -> Dictionary:
 	if efs != null:
 		efs.update_file(path)
 
-	var data := {
-		"path": path,
-		"size": content.length(),
-		"committed": true,
-		"import_settled": existed_before,
-		"import_settle": "already_known" if existed_before else "not_waited",
-		"undoable": false,
-		"reason": "File system operations cannot be undone via editor undo",
-	}
 	# `.gd.uid` is the sidecar Godot generates on scan; list both so the caller
 	# can rm the full set in one go.
 	McpResourceIO.attach_cleanup_hint(data, existed_before, [path, path + ".uid"])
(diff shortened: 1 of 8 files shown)


plugin/addons/godot_ai/handlers/script_handler.gd
```diff
@@ -165,6 +170,66 @@ func read_script(params: Dictionary) -> Dictionary:
...
+
+static func _validate_gdscript_source(content: String) -> Dictionary:
+	var script := GDScript.new()
+	script.source_code = content
+	## Keep validation off the live cached resource: assigning resource_path to
+	## this ephemeral Script can collide with loaded instances. reload() still
+	## performs normal GDScript analysis, including static initializer work, so
+	## this check is intentionally scoped to `.gd` writes where the editor would
+	## compile the file on scan anyway.
+	var err := script.reload()
```
dsarno, line 197: `reload()` recompiles the ephemeral script and — per the comment just above — runs static-initializer work. For content that parses cleanly but declares e.g. `static var x = <expr>`, validating a write would *execute* that initializer as a side effect of the write call. Is that intended/safe for agent-authored content? If a parse-only check is sufficient here, is there a lighter path that stops before static init runs? Not blocking — just want to confirm the side-effect surface is deliberate.

---

plugin/addons/godot_ai/handlers/script_handler.gd
```diff
@@ -165,6 +170,66 @@ func read_script(params: Dictionary) -> Dictionary:
...
+				"path": path,
+				"line": line,
+			},
+		},
+	}
+
+
+static func _fallback_gdscript_error_line(content: String) -> int:
+	var lines := content.split("\n")
+	for i in range(lines.size() - 1, -1, -1):
```
dsarno, line 227: The fallback line is the *last non-empty* line, which usually isn't where the error is. In the test fixture `...\n\tif\n\tpass\n` the syntax error is `if` (line 4) but this returns line 5 (`pass`). It's honestly flagged with `fallback_line: true`, so this is minor, but a confidently-wrong line can send a caller to the wrong spot. Worth weighing whether `0`/omitted would mislead less than a guessed line. Follow-up, not a blocker.

---

dsarno (commented):
Solid PR — clean plumbing, strong tests, and the Godot logger limitation (ephemeral `reload()` errors not reaching the buffer → `fallback`) is documented honestly with a canary that asserts it. CI is green and it consumes the 555 cursor API correctly (the overflow → `partial` path is tested end to end).

Two questions inline (both non-blocking): the `reload()` static-initializer side-effect surface, and the best-effort fallback line number.

On Copilot's two findings — I agree with both, and the first is the one I'd actually gate on:

1. **`_entry_matches_target` empty-path misattribution (diagnostics_capture.gd) — real bug, worth fixing before merge.** A pathless top-level entry whose `details.resolved`/`source`/`frames` point at a *different* real file gets pulled into this file's window, and then `_normalize_entry` rewrites its path to the target — actively mislabeling an unrelated diagnostic as belonging to the written file. The capture window is narrow, but the logger fires from any thread, so it's reachable. Rejecting pathless entries whose details indicate a different non-ephemeral path (as Copilot suggests) closes it.

2. **Logger canary missing the `add_logger` availability guard (test_script.gd) — valid.** `plugin.gd` gates this behind `ClassDB.class_exists("Logger")` + `OS.has_method("add_logger")`, and the teardown here already checks `OS.has_method("remove_logger")`, so the add side is asymmetric. Mirroring the production guard avoids a hard failure on a 4.5+ build where the logger API is absent/disabled.

Neither blocks the architecture; #1 is a correctness fix I'd want in before merge, #2 is cheap robustness.

---

Clubhouse1661:
Thanks for the review!

Following your feedback, I made the following changes:

- Tightened `_entry_matches_target` so pathless / `gdscript://` diagnostics are only accepted when their structured details do not point at a different concrete file. This avoids rewriting an unrelated pathless diagnostic onto the file being validated.
- Added a regression test for that misattribution case.
- Mirrored the production logger guard in the live canary test by checking `OS.has_method("add_logger")`.
- Cleaned up the `replace_all` patch fixture so it is valid GDScript and does not incidentally exercise diagnostics.

Verified locally:
- Godot import passed
- `test_run suite="script"`: 41 passed, 0 skipped, 0 failed
- `test_run suite="editor"`: 111 passed, 13 skipped, 0 failed
- `git diff --check` clean

I’m leaving the `reload()` side-effect question and fallback line-number behavior as follow-ups, since both were marked non-blocking and are separate from this PR’s current fallback-based design.

PR #558, opened 2026-06-14
Capture exact GDScript write diagnostics

### LINKED ISSUE #557: Follow-ups from #556: GDScript write-diagnostics fidelity (fallback line + reload() side effects)
Follow-up items deferred (as non-blocking) during review of #556, which is now merged. cc @Clubhouse1661

Both relate to `script_create` / `script_patch` GDScript validation in `plugin/addons/godot_ai/handlers/script_handler.gd`.

### 1. Fallback line number is a best-effort guess that's usually wrong
`_fallback_gdscript_error_line()` returns the **last non-empty line** of the content. That's rarely where the parse error actually is — e.g. for `extends Node\n\nfunc _ready() -> void:\n\tif\n\tpass\n`, the syntax error is `if` on line 4, but the fallback reports line 5 (`pass`).

It's honestly flagged with `details.fallback_line: true`, so a caller *can* know not to trust it, but a confidently-wrong line number can still send an agent to the wrong spot. Options to weigh:
- Report `0` / omit the line rather than emit a guessed one (arguably misleads less).
- ~~Capture the editor's actual import/scan parse errors after `update_file()` settles, so we get exact line numbers instead of a heuristic.~~ **Ruled out** — see update below.

**Update (2026-06-14, from @Clubhouse1661's spike):** The `update_file()` / deferred-settle route does **not** produce logger-visible parse diagnostics. Tested against a live Logger-backed buffer (attach logger → snapshot cursor → write invalid `.gd` with the error on line 4 → `update_file(path)` → poll the settle window and again up to 10s): `appended=0` both times, no entry at all. `update_file()` registers the file but doesn't trigger a logger-visible reparse.

The route that produced output instead was:

```gdscript
ResourceLoader.load(path, "", ResourceLoader.CACHE_MODE_IGNORE)
```

This is the live lead for exact line numbers. Open verification points before it can replace the fallback:
1. Does the captured diagnostic report the **correct line (4)**, not just *an* entry?
2. What **path** is on the entry — real `res://…` or a synthetic/`gdscript://` path (affects `_entry_matches_target` / normalization)?
3. Is capture **synchronous** right after `load()` returns, or does it need polling (i.e. does it fit the existing snapshot → action → `get_since` window)?
4. **Cache safety** on an already-loaded/attached script (the `script_patch` case), and note that loading a *valid* script this way still runs static initializers — so item 2's side-effect concern carries over to this route.

Original thread: https://github.com/hi-godot/godot-ai/pull/556#discussion_r3410092168

### 2. Confirm the `GDScript.reload()` static-initializer side-effect surface is deliberate
`_validate_gdscript_source()` validates by compiling an ephemeral `GDScript` and calling `reload()`. A parse failure returns before execution, but content that *compiles cleanly* and declares e.g. `static var x = <expr>` would have that initializer **executed** as a side effect of the write call.

That's an acceptable surface if intended, but worth confirming — and if a parse-only check is sufficient here, consider a lighter path that stops before static init runs. (Note: the `CACHE_MODE_IGNORE` load route in item 1 shares this surface, since loading a valid script also compiles + runs static init.)

Original thread: https://github.com/hi-godot/godot-ai/pull/556#discussion_r3410092077

---

Neither blocked the merge; filing so they aren't lost. Item 1 (exact line numbers) is the higher-value one since the diagnostic ships a wrong line today.

Clubhouse1661:
## Summary

This follows up on #557 / #556 by upgrading `.gd` write diagnostics from best-effort fallback lines to real parser diagnostics when Godot’s Logger API is available.

#556 added inline diagnostics for `script_create` and `script_patch`, but invalid GDScript usually reported `diagnostics_detail: "fallback"` because ephemeral `GDScript.reload()` errors do not flow through the editor logger. That fallback proves the content failed validation, but its line number is only a guess.

This PR keeps the cheap ephemeral `reload()` validity gate, then adds a second step only when that gate fails:

1. Attach a short-lived private validation logger.
2. Call `ResourceLoader.load(path, "", ResourceLoader.CACHE_MODE_IGNORE)`.
3. Remove the logger immediately.
4. Match captured diagnostics by `details.source.path == target`.
5. Return the real parse message and line as `diagnostics_detail: "log_capture"`.

This works for both `script_create` and `script_patch`.

## Why this approach

I spiked the `update_file()` / deferred-settle route first, but it produced no logger-visible parse diagnostics, even after polling up to 10 seconds. `update_file()` registers the file but does not appear to trigger the logger-visible reparse needed for exact lines.

`ResourceLoader.load(path, "", ResourceLoader.CACHE_MODE_IGNORE)` did produce the needed diagnostics synchronously:

- The real target path is in `details.source.path`.
- The real error line is in `details.source.line`.
- The diagnostic is available immediately after `load()` returns.
- The patch/preloaded-file case did not produce resource-path collision noise.

The top-level log entry path points at the caller/backtrace location, so this PR intentionally pivots matching and normalization to `details.source`.

## Notes

This does not fully resolve #557 item 2. The implementation still uses ephemeral `GDScript.reload()` as the cheap validity gate, so the static-initializer side-effect surface for clean-compiling content remains. The heavier `ResourceLoader.load(..., CACHE_MODE_IGNORE)` path only runs after the content is already known invalid.

## Verification

- `uv run pytest tests\unit\test_script_create_import_settle.py`
- Godot headless import
- `test_run suite="script"`
- `test_run suite="editor"`
- `git diff --check`

### DIFF
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -49,12 +49,13 @@ user-facing UI, so the default leaves it untouched.
 editor import step and include per-write diagnostics in their response:
 `diagnostics` (array of structured editor-style entries), `diagnostics_scope`
 (`"this_file"`), `diagnostics_status` (`"checked"` or `"partial"` if the scoped
-log window overflowed), and `diagnostics_detail`. `diagnostics_detail` is
-`"log_capture"` when entries came from the editor log capture window,
-`"fallback"` when `GDScript.reload()` failed but Godot did not expose structured
-log details for the ephemeral script, and `"none"` when no diagnostics were
-reported. Fallback diagnostics still prove the content failed validation, but
-their line number is a best-effort hint marked with `details.fallback_line`.
+validation log window overflowed), and `diagnostics_detail`.
+`diagnostics_detail` is `"log_capture"` when Godot 4.5+'s Logger API supplied
+real parse diagnostics for the written file, `"fallback"` when validation failed
+but Logger details were unavailable (for example on Godot < 4.5), and `"none"`
+when no diagnostics were reported. Fallback diagnostics still prove the content
+failed validation, but their line number is a best-effort hint marked with
+`details.fallback_line`.
 
 ## Domain rollups (`<domain>_manage`)
 
--- a/plugin/addons/godot_ai/handlers/script_handler.gd
+++ b/plugin/addons/godot_ai/handlers/script_handler.gd
@@ -3,12 +3,12 @@ extends RefCounted
 
 const ErrorCodes := preload("res://addons/godot_ai/utils/error_codes.gd")
 const DiagnosticsCapture := preload("res://addons/godot_ai/utils/diagnostics_capture.gd")
+const LoggerLoader := preload("res://addons/godot_ai/runtime/logger_loader.gd")
 
 ## Handles script creation, reading, attaching, detaching, and symbol inspection.
 
 var _undo_redo: EditorUndoRedoManager
 var _connection: McpConnection
-var _editor_log_buffer: McpEditorLogBuffer
 
 # Bounded settle window for `ResourceLoader.exists(path)` after `scan()` so
 # that an agent calling create_script -> attach_script back-to-back doesn't
@@ -20,10 +20,9 @@ const _IMPORT_SETTLE_MAX_FRAMES := 300
 const _IMPORT_SETTLE_MAX_MSEC := 3500
 
 
-func _init(undo_redo: EditorUndoRedoManager, connection: McpConnection = null, editor_log_buffer: McpEditorLogBuffer = null) -> void:
+func _init(undo_redo: EditorUndoRedoManager, connection: McpConnection = null) -> void:
 	_undo_redo = undo_redo
 	_connection = connection
-	_editor_log_buffer = editor_log_buffer
 
 
 func create_script(params: Dictionary) -> Dictionary:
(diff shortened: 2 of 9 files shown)


dsarno (commented):
This implements the issue #557 plan cleanly, and the verification answers from the spike are all reflected here — real line via `details.source.line`, strict `details.source.path` matching, synchronous capture, and the `reload()`-gate-then-load design so valid writes don't pay the heavier path. Nice to see the #556 pathless/`gdscript://` acceptance heuristic *retired* rather than extended — strict source-path matching is exactly the right simplification and it designs out the misattribution risk that was flagged on #556. CI is green across all platforms and the Godot 4.3 canary.

One thing worth confirming before merge (not an architecture blocker):

**Editor-log pollution.** The validation `ResourceLoader.load()` emits its parse errors to the **global** logger, so in production the *shared* editor `McpEditorLogBuffer` — the one `logs_read(source="editor")` returns — also captures them, not just the private validation buffer. The `validation_logger.gd` comment says the shared editor logger "should still drop these validation-load errors," but that's an assertion without a test here. The entry's `source.path` is the user's file while the *caller* is the addon, so whether the editor logger's feedback-loop filter actually drops it depends on which path the filter keys off. Since invalid writes are common in agent loops, each one would otherwise spam the editor diagnostics buffer that #555/#556 just built. Could you confirm the shared buffer stays clean across a validation load — ideally with a regression test that asserts it?

Minor/acknowledged: the `reload()` gate still runs on clean-compiling content, so issue #557 item 2's static-initializer surface remains (the load path correctly avoids it by only running on already-invalid content). That's fine to track separately as the PR notes.

---

Clubhouse1661:
@dsarno Good call. I added a live regression test for this in `39d803b`.

The test attaches the real shared `editor_logger` (`LoggerLoader.EDITOR_LOGGER_PATH`) to a fresh `McpEditorLogBuffer`, then runs invalid `script_create` and `script_patch` writes through the production `ScriptHandler`. Both writes still return `diagnostics_detail: "log_capture"` from the private validation logger, while the shared editor buffer sees zero new entries during the validation window.

I also added a teardown backstop for the attached logger, including a defensive detach before each attach so a failed earlier test cannot orphan a logger into the rest of the suite.

Verification:
- `git diff --check`: passed
- `uv run pytest tests\unit\test_script_create_import_settle.py`: 6 passed
- Live `test_run suite="script"`: 42 passed, 2 skipped, 0 failed

dsarno:
Confirmed — `39d803b` resolves it. The new `test_*_validation_does_not_pollute_shared_editor_log` tests attach the real production `editor_logger` to a buffer, run an invalid write through `ScriptHandler`, and assert both that the response is still `log_capture` *and* that the shared editor buffer sees zero new entries during the validation window — exactly the behavior I was worried about, now locked in. The detach-before-attach + teardown backstop are a nice touch. Green CI, no other open threads. LGTM. 🚢

---

PR #559, opened 2026-06-14
Add cursor polling for editor logs

Clubhouse1661:
## Summary

This finishes the editor-log cursor work started in #555.

#555 added the monotonic cursor primitive to `McpEditorLogBuffer`, but there was not yet an agent-facing way to use it. This PR exposes that primitive through `logs_read(source="editor", since_cursor=N)`, so agents can poll for new editor diagnostics without rereading the whole editor log buffer.

## What changed

- Adds optional `since_cursor` support to `logs_read(source="editor")`.
- Regular editor-log reads now return `next_cursor` and `appended_total`, so callers can establish a cursor.
- Cursor reads return:
  - `cursor`
  - `oldest_cursor`
  - `next_cursor`
  - `appended_total`
  - `truncated`
  - `has_more`
- Threads `since_cursor` through the Python tool wrapper and handler into `EditorHandler.get_logs`.
- Adds GDScript, Python handler, and MCP integration tests.
- Documents the cursor contract in `docs/TOOLS.md`.

One deliberate design choice: `since_cursor` mode reads only Logger-backed editor entries. Debugger Errors-tab rows are live editor UI state with no stable sequence number, so they cannot be cursor-paged. Regular `logs_read(source="editor")` calls still merge those Debugger rows as before.

## Notes for callers

- First call `logs_read(source="editor")` and save `next_cursor`.
- Later calls can pass `since_cursor=<saved cursor>` to receive only newer Logger-backed editor entries.
- `since_cursor` supersedes `offset`.
- `truncated: true` means the caller fell behind the ring and some entries were evicted before polling. Continue from the returned `next_cursor`; `oldest_cursor` is the earliest retained sequence.
- If the plugin reloads, a stale high cursor self-heals to the new tail with an empty response and corrected `next_cursor`.

## Verification

- `uv run pytest tests\unit\test_runtime_handlers.py -k logs_read`: 9 passed
- `uv run pytest tests\integration\test_mcp_tools.py::TestLogsReadTool`: 6 passed
- `uv run pytest tests\unit\test_tool_domains.py`: 13 passed
- `git diff --check`: passed
- Live `test_run suite="editor"` on Godot 4.6.3: 127 passed, 2 skipped, 0 failed

CI’s Godot 4.3 canary should cover the older-engine/null-buffer path where the Logger API is unavailable.

### DIFF
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -40,6 +40,20 @@ Godot `_log_error` code/rationale when available, error type, resolved source
 location, and stack/error-tree context corresponding to the Debugger dock's
 Errors tab.
 
+For incremental editor-log polling, call `logs_read(source="editor")` once and
+save the returned `next_cursor`; later calls can pass
+`logs_read(source="editor", since_cursor=N)` to receive only Logger-backed
+editor entries appended after that cursor. Cursor responses include
+`cursor`, `oldest_cursor`, `next_cursor`, `appended_total`, `truncated`, and
+`has_more`; `since_cursor` supersedes `offset`. If `truncated=true`, the
+caller fell behind the ring and some entries were evicted before the poll —
+continue from the returned `next_cursor` and treat `oldest_cursor` as the
+earliest retained sequence. If the plugin reloads, a stale high cursor
+self-heals to the new tail with an empty response and a corrected
+`next_cursor`. Live Debugger dock Errors-tab rows are merged into regular
+`source="editor"` reads, but they are UI state rather than ring-buffer entries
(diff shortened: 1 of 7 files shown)


dsarno:
Thanks @Clubhouse1661 — clean completion of the #555 cursor work. `since_cursor` threads through correctly end-to-end (tool → handler → `EditorHandler.get_logs` → `get_since`), and the truncation / Debugger-tab-scoping semantics are well covered across the GDScript, unit, and integration tests. Merging — appreciate the continued contributions! 🙏

PR #560, opened 2026-06-16
Fix: Added static types to iterators in game_logger.gd

straydog-1212:
My Godot setup enforces strict typing (untyped_iterator treated as an Error). Because game_logger.gd is compiled dynamically into memory, it gets a virtual gdscript:// path. This means it isn't protected by the exclude_addons folder filter, causing the engine to block compilation for the whole project. Adding explicit types to the loop iterators fixes the crash!

### DIFF
--- a/plugin/addons/godot_ai/runtime/loggers/game_logger.gd
+++ b/plugin/addons/godot_ai/runtime/loggers/game_logger.gd
@@ -81,9 +81,9 @@ func _log_error(
 		## Collect every function name in the first non-empty backtrace so
 		## game_helper can match its eval's uniquely named wrapper function.
 		var funcs := PackedStringArray()
-		for bt in script_backtraces:
+		for bt:ScriptBacktrace in script_backtraces:
 			if bt != null and bt.get_frame_count() > 0:
-				for i in bt.get_frame_count():
+				for i:int in bt.get_frame_count():
 					funcs.append(bt.get_frame_function(i))
 				break
 		_mutex.lock()
--- a/plugin/addons/godot_ai/runtime/loggers/game_logger.gd
+++ b/plugin/addons/godot_ai/runtime/loggers/game_logger.gd
@@ -71,27 +71,27 @@ func _log_error(
 	var resolved := _LogBacktrace.resolve_error(
 		function, file, line, code, rationale, error_type, script_backtraces,
 	)
-var loc := ""
-    if not resolved.path.is_empty():
-        loc = "%s:%d @ %s" % [resolved.path, resolved.line, resolved.function] if not resolved.function.is_empty() else "%s:%d" % [resolved.path, resolved.line]
-    var text: String = "%s (%s)" % [resolved.message, loc] if not loc.is_empty() else resolved.message
-    var details: Dictionary = resolved.get("details", {})
-    _append(resolved.level, text, details)
-    if error_type == _ERROR_TYPE_SCRIPT:
-        ## Collect every function name in the first non-empty backtrace so
-        ## game_helper can match its eval's uniquely named wrapper function.
-        var funcs := PackedStringArray()
-        for bt: RefCounted in script_backtraces:
-            if bt != null and bt.get_frame_count() > 0:
-                for i: int in bt.get_frame_count():
-                    funcs.append(bt.get_frame_function(i))
-                break
-        _mutex.lock()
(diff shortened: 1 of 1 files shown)


dsarno:
Thanks @straydog-1212 — typing these loops is the right instinct, and I get the goal: keeping `game_logger.gd` warning-clean for users who compile under strict typing, since it's loaded via `logger_loader.gd` rather than being covered by the addon exclude filter.

The blocker is that `for bt: ScriptBacktrace in script_backtraces` is failing CI on Godot 4.4/4.5 — 6 tests in `game_eval_errors` abort with:

```
SCRIPT ERROR: Trying to assign value of type 'RefCounted' to a variable of type 'ScriptBacktrace'.
    at: _log_error (game_logger.gd:84)
```

Those tests inject mock backtraces as plain `RefCounted`, and the strict loop-variable type rejects them at runtime. (4.3 stays green only because the logger isn't dynamically compiled there.)

The inner `for i: int in bt.get_frame_count()` is fine. To land this you'll need to reconcile the `bt` type with the test mocks — either keep `bt` untyped and silence the specific warning with a targeted `@warning_ignore(...)`, or update the `game_eval_errors` mocks to construct objects the `ScriptBacktrace` type accepts. Once CI's green I'll take another look.

dsarno (changes requested):
Thanks for chasing this down — the underlying problem is real. Because `runtime/loggers/` is `.gdignore`'d and `game_logger.gd` is compiled at runtime to a `gdscript://` path, it escapes the addon-exclude filter, so a strict `untyped_iterator = Error` config will block it. That part is a legitimate issue.

However, this specific fix will break the test suite at runtime.

**Blocking: `for bt: ScriptBacktrace` is incompatible with the duck-typed test stub.**

`script_backtraces[i]` entries are Godot's built-in `ScriptBacktrace`, which the engine exposes no constructor for. The test suite therefore feeds in a duck-typed stand-in, `testing/stub_backtrace.gd`, which `extends RefCounted` (see its header comment and the note in `utils/log_backtrace.gd`). `test_project/tests/test_game_eval_errors.gd:62` (and 77, 97, 227) calls:

```gdscript
logger._log_error(..., false, _SCRIPT_ERROR, [StubBacktrace.new("res://eval.gd", 10, fn)])
```

With `error_type == _ERROR_TYPE_SCRIPT`, that hits exactly the loop this PR retypes. GDScript enforces typed loop variables for object types at runtime, so assigning a `StubBacktrace` (a `RefCounted`) to `bt: ScriptBacktrace` raises *"Trying to assign value of type 'RefCounted' to a variable of type 'ScriptBacktrace'"* and every game_logger script-error test fails. `ci-check-gdscript` won't catch this because the folder is gdignore'd (parse-only check), but the runtime test job will.

**Suggested fix:** keep the inner annotation, drop the outer concrete type so duck-typing still works:

```gdscript
for bt in script_backtraces:
    if bt != null and bt.get_frame_count() > 0:
        for i: int in bt.get_frame_count():
            funcs.append(bt.get_frame_function(i))
        break
```

`for i: int` is correct and resolves the inner `untyped_iterator`. For the outer loop, if your strict config still flags the untyped `bt`, annotate it as `RefCounted` (the common supertype of both the real class and the stub) rather than `ScriptBacktrace`. Note the equivalent loop in `utils/log_backtrace.gd:66` is left untyped for the same duck-typing reason — worth staying consistent.

**Minor:** project style is `name: Type` with a space after the colon (`bt: ScriptBacktrace`, `i: int`), not `bt:ScriptBacktrace`.

---

dsarno:
Merged — thanks @straydog-1212! 🎉 Switching the loop type from `ScriptBacktrace` to its base `RefCounted` was exactly the right call: it keeps `game_logger.gd` warning-clean for strict-typed projects (the `untyped_iterator` error is resolved) while still accepting the plain-`RefCounted` mock backtraces the `game_eval_errors` tests inject — so the 4.4/4.5 failure from the earlier revision is gone. All 16 CI checks green. Appreciate the fix and the clear write-up of the dynamic-compilation root cause.

---

dsarno:
👋 Thanks for this @straydog-1212 — it shipped in v2.8.0! 🎉 Join our Discord (https://discord.gg/wpEjdvd24) and drop your handle so we can shout you out in the release announcement. 🚀

PR #562, opened 2026-06-17
fix(runtime): look up EditorInterface by name so exported builds parse game_helper.gd autoload

### LINKED ISSUE #561: game_helper.gd autoload fails to load in exported builds: bare `EditorInterface` identifier rejected by export-template parser
## Summary

The `_mcp_game_helper` autoload (`addons/godot_ai/runtime/game_helper.gd`) references the bare `EditorInterface` singleton, which is **compiled out of export templates**. The GDScript parser therefore rejects the identifier at load time in any exported build, so the autoload fails to instantiate in every export (Web/HTML5, desktop release, etc.).

This contradicts the file's own header docstring, which states it "silently sits idle ... (e.g. exported release builds)".

## Affected code

`addons/godot_ai/runtime/game_helper.gd`, in `_current_scene_root()`:

```gdscript
func _current_scene_root() -> Node:
	var tree := get_tree()
	if tree == null:
		return null
	var scene_root := tree.current_scene
	if scene_root == null and Engine.is_editor_hint():
		scene_root = EditorInterface.get_edited_scene_root()   # <-- bare identifier
	return scene_root
```

The `Engine.is_editor_hint()` guard does **not** save it: the guard is a runtime check, but `EditorInterface` is rejected at **parse** time, which happens before any code runs. The line is unreachable in an export, yet its mere presence breaks parsing.

## Symptom (exported build console)

```
SCRIPT ERROR: Parse Error: Identifier "EditorInterface" not declared in the current scope.
   at: GDScript::reload (res://addons/godot_ai/runtime/game_helper.gd:433)
ERROR: Failed to instantiate an autoload, script '.../game_helper.gd' does not inherit from 'Node'.
```

This then cascades into misleading downstream errors in the user's own game scripts (autoload missing / null references).

## Reproduction

1. Add the Godot AI plugin to a project (enables the `_mcp_game_helper` autoload).
2. Export to Web (HTML5) — or any release export template.
3. Run the exported build and watch the console / browser dev console.
4. Observe the parse error above and the failed autoload instantiation.

Confirmed on **Godot 4.6.3 stable**, reproduced in a Web export. The offending reference is present at HEAD (plugin version 2.7.5) at `game_helper.gd:433`.

## Fix

Reach the singleton by string name so the bare identifier never reaches the parser:

```gdscript
	if scene_root == null and Engine.is_editor_hint():
		var editor := Engine.get_singleton(&"EditorInterface")
		if editor:
			scene_root = editor.get_edited_scene_root()
	return scene_root
```

`Engine.get_singleton(&"EditorInterface")` is a string-name lookup, so the parser sees only a string literal — it parses fine in export templates, and the `is_editor_hint()` guard still gates the actual call to the editor in the editor process. Verified working in the editor and in a Web export.

This is the same class of latent-parse bug as `game_logger.gd` / `McpLogBacktrace` (#557 line); a source-pin unit test analogous to `tests/unit/test_game_logger_no_class_name_lookup.py` prevents regression.

I'll open a PR with the one-line fix plus that regression test.

andperks6:
Fixes #561.

## Problem

`game_helper.gd` is registered as the `_mcp_game_helper` autoload, so it is parsed in exported builds (see 561)

## Fix

Reach the singleton by string name so only a string literal reaches the parser:

```gdscript
	if scene_root == null and Engine.is_editor_hint():
		var editor := Engine.get_singleton(&"EditorInterface")
		if editor:
			scene_root = editor.get_edited_scene_root()
	return scene_root
```

The `is_editor_hint()` guard still gates the actual editor call in the editor process. Same class of latent-parse bug as the existing `game_logger.gd` / `McpLogBacktrace` fix.

## Testing

- Added `tests/unit/test_game_helper_no_editor_interface_identifier.py`, a source-pin test mirroring `tests/unit/test_game_logger_no_class_name_lookup.py`: it fails if a bare `EditorInterface` identifier reappears in code, and requires the `Engine.get_singleton(&"EditorInterface")` form. Verified it passes against the fixed source and fails against the pre-fix source.
- `ruff check` / `ruff format --check` clean on the new test.
- Verified working in the editor and in a Web export (Godot 4.6.3 stable).

### DIFF
--- a/plugin/addons/godot_ai/runtime/game_helper.gd
+++ b/plugin/addons/godot_ai/runtime/game_helper.gd
@@ -430,7 +430,15 @@ func _current_scene_root() -> Node:
 		return null
 	var scene_root := tree.current_scene
 	if scene_root == null and Engine.is_editor_hint():
-		scene_root = EditorInterface.get_edited_scene_root()
+		# Look the editor singleton up by name rather than referencing the bare
+		# `EditorInterface` identifier: that identifier is compiled out of export
+		# templates, so the GDScript parser rejects it ("Identifier
+		# "EditorInterface" not declared in the current scope") in an exported
+		# build even though `Engine.is_editor_hint()` would never run it there.
+		# That parse failure stops this autoload from loading in every export.
+		var editor := Engine.get_singleton(&"EditorInterface")
+		if editor:
+			scene_root = editor.get_edited_scene_root()
 	return scene_root
 
 
--- a/tests/unit/test_game_helper_no_editor_interface_identifier.py
+++ b/tests/unit/test_game_helper_no_editor_interface_identifier.py
@@ -1,0 +1,70 @@
+"""Source-pin: game_helper.gd must not reference the bare `EditorInterface`
+identifier, which is compiled out of export templates.
+
+`game_helper.gd` is registered as the `_mcp_game_helper` autoload, so it is
+loaded (and parsed) in every game process — including exported release builds.
+The bare `EditorInterface` singleton only exists in the editor; it is compiled
+out of export templates. A bare reference is rejected by the GDScript parser at
+load time in an exported build, even when the reference is guarded at runtime
+by `Engine.is_editor_hint()` (the guard never runs because parsing fails first):
+
+    SCRIPT ERROR: Parse Error: Identifier "EditorInterface" not declared in
+    the current scope.
+       at: GDScript::reload (res://addons/godot_ai/runtime/game_helper.gd:...)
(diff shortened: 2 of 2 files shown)


dsarno:
Thanks @andperks6 — great catch, and welcome. This is a real export-build breakage: the bare `EditorInterface` identifier is compiled out of export templates and the parser rejects it at load time, which takes down the `_mcp_game_helper` autoload in every exported build *before* the `Engine.is_editor_hint()` guard can run. `Engine.get_singleton(&"EditorInterface")` is exactly the right fix, and the source-pin regression test is a nice touch to keep it from creeping back. Merging — thanks for contributing! 🙏

Issue #19, opened 2026-04-16
Test artifacts leak into main.tscn when scene is saved mid-test


dsarno:
## Problem

GDScript test suites create temporary nodes and resources prefixed with `_McpTest` (e.g. `_McpTestMat`, `_McpTestAttach`, `_McpTestRemote`) in the live scene during test runs. If the scene gets saved while these exist — manually, via auto-save, or by another MCP tool — their references bake into `main.tscn`.

On next open, Godot shows "Load failed due to missing dependencies" for the cleaned-up temp files (`_mcp_test_material.tres`, `_mcp_test_script.gd`).

This has happened multiple times across branches.

## Proposed fix

Ensure all `McpTestSuite` subclasses clean up `_McpTest*` nodes and temp resources in their teardown. Options:

1. **Auto-cleanup in `McpTestSuite` base class**: after each test method, scan for and delete any child nodes whose name starts with `_McpTest`. This catches leaks even when individual tests fail mid-execution.
2. **Scene-save guard**: before `scene_save`, scan for and warn/remove `_McpTest*` nodes to prevent them from persisting.
3. **Pre-commit hook**: reject `main.tscn` changes that contain `_McpTest` references.

Option 1 is the primary fix. Options 2-3 are defense-in-depth.

PR #563, opened 2026-06-17
Report script errors in test runner and clean tracked allocations

ftab:
## Summary
- Add `McpTestSuite.track()` so tests can register manually managed `Object`s and out-of-tree `Node`s for runner-owned cleanup after each test and suite teardown.
- Add per-test `SCRIPT ERROR` capture in `McpTestRunner`, converting runtime GDScript script errors into structured failed test results like `Aborted by SCRIPT ERROR: ...` instead of letting an aborted test body look passed or ambiguous.
- Guard edited-scene-root cleanup behind an editor-hint helper so runner cleanup is safe in headless/non-editor contexts.
- Add focused Godot-side regression coverage for tracked allocation cleanup.

## Why
Projects using the Godot AI test harness can currently miss a dangerous class of failures: a runtime GDScript `SCRIPT ERROR` may abort the body of a test before assertions or cleanup finish, while the runner still reports through the normal pass/fail path. That is especially risky in CI, where the absence of a structured failed test can make harness regressions look green until a later import or suite run fails noisily.

## Implementation notes
The script-error capture uses Godot's `Logger` API when available, but keeps older editor scans safe by placing the `extends Logger` script under a `.gdignore`d folder and loading it dynamically only after `ClassDB.class_exists("Logger")` and `OS.has_method("add_logger")` pass. The capture script intentionally has no `class_name`.

`track()` ignores `RefCounted` instances because they manage their own lifetime. Plain `Object`s and `Node`s are freed by the runner, with out-of-tree nodes detached from any parent first.

## Testing
Validated locally against upstream `main` at `346259c` on Godot `4.6.3.stable`:

- `git diff --check`
- `bash script/verify-worktree`
- `/Applications/Godot.app/Contents/MacOS/Godot --headless --path test_project --import --quit`
- Focused `test_runner` invocation excluding the editor-scene leak test in headless: `9 passed, 0 failed, 1 skipped`
- Throwaway script-error probe returned `failed=1` with `Aborted by SCRIPT ERROR: ...` as the structured failure message

And last but not least

- `pytest -v`
- `ruff check src/ tests/`

```
======================================================================================= warnings summary =======================================================================================
tests/unit/test_server_status.py:3
  /Users/fury/godot-ai/tests/unit/test_server_status.py:3: StarletteDeprecationWarning: Using `httpx` with `starlette.testclient` is deprecated; install `httpx2` instead.
    from starlette.testclient import TestClient

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
==================================================================== 1168 passed, 4 skipped, 1 warning in 88.59s (0:01:28) =====================================================================
All checks passed!
```


dsarno:
Please run the tests that come with the repo's test project.

ftab:
Sorry I missed that! I thought I had gotten them all. What's the preferred convention, force push for one clean atomic PR commit or keep committing til it works?

dsarno:
> Sorry I missed that! I thought I had gotten them all. What's the preferred convention, force push for one clean atomic PR commit or keep committing til it works?

No worries.  Just push commits to fix it.  Thanks for contributing!

dsarno:
Read through the whole thing now that CI's green — solid work, and thanks for circling back on the test-project run.

- The lazy `script_error_capture_loader.gd` → compile-from-source pattern (gated on `ClassDB.class_exists("Logger")` + the `.gdignore`d folder) mirrors the existing `logger_loader.gd` approach, so it's consistent with how the addon already version-gates Logger subclasses. Good call keeping it out of the static parse path.
- Mutex around the capture path is correct — `_log_error` can fire off-thread, so that matters.
- `track()` / `_free_tracked()` running after every test (not just suite teardown) is a nice defense for the autosave-trap class of bug (#19).

Two optional, non-blocking notes:
1. In `test_script.gd` you call `expect_script_error_containing(...)` twice with the same string — that's correct (the diagnostic is emitted twice), but a one-line comment saying so would save the next reader a double-take.
2. Consider guarding `_unregister_capture()` so a hard abort mid-suite can't leave the logger registered on `OS`. Today it's reached on the normal path; a defensive cleanup would make it bulletproof.

LGTM from me — neither note blocks merge.

ftab:
@dsarno Let me know if that update works for you. I could get more explicit about the cleanup if needed. `_notification(NOTIFICATION_PREDELETE)` and maybe a public `dispose()`/`clear()` cleanup path.

I'm new to GDScript, so I hope I'm doing it right! I didn't see a way to try/catch/finally or anything like that.

dsarno:
This looks great — the constant + helper reads cleaner than what I had in mind, and the `_unregister_capture()` null-guard plus the pre-`_register_capture()` cleanup cover the case I was worried about. Mergeable as-is.

On the `PREDELETE` question — your instinct is spot on, and it matches how the rest of the plugin already handles OS loggers. `plugin.gd` and `runtime/game_helper.gd` both pair `OS.add_logger` / `OS.remove_logger` and do the removal in their Node exit hook (`_exit_tree`). Since `McpTestRunner` is a `RefCounted` rather than a `Node`, the equivalent lifecycle hook is `NOTIFICATION_PREDELETE`:

```gdscript
func _notification(what: int) -> void:
	if what == NOTIFICATION_PREDELETE:
		_unregister_capture()
```

That's the closest GDScript has to a destructor/`finally`, and it'd detach the logger even if a runner is dropped without a clean teardown. It's safe here because nothing the logger holds points back at the runner, so there's no freeing-order trap. Worth adding for consistency with the other two if you'd like — but it's polish, not required.

I'd skip a public `dispose()`/`clear()`: for an internal test helper the register/unregister pairing (plus `PREDELETE` if you add it) is enough — a public method is just another door to the same cleanup.

And yep, you're reading it right: no `try/catch/finally` in GDScript. The idiomatic substitutes are exactly those lifecycle notifications — `NOTIFICATION_PREDELETE` for `RefCounted`/`Object`, `_exit_tree()` for `Node`s. Genuinely clean, well-tested work for a first GDScript PR. 🙂

ftab:
@dsarno Good catch from my last test: in NOTIFICATION_PREDELETE, [calling another instance method on the object can trip a "null instance" error during teardown](https://github.com/godotengine/godot/issues/82841). I had to inline the tiny unregister logic in the notification hook instead of dispatching to _unregister_capture().

dsarno:
Claude, fix all Godot engine bugs. Make no mistakes. 

😉

dsarno:
Excellent work, thank you -- and welcome to the club!

If you have a discord handle and join our discord, I'll give you a shoutout in the next update announcement.

PR #565, opened 2026-06-17
Implement and refactor HTTP proxy support in update manager

AnDong35:

PR #566, opened 2026-06-17
Implement and refactor HTTP proxy support in update manager

AnDong35:
This pull request improves the update manager's handling of HTTP proxy settings in the Godot AI plugin. It introduces a helper function to automatically apply the editor's HTTP and HTTPS proxy settings to all HTTP requests made during update checks and installations, ensuring correct network behavior in environments with proxy requirements.

Network proxy support:

* Added a new `_with_editor_http_proxy` function in `update_manager.gd` to read the editor's proxy settings and apply them to `HTTPRequest` instances before making network requests.
* Updated `check_for_updates()` and `start_install()` to call `_with_editor_http_proxy` on their respective HTTP request objects, ensuring all update-related network traffic respects the user's proxy configuration. [[1]](diffhunk://#diff-26a76f0af6eb69c25621dcf0f0f9acb8567ae83b34958c424e7f11a866330101R89) [[2]](diffhunk://#diff-26a76f0af6eb69c25621dcf0f0f9acb8567ae83b34958c424e7f11a866330101R185)

### DIFF
--- a/plugin/addons/godot_ai/utils/update_manager.gd
+++ b/plugin/addons/godot_ai/utils/update_manager.gd
@@ -86,6 +86,7 @@ func check_for_updates() -> void:
 		_http_request = HTTPRequest.new()
 		_http_request.request_completed.connect(_on_update_check_completed)
 		add_child(_http_request)
+	_with_editor_http_proxy(_http_request)
 	_http_request.request(RELEASES_URL, ["Accept: application/vnd.github+json"])
 
 
@@ -181,6 +182,7 @@ func start_install() -> void:
 	_download_request.max_redirects = 10
 	_download_request.request_completed.connect(_on_download_completed)
 	add_child(_download_request)
+	_with_editor_http_proxy(_download_request)
 	var err := _download_request.request(_latest_download_url)
 	if err != OK:
 		## `request_completed` never fires when `request()` itself errors,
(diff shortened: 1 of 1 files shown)


dsarno:
Thanks for working on proxy support — reading the editor's configured proxy from `EditorSettings` and applying it to the update-check and download requests is a reasonable feature, and pulling it from settings (rather than anything hardcoded) is the right approach.

Two things before this can go in:

1. **`_with_editor_http_proxy()` won't parse.** Every line of the body after `var es` is indented with a leading space *before* the tab, so GDScript rejects it with a mixed tabs/spaces error and the whole `update_manager.gd` fails to compile. Please re-indent the function body with tabs only (matching the rest of the file) and confirm it loads in the editor.
2. **Please consolidate to a single PR.** This is the third PR for the same change (#564, #565, #566). Going forward, just push new commits to the same branch — the open PR updates automatically, no need to open a new one each time.

A small test exercising the empty-vs-set proxy paths would be welcome too, if feasible. Once it compiles and CI is green I'll review the behavior in detail.

dsarno (changes requested):
The approach is sound — `HTTPRequest.set_http_proxy()` / `set_https_proxy()` are the right APIs, reading the editor's `network/http_proxy/*` settings is correct, and the "always apply so a reused `HTTPRequest` doesn't keep a stale proxy" rationale is a nice touch. The `has_method` guards are reasonable defensive coding.

**Blocking: the new function body uses mixed tab/space indentation and won't parse.**

In `_with_editor_http_proxy`, the first line (`var es := ...`) is indented with a tab, but every line after it is indented with a *space followed by a tab*:

```
func _with_editor_http_proxy(request: HTTPRequest) -> void:
	var es := EditorInterface.get_editor_settings()
 	if es == null:        # <- space + tab
 		return            # <- space + tab
```

GDScript rejects mixed tabs and spaces with a parse error (*"Used space character for indentation instead of tab"* / inconsistent indentation). Unlike `game_logger.gd`, `utils/update_manager.gd` is parsed at import, so this will fail `script/ci-check-gdscript` and the plugin won't load. Re-indent the whole function body with tabs only.

**Minor:** the comment uses a curly apostrophe (`don't`); the codebase generally sticks to ASCII in source. Trivial, fix while you're re-indenting.

Once the indentation is fixed this should be good to go.

---

Clubhouse1661:
One additional note now that `main` has moved: self-update also fetches the `.sha256` checksum sidecar via `_verify_request` in `_verify_then_install()`. This PR applies the proxy to the release check and ZIP download, but not the checksum request, so a proxied update could still fail during verification. Please apply the proxy helper there too when rebasing/updating the branch.

dsarno:
Will close tomorrow if no action from submitter.

dsarno:
Close due to submitter inactivity.

PR #568, opened 2026-06-19
add Godot 4.7 row and harden input device IDs

Clubhouse1661:
**Description**
This is the first Godot 4.7 adoption PR. It adds 4.7 as a CI signal without promoting it to the primary supported/tested version yet.

Changes:
- Add a Linux Godot `4.7.0` row to the `godot-tests` matrix.
- Keep the existing 4.6.2 primary rows and the 4.3 canary unchanged.
- Set synthesized keyboard and mouse InputMap bindings to `device = -1` so they match events from any keyboard/mouse device under Godot 4.7’s new device-ID behavior.
- Leave joypad bindings unchanged because joypad events already use explicit device IDs.
- Add regression tests that assert both the stored device value and real `InputMap.event_is_action(...)` matching behavior.
- Pin invalid-write diagnostic tests to exactly one returned diagnostic, guarding against duplicate user-facing diagnostics on 4.7.

Validation:
- Live Godot 4.7 full `test_run`: `1426 passed, 21 skipped`
- `test_run suite="script"`: `42 passed, 2 skipped`
- `script/ci-check-gdscript`: passed
- `ruff check src/ tests/`: passed
- `git diff --check`: passed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,6 +144,9 @@ jobs:
           - os: windows-latest
             label: Windows
             godot-version: "4.6.2"
+          - os: ubuntu-latest
+            label: Linux (Godot 4.7)
+            godot-version: "4.7.0"
           ## 4.3 Linux canary — README advertises "Godot 4.3+" support and
           ## #476's parse-cascade regression went unnoticed for 24 days
           ## because every CI run was on 4.6.2. Phase 1 of #477: one cheap
--- a/plugin/addons/godot_ai/handlers/input_handler.gd
+++ b/plugin/addons/godot_ai/handlers/input_handler.gd
@@ -175,6 +175,7 @@ func _create_event(event_type: String, params: Dictionary):
 			ev.alt_pressed = params.get("alt", false)
 			ev.shift_pressed = params.get("shift", false)
 			ev.meta_pressed = params.get("meta", false)
+			ev.device = -1
 			return ev
 		"mouse_button":
 			if not params.has("button"):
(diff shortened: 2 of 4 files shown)


dsarno (commented):
Looks good overall — this is a clean way to add 4.7 as a CI signal without promoting it. A few notes, all non-blocking:

**`device = -1` (the core change):** correct. Under 4.7's device-sensitive `InputMap` matching, synthesized key/mouse events need `device = -1` to match events from any keyboard/mouse, and leaving joypad bindings on explicit device IDs is right. The new tests assert both the stored value *and* real `InputMap.event_is_action(...)` matching for default and explicit devices, which is exactly the behavior that matters.

**The triple `expect_script_error_containing` in `_expect_invalid_if_parse_errors()`:** I verified this is safe on older engines. `test_suite._unexpected_script_errors()` only flags *captured-but-unexpected* errors; surplus expectations are left unmatched in `remaining` and silently discarded. So declaring 3 when 4.6.2/4.3 emit 2 produces no failure, and 4.7's 3rd copy is absorbed exactly. The comment is accurate.

**`assert_eq(result.data.diagnostics.size(), 1)`** (replacing `assert_gt(..., 0)`): this tightens the user-facing dedup'd diagnostics to exactly 1. The reasoning is sound, but unlike the harness behavior above I can't statically prove the dedup layer yields exactly 1 on every matrix row — that's what the 4.3 canary + 4.6.2 + 4.7 jobs will confirm. Just flagging it as the one assertion whose cross-version correctness rests on CI rather than on a guarantee in the code.

Nothing here needs to change before merge from my side; deferring final approval to the matrix going green on 4.3/4.6.2/4.7.

---

PR #572, opened 2026-06-20
fix: explain missing Godot editor sessions

Clubhouse1661:
Fixes the opaque “No active Godot session” failure mode.

When the MCP server is running but no Godot editor is connected to that same server, tool calls used to fail with a bare no-session message. That made Docker/remote-agent/topology issues look like product bugs or regressions.

This PR makes the no-session path self-diagnosing.

Changes:
- Return a structured `PLUGIN_DISCONNECTED` error when a tool call has no active Godot editor session.
- Add shared no-session diagnostics with:
  - `connected: false`
  - `reason: "no_active_session"`
  - `retryable: true`
  - a `session_manage(op='list')` check
  - Docker/loopback namespace guidance
- Apply the same diagnostic payload to:
  - normal tool calls through `GodotClient.send`
  - `editor_reload_plugin`
  - `godot://project/info`
- Make circuit-open errors caused by repeated no-session failures mention the root cause in the message, not only in structured data.
- Add unit and integration coverage for direct client, FastMCP tool, and resource paths.

Note: resources still return an error-shaped JSON object rather than raising MCP tool errors; that preserves the existing tool-vs-resource convention.

**Validation**
- `uv run ruff check src/ tests/`
- Targeted no-session tests: `18 passed`
- Circuit/message follow-up tests: `14 passed`
- `uv run pytest -q -k "not test_pid_alive_for_self and not test_pid_alive_false_for_dead_process"` → `1167 passed, 9 skipped, 2 deselected`
- `git diff --check`

Full unfiltered Windows pytest is blocked by the two pre-existing POSIX-only orphan-reaper tests tracked separately in #570.


dsarno (approved):
Nice change — turning the bare `ConnectionError("No active Godot session")` into a structured `PLUGIN_DISCONNECTED` error with self-diagnosing data is a clear UX improvement, and the shared `session_diagnostics` helper keeps the three call sites consistent.

I checked the main correctness risk — that swapping `ConnectionError` → `GodotCommandError` could perturb circuit-breaker accounting — and it's clear:

- In `client.send()` the no-session raise sits at the top of the method (the `session is None` branch, `client.py:117-120`), *above* the transport `try/except (ConnectionError, TimeoutError)` at line ~145. So it was never caught by that handler, and the failure is recorded explicitly via `_record_failure(None, kind="no_active_session")` immediately before the raise. No accounting changes.
- `editor_reload_plugin`'s raise is likewise above its own `except (ConnectionError, TimeoutError)` blocks.
- The circuit-open path correctly merges `no_active_session_data` only when `session_id is None and last_failure_kind == "no_active_session"`, so per-session transport failures still get the generic message.

Test coverage across direct client / FastMCP tool / resource paths is thorough, and the resource still returns an error-shaped object rather than raising, preserving the tool-vs-resource convention.

One nit (non-blocking): the new messages use an ASCII hyphen (`" - "`) while the surrounding code uses an em dash (`—`) — e.g. the adjacent circuit-open message. Worth aligning for consistency. Also note `session_not_found` (`client.py:134`) still raises a bare `ConnectionError`; out of scope here, just flagging it as the remaining inconsistent path if you want a follow-up.

---

Clubhouse1661:
@dsarno Thanks for the review!
I pushed the requested follow-ups and rebased onto current `main`.

Changes since the previous review:
- Updated the circuit-open no-session message so callers that only read `message` still see the missing-editor root cause.
- Aligned the new user-facing separators with the existing em dash style.
- Converted missing pinned `session_id` failures from a bare `ConnectionError` into `PLUGIN_DISCONNECTED` with `data.reason = "session_not_found"` and actionable `session_manage(op="list")` guidance.
- Added the same structured payload to the repeated/circuit-open missing-session case.
- Removed the now-unused `SESSION_NOT_FOUND` enum member, since that old path only interpolated it as text and no structured wire response emitted it.

Re-ran focused checks after rebasing: ruff passed, focused pytest passed (`20 passed`), and `git diff --check` is clean.

Clubhouse1661:
Both of Copilot's points addressed in `20d9fea`.

- Restored `ErrorCode.SESSION_NOT_FOUND` as a compatibility-reserved enum member, while keeping the new emitted shape as `PLUGIN_DISCONNECTED` with `data.reason == "session_not_found"`.
- Removed the “make sure a scene is open” wording from the no-session message, since the editor can connect before a scene is open.

Focused tests are green: 563 passed.

Issue #581, opened 2026-06-20
MCP signal connections are dropped on scene save (missing Object.CONNECT_PERSIST)


cpalaka:
### Summary

Signals connected through the server (`signal_manage op=connect`) are made without `Object.CONNECT_PERSIST`, so the engine treats them as runtime-only and never writes them into the scene's `[connection]` block. The connection works in the live session — `op=list` confirms it — but it's gone the next time the scene is saved and reloaded.

It's a silent loss: nothing in the response, the readback, or the save flags a problem. An agent wiring up a `Button.pressed` handler sees every check pass, saves, and ships a `.tscn` with no connection in it. The method just never fires.

### Reproduction

1. Open a scene with two scene-owned nodes (both `owner` set to the scene root — otherwise the connection can't serialize regardless of this bug).
2. `signal_manage op=connect` — e.g. source `ready` → target `queue_free`.
3. Response is success, and `op=list` shows the connection.
4. Save, then check the `.tscn`: no `[connection ...]` entry. (Or: `PackedScene.pack(root).get_state().get_connection_count()` doesn't go up.)
5. Reload — the connection is gone.

### Root cause

In `signal_handler.gd` (current `main`, `0c88e4b`), `connect_signal` registers the connect with no flags:

```gdscript
# :155
_undo_redo.add_do_method(source, "connect", signal_name, callable)
```

`connect()` defaults to flags `0`, and only `CONNECT_PERSIST` connections get serialized into a packed scene (it's the same flag the editor's signal dialog uses). Without it the connection is runtime-only and dropped on save. `disconnect_signal` has the mirror bug at `:179` — its undo re-connects, also flagless — so undoing a disconnect restores a runtime-only connection too.

### Proposed fix

Add `Object.CONNECT_PERSIST` to both connect calls — `connect_signal`'s do-method (`:155`) and `disconnect_signal`'s undo-method (`:179`). Two lines, one argument each; nothing else changes. (The connection still only serializes when both nodes are owned by the edited scene — the flag is the part that was missing, not the only requirement.)

---

Related, filed separately: `node_set_property` silently no-ops struct-valued types (#579), and `resource_create` can't make custom `class_name` Resources (#580).

dsarno:
Nice ones, thanks. Any interest in submitting PR's for these?

cpalaka:
yup, they're on their way!

Issue #580, opened 2026-06-20
resource_create and set_property {"__class__": ...} fail for custom class_name Resources with "Unknown resource type"


cpalaka:
### Summary

`resource_create` and the inline `node_set_property` shortcut `{"__class__": "MyThing", ...}` can't instantiate a project's own `class_name` Resource — they return `Unknown resource type: <name>`. Both resolve the type through `ClassDB`, which only knows engine built-ins, so a `class_name MyThing extends Resource` is invisible and the call fails before anything is created.

Unlike the struct-write siblings, this one fails honestly — a real error, not a silent no-op. But custom `class_name` Resources are the normal way to define typed data assets in Godot (stats, item defs, configs), and there's currently no way to make one through the server. You end up hand-writing `.tres` files.

### Reproduction

1. Add a resource script and let the editor register the global class (reopen or rescan):

   ```gdscript
   @tool
   class_name MyThing
   extends Resource

   @export var label: String = ""
   ```

2. `resource_create` with `type: "MyThing"`, `resource_path: "res://my_thing.tres"`, `properties: {"label": "hi"}` → `Unknown resource type: MyThing`, no file written.
3. The inline path fails the same way: `node_set_property` on a Resource property with `value: {"__class__": "MyThing", "label": "hi"}` → same `Unknown resource type: MyThing`, the property left unchanged.

The same calls with a built-in (`BoxMesh`) work, so it's specifically the `ClassDB`-only lookup.

### Root cause

Both paths go through `_validate_resource_class` in `resource_handler.gd` (current `main`, `0c88e4b`) at `:207-209`, which returns `Unknown resource type` whenever `ClassDB.class_exists(type_str)` is false — and `ClassDB` lists engine classes only. `create_resource` (`:181`, `:185`) and the `set_property` `{"__class__": ...}` branch in `node_handler.gd` (`:242`, `:245`) both validate-then-`ClassDB.instantiate`, so a custom class is rejected before it's ever instantiated.

The registry that *does* know these types — `ProjectSettings.get_global_class_list()` — is never consulted.

### Proposed fix

Add an `_instantiate_resource(type_str)` helper in `resource_handler.gd`: keep the `ClassDB` path for built-ins, and fall back to `get_global_class_list()` for project classes (`load()` the script, check `can_instantiate()`, `new()`, confirm it's a `Resource`). Route both `create_resource` and the node `__class__` branch through it. Built-ins are unchanged, and the wire format and schemas don't move.

---

[...]

PR #582, opened 2026-06-20
Coerce struct-valued property writes (Rect2, Transform, Vector4, …) instead of silently no-op-ing

### LINKED ISSUE #579: node_set_property silently no-ops struct-valued types (Rect2, Transform2D, Vector2i, …) and reports success
### Summary

`node_set_property` returns success for a range of struct-valued types but never writes the value. The property keeps its old value, and nothing lands in the saved `.tscn` — a silent no-op that looks like it worked.

There's no error to react to, and the response echoes the property back at its *old* value, so a quick read looks fine too. You only catch it by saving and grepping the `.tscn`. For struct-heavy work (transforms, region rects, rigs, typed resources) that makes the tool unusable — you fall back to hand-editing `.tscn`.

(The struct sub-properties of an inline `{"__class__": ...}` resource hit the same bug, e.g. `BoxMesh.size`, since they go through the same coercion code.)

### Affected types

These 13 compound types have no coercion branch and silently no-op:

`Vector2i`, `Vector3i`, `Vector4`, `Vector4i`, `Quaternion`, `Rect2`, `Rect2i`, `AABB`, `Plane`, `Basis`, `Transform2D`, `Transform3D`, `Projection`

`PackedVector4Array` has the same gap — the read serializer emits it, but `_coerce_value` has no branch for it — so it no-ops the same way.

`Vector2`, `Vector3`, `Color`, and the other `Packed*Array` types already coerce and work — this doesn't touch them. (`PackedByteArray` is serialized but deliberately left un-coerced: base64-vs-int is an open design choice.)

### Reproduction

1. Add a `Sprite2D`.
2. `node_set_property` with `property: "region_rect"`, `value: {"position": {"x": 0, "y": 0}, "size": {"x": 6, "y": 6}}`.
3. Response is success; the echoed `value` comes back as the old `region_rect`.
4. The property is unchanged, and `grep region_rect <scene>.tscn` after saving finds nothing.

Same for `Node2D.transform`, a `Vector2i`, `Bone2D.rest`, and any other property typed as one of the 13.

### Root cause

In `node_handler.gd` (current `main`, `0c88e4b`): `_coerce_value` (`:657-762`) has no `match` case for these types, so the raw JSON dict falls through un-coerced. `_check_coerced`'s `_:` default (`:571-572`) then returns `null` ("no error"), so `set_property` assigns a `Dictionary` to a hard-typed property — the engine refuses the conversion, keeps the old value, and the response echoes that old value back.

The read serializer (`variant_serializer.gd:7-71`) already defines a round-trippable JSON shape for all 13 types, so the write is just its inverse — no new format to design. (That same coercion pair is reused for resource sub-properties via `_apply_resource_properties`, which is why the `__class__` shortcut is affected.)

### Proposed fix

Two small changes in `node_handler.gd`:

- Make `_check_coerced`'s default reject un-coercible values (still passing `null`, untyped, and already-correct ones), so unsupported types fail honestly instead of no-op'ing.
- Add the inverse-serializer cases to `_coerce_value` — the 13 compound types plus `PackedVector4Array` — with the matching `is`-checks and shape hints. Coercion recurses, so nested shapes come for free. `PackedByteArray` stays out — base64-vs-int is an open design choice (JSON has no native byte type).

---

Related, filed separately: custom `class_name` Resources can't be instantiated (#580), and signal connections drop on save without `CONNECT_PERSIST` (#581).

cpalaka:
Fixes #579.

`node_set_property` silently no-op'd struct-valued types: `_coerce_value` had no branch for them, so an un-coerced dict was assigned to a hard-typed property — the engine rejected it, the old value stuck, and the response echoed the old value back with no error.

Three commits:

- **fix(node): return honest WRONG_TYPE for un-coercible property writes** — `_check_coerced`'s default now rejects un-coercible values (still passing `null`, untyped, and already-correct ones), so unsupported types fail honestly instead of no-op'ing.
- **feat(node): coerce struct-valued property writes** — adds the 13 inverse-serializer cases to `_coerce_value` (`Vector2i`, `Vector3i`, `Vector4`, `Vector4i`, `Quaternion`, `Rect2`, `Rect2i`, `AABB`, `Plane`, `Basis`, `Transform2D`, `Transform3D`, `Projection`) plus the matching `is`-checks and shape hints. Coercion recurses, so nested shapes come for free.
- **feat(node): coerce PackedVector4Array property writes** — it was serialized on read but had no coercion branch (same gap); now mirrored alongside the other `Packed*Array` types. `PackedByteArray` stays the only serialized type intentionally left un-coerced (base64-vs-int design choice).

The read serializer already defined the JSON shape for all of these, so each write is just its inverse — no new wire format.

Tests: per-type coercion, honest-error-with-shape-hint on bad input, and round-trips. Full GDScript suite green (1452 passed / 0 failed / 7 skipped, 55 suites); ruff + pytest unaffected (no Python touched).

Sibling fixes filed separately — #580 (custom `class_name` Resources) and #581 (signal persistence); independent, different files, PRs to follow.

## Summary by CodeRabbit

* **New Features**
  * Expanded automatic type coercion to support more Godot math/geometry types (including packed Vector4 arrays), converting structured inputs into the correct runtime types.
  * Improved validation and coercion error feedback so unsupported target types report clearer “wrong type” results.

* **Tests**
  * Added/extended unit and end-to-end tests covering packed Vector4 array coercion, structured dict-to-geometry conversions, failure cases, and real property assignment with undo verification.

### DIFF
--- a/plugin/addons/godot_ai/handlers/node_handler.gd
+++ b/plugin/addons/godot_ai/handlers/node_handler.gd
@@ -528,6 +528,7 @@ func _set_owner_recursive(node: Node, owner: Node) -> void:
 ## is optional — the coercer defaults it to 1.0 when absent.
 const VECTOR2_KEYS: Array[String] = ["x", "y"]
 const VECTOR3_KEYS: Array[String] = ["x", "y", "z"]
+const VECTOR4_KEYS: Array[String] = ["x", "y", "z", "w"]
 const COLOR_KEYS: Array[String] = ["r", "g", "b"]
 
 
@@ -568,6 +569,19 @@ static func _check_coerced(value: Variant, target_type: int, prefix: String = ""
 			ok = value is PackedFloat64Array
 		TYPE_PACKED_STRING_ARRAY:
 			ok = value is PackedStringArray
+		TYPE_VECTOR2I: ok = value is Vector2i
+		TYPE_VECTOR3I: ok = value is Vector3i
+		TYPE_VECTOR4: ok = value is Vector4
+		TYPE_VECTOR4I: ok = value is Vector4i
+		TYPE_QUATERNION: ok = value is Quaternion
+		TYPE_RECT2: ok = value is Rect2
+		TYPE_RECT2I: ok = value is Rect2i
+		TYPE_AABB: ok = value is AABB
+		TYPE_PLANE: ok = value is Plane
+		TYPE_BASIS: ok = value is Basis
+		TYPE_TRANSFORM2D: ok = value is Transform2D
+		TYPE_TRANSFORM3D: ok = value is Transform3D
+		TYPE_PROJECTION: ok = value is Projection
 		_:
 			# null / untyped-TYPE_NIL / already-correct-type are handled by
 			# Godot's setter; anything else would silently no-op, so error.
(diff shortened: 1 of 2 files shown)


dsarno:
@Coderabbitai review

cpalaka:
Addressed the per-type coverage nitpick in `61f5c3e`: added dedicated `_coerce_value` unit tests for the 8 compound types that lacked them — Vector3i, Vector4i, Quaternion, Rect2i, AABB, Plane, Basis, Projection — following the existing pattern (assert `result is <Type>` before `assert_eq`, since a raw dict compares `==` to a struct under GDScript's permissive cross-type comparison). The `node` suite now runs 42/42 coercion tests green.

Clubhouse1661:
Looks good to me. I reviewed the current coercion logic and tests.

The implementation mirrors the existing serialized shapes for the new struct/geometry types, preserves the important permissive cases (`null`, `TYPE_NIL`, already-correct values), and now fails honestly for unsupported mismatches instead of silently no-oping. Leaving `PackedByteArray` out is also the right call until its encoding is designed deliberately.

CodeRabbit’s per-type coverage nit was addressed in the final commit, Copilot had no comments, and CI is green across the full matrix.

Approved from my side.

PR #588, opened 2026-06-21
Harden self-update

ftab:
Reject backslash-containing release asset URLs before authority parsing so spoofed text cannot be interpreted as a trusted GitHub host.

Accept tab/newline-separated sha256 sidecars by normalizing whitespace before reading the digest token, matching common checksum tool output.

Defer the verified install step until after the HTTPRequest checksum callback returns, avoiding update extraction/reload work on the request completion stack.

- `script/ci-check-gdscript`: All GDScript files OK
- `.venv/bin/python -m pytest -v`: 1172 passed, 4 skipped, 2 warnings in 78.21s
- `.venv/bin/ruff check src/ tests/`: All checks passed!
- `.venv/bin/ruff format --check src/ tests/`: 161 files already formatted
- `git diff --check`: passed
- Full Godot-side MCP test_run: 1439 passed, 0 failed, 19 skipped, total 1458
- `script/local-self-update-smoke --force`: completed successfully, exit code 0

The smoke harness did print Godot shutdown warnings about leaked ObjectDB/resources, but the harness verification itself passed and exited 0.

Did I miss anything?

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened trusted URL validation by rejecting malformed download and checksum URLs containing backslashes
  * Improved checksum verification flow by deferring installation until after a successful SHA-256 match
  * Enhanced checksum sidecar parsing by normalizing whitespace for more reliable digest extraction
* **Tests**
  * Added coverage for additional malicious URL patterns and expanded checksum digest parsing scenarios

### DIFF
--- a/plugin/addons/godot_ai/utils/update_manager.gd
+++ b/plugin/addons/godot_ai/utils/update_manager.gd
@@ -314,6 +314,8 @@ static func _is_trusted_download_url(url: String) -> bool:
 	const SCHEME := "https://"
 	if not url.begins_with(SCHEME):
 		return false
+	if url.find("\\") >= 0:
+		return false
 	var rest := url.substr(SCHEME.length())
 	var authority := rest
 	var slash := rest.find("/")
@@ -457,7 +459,7 @@ func _on_checksum_completed(
 
 	print("MCP | self-update checksum verified (sha256 %s)" % actual)
 	install_state_changed.emit({"button_text": "Installing..."})
-	_install_zip()
+	_install_zip.call_deferred()
 
 
 ## Surface an integrity-check failure and drop the staged zip so the bad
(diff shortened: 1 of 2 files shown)


dsarno:
@Coderabbitai review

Clubhouse1661:
Thanks for this. The implementation itself looks good: the backslash rejection is conservative, the checksum parsing change still keeps the 64-char hex validation, and deferring `_install_zip()` after checksum verification makes sense so install/reload work doesn’t run on the HTTPRequest callback stack.

Before merge, could you add the two small regression tests Copilot pointed out?

- `_is_trusted_download_url()` should reject backslash spoof forms like `https://github.com\@evil.com/x.zip` / `https://github.com\evil.com/x.zip`.
- `_parse_sha256_digest()` should accept the new whitespace forms this PR enables, e.g. `<hex>\tgodot-ai-plugin.zip` and `<hex>\ngodot-ai-plugin.zip`.

After those are covered and CI is rerun on the current base, I think this is good to merge.

ftab:
Let me know if that works. Thanks for the review!

Clubhouse1661:
I re-reviewed the current head and this looks ready to merge.

The three security properties hold:

- Backslash rejection happens before URL authority parsing and applies to both the ZIP download URL and the checksum URL.
- Checksum parsing stays strict: whitespace is normalized, but only the first token is used, and it still must be exactly 64 lowercase/uppercase hex chars after normalization.
- Install remains fail-closed when a checksum is present: untrusted checksum host, malformed checksum, hash failure, and mismatch all route through `_fail_verification`, which deletes the staged ZIP and never reaches extraction.

One non-blocking follow-up worth tracking separately: the updater is still “verify if present,” not “verification required.” If a release has no `.sha256` asset, `_verify_then_install()` intentionally installs after host validation only. That preserves compatibility with older releases, so I don’t think it should block this PR, but we should later tighten the release-path allowlist and eventually require checksums once all supported update targets publish sidecars.

Thanks for the focused hardening and the regression tests.

dsarno:
Thank you @ftab and @Clubhouse1661 !

PR #589, opened 2026-06-21
Fix client configure stuck

Clubhouse1661:
## Summary

Fixes a dock failure mode where client status/configuration work could leave the Clients UI stuck in an in-progress state.

This PR hardens the client-probing path in three layers:

- Bounds CLI discovery subprocesses through `McpCliExec.run` instead of raw blocking `OS.execute(..., true)`.
- Lets timed-out client status refreshes keep showing a warning badge without keeping Configure actions disabled forever.
- Avoids noisy Windows `PeekNamedPipe` errors when draining `OS.execute_with_pipe` output.

## Details

`McpCliFinder` now routes `where` / `which` / Unix login-shell lookups through the bounded CLI helper. This prevents a stuck lookup from permanently trapping the refresh worker.

The dock state machine now distinguishes between:

- a worker that is still within its normal running budget, and
- a timed-out/orphanable worker that should no longer block user actions.

So the summary can still show “client probe still running,” while Configure / Configure all are available again.

I also updated `_cli_exec.gd` to drain process pipes with bounded `get_buffer()` chunks instead of `get_as_text()`. On Windows, `get_as_text()` on these pipe handles can emit native `PeekNamedPipe` errors into Godot’s Output panel.

## Tests

- `script/ci-check-gdscript`
- Live `test_run`: 1439 passed, 21 skipped
- Focused suites:
  - `client_refresh_state`
  - `cli_finder`
  - `clients`
  - `dock`
  - `cli_exec`
- `uv run pytest -q tests/unit/test_dock_cli_timeout.py`
- `uv run ruff check tests/unit/test_dock_cli_timeout.py`

### DIFF
--- a/plugin/addons/godot_ai/client_configurator.gd
+++ b/plugin/addons/godot_ai/client_configurator.gd
@@ -509,8 +509,8 @@ static func invalidate_uvx_cli_cache() -> void:
 ## Thread safety: `CliFinder.invalidate()` guards `_cache` / `_searched`
 ## with a mutex so it can race safely against worker threads calling
 ## `find()` from `_run_client_action_worker`. The mutex is held only
-## across the dictionary clear, never across `OS.execute`, so this call
-## can never block the main thread on a subprocess.
+## across the dictionary clear, never across the bounded subprocess lookup,
+## so this call can never block the main thread on a subprocess.
 static func invalidate_cli_cache() -> void:
 	CliFinder.invalidate()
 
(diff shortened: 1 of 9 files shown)


Clubhouse1661:
Follow-up pushed after the 4.3 canary failure.

What happened:
- The first fix (`b4a1bc2`) did resolve the real startup regression on Godot 4.3. The canary moved from timing out with no connected session to loading the plugin and running the suite.
- The remaining failure was only in `cli_exec.test_run_kills_subprocess_when_budget_expires`.

That test asserts the 4.4+ `OS.execute_with_pipe` behavior: spawn, poll, kill on timeout, return `timed_out=true`. On Godot 4.3 we now intentionally use the legacy blocking `OS.execute(..., true)` path because `execute_with_pipe` has known capture/exit-code differences there and was breaking startup discovery.

So I pushed `fc1ad1b` to skip that timeout/kill test on Godot < 4.4, matching the existing 4.3 skips in the same suite.

Issue #597, opened 2026-06-25
Agents feel too comfortable starting the project when there are parse errors


Eleos5:
Including parse errors in project_run tool output would probably fix this

Clubhouse1661:
I’m happy to work on this.

PR #605, opened 2026-06-26
Fix game log run lifecycle

Clubhouse1661:
## Summary

Addresses the stale game-log lifecycle problem from #597. Builds on #603 and #604.

This removes the second false-confidence signal from the parse/load failure path: `logs_read(source="game")` can no longer report stale lines from a prior run while the current run never reached `_mcp_game_helper`.

Game-log identity is now established at play-start, not at `mcp:hello`, so a no-hello run gets its own current run id even when the game fails before the helper can register.

## What changed

- Rotate game-log `run_id` in `begin_game_run()` instead of the `mcp:hello` path.
- Keep historical game-log lines instead of clearing them on run rotation.
- Tag game-log entries with their `run_id`.
- Make `logs_read(source="game")` return current-run lines by default.
- Add `since_run_id` support for retrieving prior-run lines explicitly.
- Add `game_status`, `current_run_id`, and `stale_run_id` to game-log responses.
- Compute `is_running` from `game_status`, not raw editor play-state:
  - `false` for `not_live` / `stopped`
  - `true` for `live` / `launching` / `no_helper`
- Scope `logs_read(source="all")` to current-run game lines so stale game lines do not leak through the combined view.
- Update `docs/TOOLS.md` for the changed game-log contract.

## Why

Before this change, a parse/load failure could leave editor play-state active while the helper never checked in. Because game-log identity was only created at `mcp:hello`, `logs_read(source="game")` could still expose the previous run's lines and report `is_running=true`.

That made a dead current run look alive and reinforced the wrong agent conclusion.

After this change, a no-hello run reports as the current run with empty current-run game logs, `game_status.status="not_live"`, and `is_running=false`.

## Validation

- `test_run suite=editor`: 142 passed, 0 failed, 2 skipped
- Full `test_run`: 1490 passed, 0 failed, 19 skipped
- `git diff --check`: clean aside from existing line-ending warnings

## Summary by CodeRabbit

* **New Features**
  * Game log reads are now run-aware with `since_run_id` for older runs; default results return only the current run.
  * Game log responses now include `run_id`, `current_run_id`, `stale_run_id`, `game_status`, `is_running`, and `dropped_count`.
* **Bug Fixes**
  * New play sessions no longer mix buffered entries across runs; buffered lines retain their original run identifier.
  * Stale hello/ready handling no longer changes the active run identity.
* **Documentation**
  * Updated tool docs to explain run scoping, retention behavior, and the session status fields.
* **Tests**
  * Expanded coverage for run-scoped pagination and readiness/liveness transitions.

### DIFF
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -40,6 +40,25 @@ Godot `_log_error` code/rationale when available, error type, resolved source
 location, and stack/error-tree context corresponding to the Debugger dock's
 Errors tab.
 
+For game logs, `logs_read(source="game")` returns lines from the current game
+run only. Each play-start creates a new `run_id`, even if the game never reaches
+the `_mcp_game_helper` hello beacon; prior run lines stay retained but do not
+appear in the default response. To retrieve a prior run, keep the `run_id` from
+an earlier response and pass `logs_read(source="game", since_run_id="...")`;
+the response includes both `run_id` (the run being read) and `current_run_id`
+(the latest run). `stale_run_id=true` means the requested run is not the current
+one. There is no single `source="game"` call that returns every retained game
+line across all runs; consumers that need history should retain run ids and
+query each run explicitly.
+
+Game and combined log responses also include `game_status` and a liveness-based
+`is_running`. `is_running` is no longer raw editor play-state: it is `false` for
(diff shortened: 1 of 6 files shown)


Clubhouse1661:
@coderabbitai full review

PR #608, opened 2026-06-27
Suggest a free port inline + docs link for foreign port conflicts

### LINKED ISSUE #607: Improve foreign-process port-conflict UX: suggest a free port inline + docs link
## Summary

When the HTTP port (default `8000`) is held by a **non-godot-ai process**, the plugin lands in terminal `INCOMPATIBLE` and the dock shows:

> Port 8000 is occupied by another process; plugin expects godot-ai vX. Stop the other process or change both HTTP and WS ports.

This is informationally adequate but leaves the user with a multi-step chore (open Editor Settings → find `godot_ai/http_port` → change it → also hand-edit each MCP client config). This issue proposes the **cheap, low-risk improvement** first, and documents the **expensive option** to fall back to if this turns out to be a frequent pain point in the wild.

Port `8000` is a very common dev port (Django, `python -m http.server`, many dev servers), so a genuinely foreign occupant is not rare.

## Cheap version (do this first)

1. **Inline copy: name a concrete free port.** Update the `INCOMPATIBLE` (foreign-occupant, no recovery proof) crash-body copy in `mcp_dock.gd` (`_crash_body_for_state`) to surface a suggested free port, e.g.:
   > Port 8000 is in use (8001 is free). Change `godot_ai/http_port` in Editor Settings, then update your client config.

   The suggestion already exists via `ClientConfigurator.suggest_free_port(...)`. On Windows, route the suggestion through the reservation-aware `WindowsPortReservation.suggest_non_excluded_port` so the suggested port doesn't itself fail with WinError 10013. String/seam change only — **zero new risk**.

2. **Add a docs link** next to the message pointing at a short "change the port and reconfigure your clients" section. This is the one piece that genuinely doesn't fit inline (per-client reconfigure steps), and it's the main thing a help link adds over the current message.

3. *(Optional)* **Deep-link button that opens the Editor Settings port field.** Saves the hunt, carries none of the config-rewriting risk.

### Scope notes
- This applies specifically to the `INCOMPATIBLE && !can_recover_incompatible` case (genuinely foreign occupant). When `can_recover_incompatible` is true (a stale/older godot-ai we can prove we own), keep **Restart Server** as the primary action — reclaiming `8000` beats fleeing to a random port.
- The existing `PORT_EXCLUDED` / `FOREIGN_PORT` states already show the `port_picker_panel`; this issue does **not** change those.

## Rationale

- The status quo already tells the user *what* to do; the gap is *effort*, not information. So a bare docs link is only marginally better than today's message — its real value is carrying the client-reconfigure steps.
- The full port-picker (below) is the better end state, but its value is gated behind auto-reconfiguring client configs, which is also ~80% of the work and risk. A picker that moves the port **without** rewriting client configs is a **net negative**: the dock goes green while every MCP client silently points at a dead port.
- So the cheap version captures most of the user-facing value (hands them a free port + the reconfigure procedure) at near-zero risk, and lets us confirm whether this case is actually frequent before investing in the expensive version.

## Expensive version (fallback if cheap isn't enough once in the wild)

Extend the existing `port_picker_panel` to the foreign-occupant case:

- Show the editable port field + "Apply + Reload" when `INCOMPATIBLE && !can_recover_incompatible` **and** only the HTTP port is contested (the WS port `9500` is free). Incompatible godot-ai servers commonly hold *both* HTTP and WS ports — that's why the picker is suppressed for them today (`mcp_dock.gd` ~L914).
- **On Apply, also re-run `ClientConfigurator.configure()` for the already-configured clients** so their URLs follow the new port. This is the critical part — without it we just relocate the dead end from "won't start" to "started but nothing connects."
- Seed the suggestion via the reservation-aware helper; keep it a confirmed click (TOCTOU-racy: another process can grab the suggested port between suggestion and reload — recoverable, the panel just reappears).
- Note `godot_ai/http_port` is stored in **EditorSettings** (per-editor-install, not per-project), so a one-time conflict persists across projects — make the change discoverable and easy to revert.

## Acceptance (cheap version)

- [ ] Foreign-occupant `INCOMPATIBLE` crash body names a concrete free port (reservation-aware on Windows)
- [ ] Docs link surfaced from the dock covering "change the port + reconfigure clients"
- [ ] No behavior change for `PORT_EXCLUDED` / `FOREIGN_PORT` / recoverable-godot-ai (`can_recover_incompatible`) cases

dsarno:
Closes #607.

## Summary

Implements the **cheap version** from #607. When the HTTP port (default `8000`) is held by a genuinely **foreign** (non-godot-ai) process, the dock lands in terminal `INCOMPATIBLE` and tells the user *what* to do but leaves them with the chore — hunt for a free port, then hand-edit each MCP client config. This surfaces a concrete free port inline and adds a docs link carrying the reconfigure steps.

## Changes

- **Inline copy names a concrete free port.** `_crash_body_for_state` in `mcp_dock.gd` now appends a suggestion for the `INCOMPATIBLE && !can_recover_incompatible` case, e.g.:
  > Port 8001 is free — set `godot_ai/http_port` in Editor Settings, then update your client config (How to change the port, below).

  The suggestion comes from the existing `ClientConfigurator.suggest_free_port(...)`, which already routes through `WindowsPortReservation.suggest_non_excluded_port`, so the suggested port won't itself fail with `WinError 10013` on Windows. String/seam change only.

- **Docs link in the crash panel.** A new "How to change the port" button surfaces in the crash panel **only** for the foreign-occupant case, opening `docs/port-conflicts.md` — a short guide covering "change the port + reconfigure your clients" (the per-client reconfigure steps that don't fit inline).

- **New doc:** `docs/port-conflicts.md`.

## Scope / no behavior change

- Applies only to `INCOMPATIBLE && !can_recover_incompatible` (genuinely foreign occupant). When `can_recover_incompatible` is true (a stale/older godot-ai we can prove we own), **Restart Server** stays the primary action and the docs link/free-port hint are suppressed — reclaiming `8000` beats fleeing to a random port.
- No change to `PORT_EXCLUDED` / `FOREIGN_PORT` states or their `port_picker_panel`.

## Acceptance (cheap version)

- [x] Foreign-occupant `INCOMPATIBLE` crash body names a concrete free port (reservation-aware on Windows)
- [x] Docs link surfaced from the dock covering "change the port + reconfigure clients"
- [x] No behavior change for `PORT_EXCLUDED` / `FOREIGN_PORT` / recoverable-godot-ai (`can_recover_incompatible`) cases

## Testing

- `script/ci-check-gdscript` — clean (no parse errors).
- Full Godot handler suite via `script/ci-godot-tests` against a live server + headless editor: **1505/1524 passed, 0 failed** (gap is pre-existing platform skips).
- New `test_dock.gd` coverage: foreign-occupant body names a free port + points at `godot_ai/http_port`; recoverable body keeps Restart Server copy and omits the free-port hint; docs button visibility is gated to the foreign case.

https://claude.ai/code/session_01WwE7zCBJ6qvjk7wyw3X5aU

---
_Generated by [Claude Code](https://claude.ai/code/session_01WwE7zCBJ6qvjk7wyw3X5aU)_

## Summary by CodeRabbit

* **New Features**
  * Added a version-pinned “How to change the port” help button in the dock’s crash panel for non-recoverable incompatible server situations.
  * Enhanced incompatible-server messaging with concrete free-port suggestions for both HTTP and WebSocket settings, including client reconfiguration guidance.

* **Documentation**
  * Added a step-by-step guide for resolving “Port 8000 is occupied” conflicts and updating MCP client URLs.

* **Tests**
  * Expanded dock UI/integration coverage for recoverable vs genuinely foreign incompatible cases, including verification that the reconfiguration docs link is pinned to the installed plugin version.

### DIFF
--- a/docs/port-conflicts.md
+++ b/docs/port-conflicts.md
@@ -19,25 +19,33 @@ second half: changing the port and pointing your MCP clients at the new one.
 > older godot-ai server it *can* reclaim — click that rather than changing the
 > port. This guide is only for the foreign-process case.
 
-## 1. Pick a free port
+## 1. Pick free ports
 
-The crash body already suggests one (e.g. `8001`). On Windows that suggestion
-is checked against the Hyper-V / WSL2 / Docker reservation table, so it won't
-itself fail with `WinError 10013`. You can use the suggested port or choose
-your own free port.
+The plugin uses **two** ports: HTTP (`8000`, the one your MCP clients talk to)
+and WebSocket (`9500`, used internally between the server and the editor). The
+crash body suggests a free value for each (e.g. HTTP `8001`, WS `9501`). On
+Windows those suggestions are checked against the Hyper-V / WSL2 / Docker
+reservation table, so they won't themselves fail with `WinError 10013`. You can
+use the suggested ports or choose your own free ones.
 
-## 2. Change `godot_ai/http_port` in Editor Settings
+If only port `8000` is taken, you technically only need to move the HTTP port —
+but the incompatible-server case that lands you here can hold both, so changing
+both settings is the reliable fix.
+
+## 2. Change `godot_ai/http_port` and `godot_ai/ws_port` in Editor Settings
 
 1. In the Godot editor, open **Editor → Editor Settings**.
-2. Search for `godot_ai/http_port`.
-3. Set it to the free port from step 1 (e.g. `8001`).
+2. Search for `godot_ai/http_port` and set it to the free HTTP port from step 1
+   (e.g. `8001`).
+3. Search for `godot_ai/ws_port` and set it to the free WS port from step 1
+   (e.g. `9501`).
 4. Reload the plugin (toggle it off/on in **Project → Project Settings →
    Plugins**, or restart the editor).
 
-> **Note:** `godot_ai/http_port` is an **Editor Setting**, not a project
-> setting — it is stored per editor install, so the change applies to *every*
-> project you open with this editor. If you only hit the conflict on one
-> machine, remember to revert it later if the foreign process goes away.
+> **Note:** both are **Editor Settings**, not project settings — they are
+> stored per editor install, so the change applies to *every* project you open
+> with this editor. If you only hit the conflict on one machine, remember to
+> revert them later if the foreign process goes away.
 
 ## 3. Reconfigure your MCP clients
 
@@ -65,6 +73,7 @@ and format.
 
 ## Reverting
 
-If the foreign process is gone and you want `8000` back, set
-`godot_ai/http_port` back to `8000` (or clear the override) in Editor Settings,
-reload the plugin, and re-run **Configure all** to point your clients back.
+If the foreign process is gone and you want the defaults back, set
+`godot_ai/http_port` back to `8000` and `godot_ai/ws_port` back to `9500` (or
+clear the overrides) in Editor Settings, reload the plugin, and re-run
+**Configure all** to point your clients back.
--- a/docs/port-conflicts.md
+++ b/docs/port-conflicts.md
@@ -1,0 +1,70 @@
+# Port 8000 is in use by another process
+
+Godot AI's Python server listens on HTTP port `8000` (and WebSocket port
+`9500`). Port `8000` is a popular default for other dev tools — Django,
+`python -m http.server`, and many local servers grab it — so a genuinely
+foreign occupant is not rare.
+
+When a **non-godot-ai** process is already bound to `8000`, the dock can't
+reclaim the port (it has no proof it owns whatever is there), so it stops and
+shows a message like:
+
+> Port 8000 is occupied by an incompatible server. Port 8001 is free — set
+> `godot_ai/http_port` in Editor Settings, then update your client config.
+
+The crash panel names a concrete free port for you. This guide covers the
+second half: changing the port and pointing your MCP clients at the new one.
+
+> If the dock instead offers a **Restart Server** button, the occupant is an
+> older godot-ai server it *can* reclaim — click that rather than changing the
+> port. This guide is only for the foreign-process case.
+
+## 1. Pick a free port
+
+The crash body already suggests one (e.g. `8001`). On Windows that suggestion
+is checked against the Hyper-V / WSL2 / Docker reservation table, so it won't
+itself fail with `WinError 10013`. You can use the suggested port or choose
+your own free port.
+
+## 2. Change `godot_ai/http_port` in Editor Settings
+
+1. In the Godot editor, open **Editor → Editor Settings**.
+2. Search for `godot_ai/http_port`.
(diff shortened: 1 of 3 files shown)


dsarno:
@coderabbitai Good catch on the `blob/main` drift. Fixed in 3c6c98a: the docs button now resolves the URL against the installed plugin version (`blob/v<version>/docs/port-conflicts.md`) via `_port_conflict_docs_url()`. Releases are tagged `v<version>` and the button only exists in builds that ship the doc, so the versioned ref always resolves — and a shipped build never sends users to a tip-of-main guide that has drifted from its UI. Added a test pinning the versioned-URL behavior.

On the earlier "Linked Issues" warning about Windows reservation-awareness: that's already satisfied — `ClientConfigurator.suggest_free_port()` routes through `WindowsPortReservation.suggest_non_excluded_port()` (`client_configurator.gd:141-143`), so the suggested port clears the winnat reservation table.

---

Clubhouse1661:
This looks good as the cheap/manual-path fix from #607, especially now that the docs link is pinned to the installed version.

One coordination note: the current hint/docs are HTTP-only:

> set `godot_ai/http_port`

But the existing incompatible-server message tells users to change both HTTP and WS ports, and the server really does need the pair to stay valid. So I think the manual guidance should name both settings, e.g. `godot_ai/http_port=8123` and `godot_ai/ws_port=9501`, rather than only suggesting a free HTTP port.

dsarno:
> This looks good as the cheap/manual-path fix from #607, especially now that the docs link is pinned to the installed version.
> 
> 
> 
> One coordination note: the current hint/docs are HTTP-only:
> 
> 
> 
> > set `godot_ai/http_port`
> 
> 
> 
> But the existing incompatible-server message tells users to change both HTTP and WS ports, and the server really does need the pair to stay valid. So I think the manual guidance should name both settings, e.g. `godot_ai/http_port=8123` and `godot_ai/ws_port=9501`, rather than only suggesting a free HTTP port.

Good catch fixing....

dsarno:
@Clubhouse1661 Good catch — fixed in c10feed. You're right that this branch isn't purely HTTP-only: it also fires for an incompatible godot-ai server we can't get strong proof on, which commonly holds *both* 8000 and 9500, so suggesting only a free HTTP port would leave the new server unable to bind WS (and it contradicted the embedded "change both HTTP and WS ports" message).

The hint now reads e.g.:
> Ports 8001 (HTTP) and 9501 (WS) are free — set `godot_ai/http_port` and `godot_ai/ws_port` in Editor Settings, then update your client config with the new HTTP port (How to change the port, below).

Both suggestions go through the reservation-aware `suggest_free_port`. Only the HTTP port reaches client configs (WS is server↔plugin), so the wording keeps that distinction, and `docs/port-conflicts.md` + the tests track it. Full suite green locally (1506/1525, 0 failed).

---

PR #613, opened 2026-06-28
Add dock readiness headline and client CTA

Clubhouse1661:
## Summary

Makes the dock headline answer the first user-facing readiness question: whether any AI client is configured to use the connected Godot AI server.

Previously, the top status could show a bare `Connected`, which only meant the editor was connected to the Python server. A first-time user with `0 / 19 configured` could reasonably read that as “setup is done,” even though no AI client could drive Godot yet.

## Changes

- Updates the connected status copy to include client readiness:
  - `Server connected · no AI client configured`
  - `Server connected · 1 AI client configured`
  - `Server connected · N AI clients configured`
- Adds a first-run dock CTA:
  - `Configure an AI client ->`
  - opens the existing Clients tab/window
- Hides the CTA once at least one client is configured.
- Shows the CTA again if the last configured client is removed.
- Keeps disconnected / starting / failure status states unchanged.

## Notes

This is the shallow readiness tier: it reflects configured-client presence from the dock’s existing client status cache. A later server-side follow-up can upgrade the same headline to true live-client readiness using active sessions / recent tool-call data.

A stale or mismatched client still does not count as configured for this server, so the headline can say no client is configured while the drift banner explains the stale config.

## Tests

- `bash script/ci-check-gdscript`
- `test_run suite="dock"` → 63 passed, 0 failed, 3 skipped

## Summary by CodeRabbit

* **New Features**
  * Added a clearer empty-state button in the Clients section when no AI clients are configured.
  * Improved connected status messaging, including “checking” feedback and correct 0/1/many configured-client text.

* **Bug Fixes**
  * Hardened connection status handling to safely cope when connection data isn’t available.
  * Updated the client summary UI so the empty-state button visibility and status label correctly update during configuration changes.

* **Tests**
  * Added deterministic coverage for connected status transitions and empty-state button visibility rules.

### DIFF
--- a/plugin/addons/godot_ai/mcp_dock.gd
+++ b/plugin/addons/godot_ai/mcp_dock.gd
@@ -69,6 +69,7 @@ var _status_icon: ColorRect
 var _status_label: Label
 var _client_grid: VBoxContainer
 var _client_configure_all_btn: Button
+var _client_empty_cta_btn: Button
 var _clients_summary_label: Label
 var _clients_window: Window
 var _dev_mode_toggle: CheckButton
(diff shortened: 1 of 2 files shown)


dsarno:
Smart ux improvement 👍🏼

PR #621, opened 2026-07-01
Error surfacing watermark

Clubhouse1661:
## Summary

- add a central `McpSurfacedErrorTracker` that promotes Debugger Errors-tab rows into a monotonic surfaced-error watermark
- stamp `error_watermark` on plugin responses and inject `new_errors_since_last_call` / `new_errors_hint` on the Python client side
- route `recent_errors` through the same complete source so Debugger-tab-only runtime errors are no longer missed
- keep dispatch watermarks cheap with cached/throttled debugger scans, silent first-observation baselining, and capped promoted Debugger entries
- consolidate Debugger Errors-tree parsing in the tracker and remove the duplicate reader from `EditorHandler`

## Testing

- `python -m compileall -q src tests`
- `ruff check src/ tests/`
- `pytest -q`
  - `1204 passed, 11 skipped, 1 warning`
- Godot `test_run suite=editor`
  - `147 passed, 2 skipped`
- Godot full `test_run`
  - `1537 passed, 19 skipped`
- Live dev-Python smoke on `server_launch_mode=dev_venv`
  - triggered a deferred game runtime error
  - next successful `editor_state` returned `new_errors_since_last_call: 2`
  - `logs_read(source="editor", include_details=true)` showed the Debugger Errors-tab row
  - `logs_read(source="game", include_details=true)` showed the game error row

## Summary by CodeRabbit

* **New Features**
  * Added surfaced error tracking using per-response `error_watermark`, plus `new_errors_since_last_call` and a one-time `new_errors_hint` to guide you to relevant logs.
  * Improved debugger/editor error aggregation by promoting debugger errors into the surfaced error stream and supporting cursor-based pagination.
  * Added run-scoped game warning/error counters used for watermarking.

* **Bug Fixes**
  * Ensured success, handler-error, deferred-timeout, and backpressure responses consistently stamp/propagate watermark metadata.

* **Tests**
  * Expanded unit/integration coverage for watermark stamping, delta/hint behavior, reset/de-duplication, and capped debugger promotion.

### DIFF
--- a/plugin/addons/godot_ai/debugger/mcp_debugger_plugin.gd
+++ b/plugin/addons/godot_ai/debugger/mcp_debugger_plugin.gd
@@ -82,6 +82,7 @@ var _game_run_token := 0
 var _ready_run_token := -1
 var _game_session_id := -1
 var _game_run_active := false
+var _manual_run_armed := false
 var _game_run_started_msec := 0
 var _game_run_started_editor_cursor := 0
 var _game_run_started_debugger_cursor := 0
@@ -110,8 +111,9 @@ func _has_capture(prefix: String) -> bool:
 ## plugin.gd's _enter_tree logs "plugin loaded", and ci-reload-test
 ## asserts "plugin loaded" is the first line after a plugin reload.
 func _setup_session(session_id: int) -> void:
+	_connect_session_stopped(session_id)
 	if EditorInterface.is_playing_scene() and not _game_run_active:
-		_begin_game_run_tracking(_editor_log_cursor(), true, true, false)
+		_begin_game_run_tracking(_editor_log_cursor(), true, true, true, true, true)
 	else:
 		_game_ready = false
 		_ready_run_token = -1
@@ -127,9 +129,12 @@ func _begin_game_run_tracking(
 	helper_expected: bool = true,
 	rotate_game_log: bool = true,
 	sticky_debugger_scan: bool = true,
+	quiet: bool = false,
+	manual_armed: bool = false,
 ) -> void:
 	_game_run_token += 1
 	_game_run_active = true
+	_manual_run_armed = manual_armed
 	_game_ready = false
 	_ready_run_token = -1
 	_game_session_id = -1
@@ -144,7 +149,7 @@ func _begin_game_run_tracking(
 	var run_id := ""
 	if _game_log_buffer and rotate_game_log:
 		run_id = _game_log_buffer.clear_for_new_run()
-	if _log_buffer:
+	if _log_buffer and not quiet:
 		var log_text := "[debug] game capture pending run token %d" % _game_run_token
 		if not run_id.is_empty():
 			log_text += " (run %s)" % run_id
@@ -157,11 +162,29 @@ func _editor_log_cursor() -> int:
 
 func end_game_run() -> void:
 	_game_run_active = false
+	_manual_run_armed = false
 	_game_ready = false
 	_ready_run_token = -1
 	_game_session_id = -1
 	if _surfaced_error_tracker != null:
 		_surfaced_error_tracker.note_game_run_stopped()
+
+
+func _connect_session_stopped(session_id: int) -> void:
+	var session = get_session(session_id)
+	if session == null:
+		return
+	var stopped := Callable(self, "_on_debugger_session_stopped").bind(session_id)
+	if not session.stopped.is_connected(stopped):
+		session.stopped.connect(stopped)
+
+
+func _on_debugger_session_stopped(session_id: int) -> void:
+	if not _manual_run_armed:
+		return
+	if _game_session_id != -1 and session_id != _game_session_id:
+		return
+	end_game_run()
 
 
 func is_game_capture_ready() -> bool:
--- a/plugin/addons/godot_ai/debugger/mcp_debugger_plugin.gd
+++ b/plugin/addons/godot_ai/debugger/mcp_debugger_plugin.gd
@@ -110,12 +110,24 @@ func _has_capture(prefix: String) -> bool:
 ## plugin.gd's _enter_tree logs "plugin loaded", and ci-reload-test
 ## asserts "plugin loaded" is the first line after a plugin reload.
 func _setup_session(session_id: int) -> void:
-	_game_ready = false
-	_ready_run_token = -1
+	if EditorInterface.is_playing_scene() and not _game_run_active:
+		_begin_game_run_tracking(_editor_log_cursor(), true, true, false)
+	else:
+		_game_ready = false
+		_ready_run_token = -1
 	_game_session_id = session_id
 
 
 func begin_game_run(editor_log_cursor: int = 0, helper_expected: bool = true) -> void:
+	_begin_game_run_tracking(editor_log_cursor, helper_expected, true, true)
+
+
+func _begin_game_run_tracking(
+	editor_log_cursor: int = 0,
+	helper_expected: bool = true,
+	rotate_game_log: bool = true,
+	sticky_debugger_scan: bool = true,
+) -> void:
 	_game_run_token += 1
 	_game_run_active = true
 	_game_ready = false
@@ -124,19 +136,23 @@ func begin_game_run(editor_log_cursor: int = 0, helper_expected: bool = true) ->
 	_game_run_started_msec = Time.get_ticks_msec()
 	_game_run_started_editor_cursor = maxi(0, editor_log_cursor)
 	if _surfaced_error_tracker != null:
-		_surfaced_error_tracker.note_game_run_started()
+		_surfaced_error_tracker.note_game_run_started(sticky_debugger_scan)
 		_game_run_started_debugger_cursor = _surfaced_error_tracker.debugger_promoted_total()
 	else:
 		_game_run_started_debugger_cursor = 0
 	_game_helper_expected = helper_expected
 	var run_id := ""
-	if _game_log_buffer:
+	if _game_log_buffer and rotate_game_log:
 		run_id = _game_log_buffer.clear_for_new_run()
 	if _log_buffer:
 		var log_text := "[debug] game capture pending run token %d" % _game_run_token
 		if not run_id.is_empty():
 			log_text += " (run %s)" % run_id
 		_log_buffer.log(log_text)
+
+
+func _editor_log_cursor() -> int:
+	return _editor_log_buffer.appended_total() if _editor_log_buffer != null else 0
 
 
 func end_game_run() -> void:
@@ -235,7 +251,11 @@ func _recent_editor_errors_since(cursor: int) -> Dictionary:
 	var out: Array[Dictionary] = []
 	var truncated := false
 	if _surfaced_error_tracker != null:
-		var captured_by_tracker: Dictionary = _surfaced_error_tracker.editor_entries_since(maxi(0, cursor), _game_run_started_debugger_cursor)
+		var captured_by_tracker: Dictionary = _surfaced_error_tracker.editor_entries_since(
+			maxi(0, cursor),
+			_game_run_started_debugger_cursor,
+			false,
+		)
 		truncated = bool(captured_by_tracker.get("truncated", false))
 		for raw_entry in captured_by_tracker.get("entries", []):
 			var compact := _compact_editor_error(raw_entry)
--- a/plugin/addons/godot_ai/debugger/mcp_debugger_plugin.gd
+++ b/plugin/addons/godot_ai/debugger/mcp_debugger_plugin.gd
@@ -123,7 +123,11 @@ func begin_game_run(editor_log_cursor: int = 0, helper_expected: bool = true) ->
 	_game_session_id = -1
 	_game_run_started_msec = Time.get_ticks_msec()
 	_game_run_started_editor_cursor = maxi(0, editor_log_cursor)
-	_game_run_started_debugger_cursor = _surfaced_error_tracker.debugger_promoted_total() if _surfaced_error_tracker != null else 0
+	if _surfaced_error_tracker != null:
+		_surfaced_error_tracker.note_game_run_started()
+		_game_run_started_debugger_cursor = _surfaced_error_tracker.debugger_promoted_total()
+	else:
+		_game_run_started_debugger_cursor = 0
 	_game_helper_expected = helper_expected
 	var run_id := ""
 	if _game_log_buffer:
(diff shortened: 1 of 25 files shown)


plugin/addons/godot_ai/utils/surfaced_error_tracker.gd
```diff
@@ -0,0 +1,337 @@
...
+
+
+func refresh_debugger_errors(force: bool = true) -> void:
+	var now := Time.get_ticks_msec()
+	if not force and not _should_scan_debugger_for_cached_watermark(now):
+		return
+	_last_debugger_refresh_msec = now
+	for entry in read_debugger_error_entries():
+		var key := _log_entry_key(entry)
+		if _promoted_debugger_keys.has(key) or _suppressed_debugger_keys.has(key):
```
dsarno: **A recurring identical error is permanently invisible to the watermark.**

The promotion dedup key (`_log_entry_key`: `level|text|path|line`, no timestamp) lives in `_promoted_debugger_keys`/`_suppressed_debugger_keys`, and nothing ever clears those sets — not `clear_logs` (`_clear_debugger_error_trees` only clears the UI Tree), not `note_game_run_started` (only toggles scan flags).

Concrete repro of the feature's core loop:

1. Run 1 crashes with `Assertion failed` at `res://foo.gd:10` → promoted with sequence N, hint fires. ✅
2. `begin_game_run` baselines `_game_run_started_debugger_cursor = N`.
3. Agent attempts a fix, reruns; the same crash appears as a *new* Errors-tab row (same key, different timestamp) → skipped here as already-seen.
4. `debugger_promoted` never advances → no `new_errors_since_last_call` hint; `editor_entries_since(cursor, N)` returns nothing run-scoped, so `_explain_not_live` reports the error as "may predate this run".

The agent concludes the fix worked. Since the tracker's own header says these errors can land *solely* in the Debugger tab, no other channel restores attribution.

A per-run component in the key (e.g. include a run counter bumped in `note_game_run_started`), or clearing the key sets at run start, would fix this — trading off re-counting rows that persist in the tab across runs, which the run-start re-baseline in `begin_game_run` already absorbs.

---

src/godot_ai/transport/websocket.py
```diff
@@ -320,3 +324,32 @@ async def send_command(
...
+    for key, raw_current in value.items():
+        try:
+            current = max(0, int(raw_current))
+        except (TypeError, ValueError):
+            continue
+
+        previous = session.error_watermark.get(key)
+        if previous is not None:
+            previous_int = max(0, int(previous))
+            if current >= previous_int:
```
dsarno: **Reset detection misses reset-then-reaccumulate, under-reporting a rerun's errors.**

A reset is only detected on a strict decrease, but the server never observes the zero: `run_project` → `begin_game_run` → `clear_for_new_run()` resets `_error_warn_total` synchronously, then the handler returns `DEFERRED_RESPONSE` — the dispatcher's deferred branch returns before `_stamp_error_watermark`, so no envelope carrying `game_error_warn=0` is sent. The deferred reply is stamped only at send time, after `_finish_run_project_deferred` waits up to `RUN_READY_WAIT_SEC` for liveness — ample time for the new run to re-emit its startup errors.

So: run 1 ends with `game_error_warn=3` (baseline 3). Run 2 reproduces the same 3 startup errors before the deferred reply is stamped → `current=3 >= previous=3` → delta 0. Three real new errors, no hint — again the fix-and-rerun loop this feature targets. (5 errors would report only 2.)

One fix: have the plugin include a monotonic run counter (or run_id) in the watermark so the server can detect the run boundary explicitly instead of inferring it from a decrease.

---

src/godot_ai/godot_client/client.py
```diff
@@ -189,4 +189,16 @@ async def send(
                 data=error.data if error else {},
             )
 
+        if response.new_errors_since_last_call > 0:
```
dsarno: **Error responses and internal probes consume the delta without surfacing it.**

`_sync_error_watermark_for_session` advances the session baseline for *every* response in `websocket.py`, but the hint is only injected here on the success path:

- `status == "error"` raises `GodotCommandError` above, before this block — the delta carried on that envelope is consumed and never shown (and the dispatcher stamps error responses too, per the new tests).
- Internal probes whose data dict is discarded (e.g. the `require_writable_async` readiness probe) consume the delta the same way.
- A late response arriving after `send_command` timed out (future already popped) advances the baseline with nothing delivered.

In all three cases the next successful call diffs against the already-advanced baseline and reports 0 — e.g. game logs 2 new errors, agent's next call fails with `NODE_NOT_FOUND`, and the errors are never hinted.

Consider making the delta consume-on-surface rather than consume-on-observe: accumulate undelivered new-error counts on the `Session` and only reset when a hint actually reaches a tool result.

---

plugin/addons/godot_ai/utils/surfaced_error_tracker.gd
```diff
@@ -0,0 +1,337 @@
...
+		_promoted_debugger_entries.append(promoted)
+	_trim_promoted_debugger_entries()
+
+
+func watermark(force_debugger_scan: bool = false) -> Dictionary:
+	refresh_debugger_errors(force_debugger_scan)
+	return {
+		"editor_ring": _editor_log_buffer.appended_total() if _editor_log_buffer != null else 0,
+		"debugger_promoted": _debugger_promoted_total,
+		"game_error_warn": _game_log_buffer.error_warn_total() if _game_log_buffer != null else 0,
```
dsarno: **Violates AGENTS.md's self-update release-shape rule — new file fails to compile against a stale `game_log_buffer.gd`.**

AGENTS.md (Self-update section): *"avoid adding new files that reference constants, methods, or static/non-static shape changes added to existing load-surface scripts in the same release. This applies to both `class_name` scripts and preload-only scripts because the failure mode is stale Script-object content, not just class registry skew."*

This new file calls `_game_log_buffer.error_warn_total()` through a statically typed `McpGameLogBuffer` reference, and `error_warn_total()` is added to that pre-existing `class_name` script in this same PR (absent at base). During the two-phase-runner upgrade window the rule describes, a stale `McpGameLogBuffer` Script object makes this entire file fail to compile at load — the exact failure mode the rule names. (All other cross-script references here — `appended_total()`, `get_since()`, etc. — exist at base; this is the only violation.)

Options: type the field loosely and guard with `has_method("error_warn_total")`, or land `error_warn_total()` on `McpGameLogBuffer` one release ahead of this file.

---

plugin/addons/godot_ai/utils/surfaced_error_tracker.gd
```diff
@@ -0,0 +1,337 @@
...
+	return "%s|%s|%s|%s" % [
+		str(entry.get("level", "")),
+		str(entry.get("text", "")),
+		str(entry.get("path", "")),
+		str(entry.get("line", 0)),
+	]
+
+
+func _should_scan_debugger_for_cached_watermark(now_msec: int) -> bool:
+	if not _debugger_scan_active and now_msec > _debugger_scan_until_msec:
```
dsarno, line 411: **Manually launched runs (F5) never advance the debugger watermark.**

`begin_game_run` has exactly one caller — the MCP `run_project` handler — and `_setup_session`, the only hook that fires for manual debug sessions, never calls `note_game_run_started`. So for an F5 run, `_debugger_scan_active` stays false and `_debugger_scan_until_msec` stays 0: every dispatch-path `watermark()` stamp uses `force=false` and is blocked here.

The catch-22: forced scans only happen from `logs_read(source=editor)` / `_explain_not_live`, and the `new_errors_hint` is the mechanism meant to prompt those calls — so the gap never self-heals. `game_error_warn` partially covers manual runs (the `mcp:log_batch` path isn't gated on `_game_run_active`), but only on Godot 4.5+ with the helper registered — not the "errors that land solely in the Debugger Errors tab" case this tracker exists for, and the PR summary claims those are "no longer missed."

Arming the scan window from `_setup_session` (which already fires per debug session, manual or MCP-launched) would close this.

---

plugin/addons/godot_ai/debugger/mcp_debugger_plugin.gd
```diff
@@ -224,6 +234,28 @@ func recent_editor_errors_since(cursor: int) -> Dictionary:
 func _recent_editor_errors_since(cursor: int) -> Dictionary:
 	var out: Array[Dictionary] = []
 	var truncated := false
+	if _surfaced_error_tracker != null:
+		var captured_by_tracker: Dictionary = _surfaced_error_tracker.editor_entries_since(maxi(0, cursor), _game_run_started_debugger_cursor)
```
dsarno: **Per-frame forced full-tree scans during game launch.**

`project_handler._finish_run_project_deferred` polls `recent_editor_errors_since` in an `await process_frame` loop while the game boots, and this tracker path calls `editor_entries_since` → `refresh_debugger_errors(true)` — `force=true` bypasses the 250 ms throttle, so every frame pays a recursive walk of the debugger subtree plus full entry construction (children walk, frames parse, nested dict allocation) for *every* Errors-tab row, ~60×/s on the editor main thread. Pre-PR, each iteration was a cheap ring-buffer `get_since`.

With a large Errors tab this stalls the editor for the whole ready-wait window. Passing `force=false` here (letting the throttle apply — the run-active window already keeps scans fresh at 250 ms), or computing the key from `get_text(1)`/`get_metadata(0)` first and only building full entries for unseen keys, would keep the launch loop cheap.

---

src/godot_ai/godot_client/client.py
```diff
@@ -189,4 +189,16 @@ async def send(
...
+            data = dict(response.data)
+            count = response.new_errors_since_last_call
+            data["new_errors_since_last_call"] = count
+            plural = "s" if count != 1 else ""
+            data["new_errors_hint"] = (
```
dsarno, line 202: **Warnings are reported as "GDScript errors".**

Both counters feeding the watermark include warnings: `game_error_warn` explicitly counts `warn`-level entries, and `editor_ring` is `McpEditorLogBuffer.appended_total()`, whose feeder (`editor_logger._log_error`) captures `push_warning` and deprecation warnings too. A single benign `push_warning` during a save produces `new_errors_hint: "1 new GDScript error since your last call"` and sends the agent hunting for an error that doesn't exist — repeatedly, on warning-heavy projects.

Either count only `error`-level entries into the watermark, or word the hint as "errors/warnings".

---

plugin/addons/godot_ai/utils/surfaced_error_tracker.gd
```diff
@@ -0,0 +1,337 @@
...
+	var entries: Array[Dictionary] = []
+	var truncated := false
+	if _editor_log_buffer != null:
+		var captured: Dictionary = _editor_log_buffer.get_since(maxi(0, editor_cursor), -1)
+		truncated = bool(captured.get("truncated", false))
+		for entry in captured.get("entries", []):
+			entries.append(entry)
+	if debugger_cursor < _oldest_retained_debugger_sequence - 1:
+		truncated = true
+	for entry in _promoted_debugger_entries:
```
dsarno, line 145: **Two `recent_errors` regressions vs the pre-PR reader.**

1. **Missing dedup**: `editor_entries_since` appends editor-buffer entries and promoted debugger entries with no `_log_entry_key` check — unlike its sibling `collect_editor_log_entries` (and the old `_has_equivalent_log_entry` merge in `editor_handler.gd`) which dedup exactly these two sources. An error landing in both (realistic for prefix-less titles like `Parse Error: Expected statement`) appears twice and counts twice against the 5-entry cap in `_recent_editor_errors_since`, crowding out distinct errors.

2. **Fallback ordering**: `retained_recent_editor_entries` reverses `collect_editor_log_entries()`, which appends debugger rows *after* all buffer entries — so reversal puts debugger rows first. Since the fallback only runs when the run-scoped path found nothing, any debugger rows reaching it necessarily predate the run cursor; a stale row now outranks newer editor-buffer errors, and `_explain_not_live`'s "A recent editor error may be related: …" can name the wrong error. The old fallback was newest-first buffer entries only.

---

dsarno (commented):
Reviewed at head 44da77c. The feature is well-motivated and the layering is clean — optional `error_watermark` field keeps old plugins compatible, stamping mirrors the existing `readiness` pattern on all response paths, and test coverage of the stamping/delta/baseline semantics is genuinely good. The earlier CodeRabbit findings (suppressed-key cap, per-key baselining, dedup scans, duplicated stamp helper) are all addressed at this head; the findings below are distinct and were each verified against the code.

The three most important ones (permanent dedup key, reset-then-reaccumulate, delta consumed on error responses) share a theme: they make the feature silently report 0 new errors in its primary scenario — an agent rerunning the game after attempting a fix for a reproducible error. Inline comments have concrete repros and suggested directions. The AGENTS.md release-shape issue is the fourth blocker-grade item, since it breaks the documented self-update window.

Two lower-stakes notes not worth inline threads:

- The old `_locate_debugger_error_trees` guard (`if _debugger_plugin == null and _debugger_errors_root == null: return`) was dropped in the move to the tracker, and `EditorHandler._init` now fabricates a tracker when none is injected — so no-plugin/no-root handlers (several in-editor tests, e.g. `test_get_logs_source_editor_empty_when_no_buffer`) now scan the real editor UI from `get_base_control()`. Production is unaffected, but those tests can flake against real Errors-tab contents, and `clear_logs` on such a handler could press the real Clear button.
- Cleanup opportunities in the tracker: it hand-rolls a capped buffer with `pop_front()` eviction that `McpStructuredLogRing` already provides (including `appended_total()`/`get_since()`/truncation), and `dispatcher.set_surfaced_error_tracker()` has no callers.

---

Clubhouse1661:
Thanks for the review @dsarno!

Updated in `301a03c` to address the round-two review feedback.

Changes in this commit:
- stopped clearing the user-visible Debugger Errors history on run start; run start now resets only the tracker’s internal scan/re-arm state
- changed editor-ring fallback behavior so missing `error_appended_total()` does not inflate surfaced errors from non-error log lines
- documented the `run_seq` boundary behavior in `_sync_error_watermark_for_session`
- simplified `send_command()` now that test doubles accept `surface_error_hints`
- kept promoted Debugger entries bounded while retaining longer-lived per-key occurrence counts so repeated identical rows remain monotonic without re-promotion churn
- added/updated tests for manual run re-arming, warning-only behavior, tracker retention, and the session watermark edge cases

Validation after the commit:
- `python -m compileall -q src tests`
- `ruff check src/ tests/`
- full Python `pytest`: `1208 passed, 11 skipped, 1 warning`
- full Godot `test_run`: `1541 passed, 19 skipped`
- live dev-Python smoke on `server_launch_mode=dev_venv`
- true UI/manual F5 smoke against the Godot editor:
  - manual F5 runtime-error scene surfaced `new_errors_since_last_call: 1`
  - stop + second manual F5 surfaced again, verifying re-arm behavior
  - `logs_read(source="editor", include_details=true)` showed the Debugger Errors-tab-only row with `appended_total: 0`
  - warning-only scene produced warning logs but no surfaced-error hint

CodeRabbit’s latest review on `301a03c` reported no actionable comments.

PR #622, opened 2026-07-03
Add TileMap/TileSet MCP tools and 2D viewport screenshot support

saschiwy:
## PR Description
This PR expands 2D authoring support by adding new TileMap and TileSet MCP operations and improving editor screenshots for 2D scenes.

### Includes
- New tilemap_manage operations for set cell, fill rect, clear layer, and read used cells
- New tileset_manage operations for specialized tileset generation, atlas tile listing, and atlas image export (Base64 PNG)
- New viewport_2d screenshot source with improved guidance for 2D scene capture
- Server/domain/tool-catalog wiring for tilemap and tileset
- Added unit and Godot-side tests for behavior, errors, and invariants

## Summary by CodeRabbit

* **New Features**
  * Added tilemap management tools to set cells, fill rectangles, clear layers, and query used cells with undo/redo support.
  * Added tileset atlas tools to list atlas tiles and export atlas images as Base64 PNGs (with optional max size).
  * Extended editor screenshot capture with `source="viewport_2d"` for 2D scenes.
* **Bug Fixes**
  * Improved 2D scene detection and updated validation messages, error hints, and suggestions for `viewport_2d`.
* **Tests**
  * Expanded unit, in-editor, and integration coverage for tilemap/tileset tools, screenshot behavior, and undo/redo.


saschiwy:
BTW. Thank you for your great project!

plugin/addons/godot_ai/handlers/tilemap_handler.gd
```diff
@@ -0,0 +1,159 @@
...
+	var snapshot: Array[Dictionary] = []
+	for x in range(rx, rx + rw):
+		for y in range(ry, ry + rh):
+			var pos := Vector2i(x, y)
+			cells.append(pos)
+			snapshot.append({"pos": pos, "state": _capture_cell_state(node, pos)})
+	_undo_redo.create_action("MCP: TileMap set_cells_rect %dx%d" % [rw, rh])
+	for pos in cells:
+		_undo_redo.add_do_method(node, "set_cell", pos, src, atlas)
+	_undo_redo.add_undo_method(self, "_restore_cells_snapshot", node, snapshot)
```
dsarno: **Blocker — undoing a rect fill destroys every tile outside the rect.**

The snapshot built above only covers cells *inside* the rect, but the undo callback `_restore_cells_snapshot` starts with `node.clear()`:

```gdscript
func _restore_cells_snapshot(node: TileMapLayer, snapshot: Array[Dictionary]) -> void:
	node.clear()
	for entry in snapshot:
		_restore_cell_state(node, entry.pos, entry.state)
```

Repro: a layer has 500 hand-placed tiles; call `tilemap_set_cells_rect` for a 3×3 patch; press Ctrl+Z. The layer is cleared and only the 9 snapshotted cells come back — the other ~491 tiles are permanently lost (redo can't restore them either, since the do-methods only set the rect cells).

`clear_layer` is fine (its snapshot covers all used cells), so the simplest fix is a rect-only restore for this op:

```gdscript
func _restore_rect_snapshot(node: TileMapLayer, snapshot: Array[Dictionary]) -> void:
	for entry in snapshot:
		_restore_cell_state(node, entry.pos, entry.state)
```

(`_restore_cell_state` already erases cells whose prior state was `has_tile: false`, so no `clear()` is needed.) Alternatively, capture/restore `node.tile_map_data` wholesale via `add_undo_property` — that's O(1) undo registrations and would also cover `clear_layer` more cheaply.

Worth adding a regression test that undoes a rect fill on a layer with pre-existing tiles outside the rect and asserts they survive — the current undo tests only cover `clear` on cells the test itself placed.

---

plugin/addons/godot_ai/handlers/tileset_handler.gd
```diff
@@ -0,0 +1,357 @@
...
+		)
+
+	var source_index: int = params.get("source_id", -999)
+	if source_index < 0 or source_index >= ts.get_source_count():
+		return ErrorCodes.make(
+			ErrorCodes.VALUE_OUT_OF_RANGE,
+			"source_id %d is out of range (TileSet has %d sources)" % [source_index, ts.get_source_count()]
+		)
+
+	var source_id: int = ts.get_source_id(source_index)
```
dsarno: **Blocker — `source_id` means two different things within the same tool.**

Here (and in `get_atlas_image`), the `source_id` param is treated as a *positional index* into the TileSet's source list (`ts.get_source_id(source_index)`), while `tileset_generate_specialized`'s `layer_sources` and every `tilemap_*` op use *raw* TileSet source IDs (`main_ts.get_source(int(src_id))`, `set_cell(pos, src, ...)`).

On a TileSet with sparse IDs (say sources 5 and 42 — exactly the case in `test_source_index_uses_tileset_order_not_raw_source_id`), an agent that just painted with source ID 42 and then calls `tileset_get_atlas_tiles(source_id=42)` gets `VALUE_OUT_OF_RANGE`; calling `source_id=1` silently returns source 42's tiles. The palette an agent reads must be addressable the same way it paints, otherwise it draws the wrong tiles with no error signal.

Suggest making the atlas ops take raw source IDs (validate via `ts.has_source(source_id)`), which matches both the rest of this PR and Godot's own API vocabulary. If index-based lookup is genuinely wanted, name the param `source_index` — but don't overload `source_id`.

---

plugin/addons/godot_ai/handlers/tileset_handler.gd
```diff
@@ -0,0 +1,357 @@
...
+	if max_size > 0:
+		var longest_edge: int = max(original_width, original_height)
+		if longest_edge > max_size:
+			var scale: float = float(max_size) / float(longest_edge)
+			var new_w: int = max(1, int(original_width * scale))
+			var new_h: int = max(1, int(original_height * scale))
+			img = img.duplicate()
+			img.resize(new_w, new_h, Image.INTERPOLATE_BILINEAR)
+
+	var png_bytes: PackedByteArray = img.save_png_to_buffer()
```
dsarno: **VRAM-compressed atlas textures return a "success" response with an empty image.**

If the atlas texture was imported with VRAM compression (very common for game textures), `tex.get_image()` returns an `Image` in a compressed format (DXT/ETC/BPTC). `Image.resize()` and `save_png_to_buffer()` both fail on compressed formats — `png_bytes` ends up empty and the handler returns `{"image_base64": "", ...}` as success, so the agent receives a zero-byte image with no error signal.

Fix: decompress before use, and treat an empty PNG buffer as an error:

```gdscript
if img.is_compressed():
	var derr := img.decompress()
	if derr != OK:
		return ErrorCodes.make(ErrorCodes.INTERNAL_ERROR,
			"Could not decompress texture of source %d: %s" % [source_id, error_string(derr)])
```

Two smaller things in the same block: the `img.duplicate()` before `resize()` is a redundant full-image copy (`tex.get_image()` already returns a fresh copy — resize in place); and `editor_handler.gd`'s `_finalize_image` already implements this exact max-size + PNG + Base64 pipeline (with LANCZOS rather than BILINEAR) — reusing or extracting it would keep the two paths from drifting and give consistent scaling quality. Note `get_atlas_image` currently has no test coverage on either the GDScript or Python side, so none of this is caught by CI.

---

plugin/addons/godot_ai/handlers/editor_handler.gd
```diff
@@ -631,8 +631,24 @@ func take_screenshot(params: Dictionary) -> Dictionary:
...
+				return ErrorCodes.make(ErrorCodes.EDITOR_NOT_READY,
+					"No scene open — open a scene first")
+			## Capture the 2D editor viewport directly; no view_target/coverage for 2D.
+			var image_2d: Image = viewport.get_texture().get_image()
+			if image_2d == null or image_2d.is_empty():
+				return _empty_image_error(
+					"viewport_2d",
+					"Captured an empty image from the 2D viewport. The 2D viewport produced no output — typically headless mode or the 2D viewport has not drawn a frame yet."
+				)
+			return _finalize_image(image_2d, "viewport_2d", max_resolution)
```
dsarno, line 440: **`viewport_2d` is undiscoverable — and contradicted — at the client-facing contract.**

The MCP tool docstring in `src/godot_ai/tools/editor.py` (untouched by this PR) still documents `source` as `"viewport" | "cinematic" | "game"`, while this branch accepts `viewport_2d` and the error hints below actively tell agents to send it (`"use source='viewport_2d' for 2D scenes"`). An agent composing calls from the tool schema either never discovers the new 2D capture, or sees the server's own suggestion conflict with the documented enum. Please add `viewport_2d` to the `source` docs and the "Picking a source" guidance in `editor.py` (2D scene → `viewport_2d`).

Also: this branch returns before the `view_target`/`coverage`/`elevation`/`azimuth`/`fov` handling, so those params are silently ignored for 2D captures while the Python side still echoes `view_target` in the response metadata as if it was honored. Ignoring them may be fine for a first cut, but the contract should say so — and ideally the response should carry a `view_target_ignored` note rather than implying the framing happened.

One more capture-freshness concern: unlike the 3D `view_target` path, there's no `RenderingServer.force_draw()` before `viewport.get_texture().get_image()`. If the 2D main screen isn't currently visible (agent is on the 3D/Script tab while editing tiles via `tilemap_manage`), this can return a stale frame that predates the edits — worth a force-draw or at least a docs caveat, since "screenshot to verify my tile edits" is the primary use case for this source.

---

plugin/addons/godot_ai/handlers/tileset_handler.gd
```diff
@@ -0,0 +1,357 @@
...
+		for src_id in source_ids:
+			var src := main_ts.get_source(int(src_id))
+			new_ts.add_source(src.duplicate(true), new_id)
+			new_id += 1
+
+		var err := ResourceSaver.save(new_ts, output_path)
+		if err != OK:
+			return ErrorCodes.make(
+				ErrorCodes.INTERNAL_ERROR,
+				"Failed to save %s (Godot error %d)" % [output_path, err]
```
dsarno: Three conventions from AGENTS.md this write path still misses (the follow-up commit's path validation is a good step, thanks):

1. **`error_string(err)`** — AGENTS.md: "Use Godot's built-in `error_string(err)` to translate numeric error codes in error messages." `"Godot error %d"` gives the agent an opaque number; `error_string(err)` gives it something actionable.
2. **`"undoable": false` + reason** — AGENTS.md: "If an operation genuinely can't be undone (file writes...), include `\"undoable\": false` with a reason" (cf. `filesystem_write_text`). This op writes `.tres` files to disk with no undo, but the response doesn't disclose that, so an agent may assume Ctrl+Z cleanup applies.
3. **`McpResourceIO.save_to_disk`** (`utils/resource_io.gd`) — the shared save helper the texture/curve/environment/resource handlers use for exactly this operation. Calling it instead of raw `ResourceSaver.save` gets you the write-path validation, the `EditorFileSystem.update_file()` rescan (without it, freshly created `.tres` files don't appear in the FileSystem dock), and the save re-entrancy guard — plus items 1 and 2's formatting for free.

Same reuse note applies to `_resolve_layer` in `tilemap_handler.gd`: `McpNodeValidator.resolve_or_error` is the shared node-resolution helper used by the other 16 handlers, and hand-rolling it here skips the edited-scene pinning (`EDITED_SCENE_MISMATCH`) protection and the richer NODE_NOT_FOUND recovery hints.

---

src/godot_ai/tools/tileset.py
```diff
@@ -0,0 +1,106 @@
...
+          "created": ["res://.../volcano_floor.tres", ...],
+          "skipped": ["res://.../volcano_walls.tres", ...]
+        }
+
+        Source-ID remapping: Source-IDs in specialized files start at 0.
+        Example: volcano lava (Source 8 in main .tres) → Source 0 in
+        volcano_animated.tres. Use the remapped ID in tilemap_manage calls.
+
+        Hard stop: if the main {biom}.tres does not exist, the command
+        returns RESOURCE_NOT_FOUND with a hint to run setup_{biom}.gd first.
```
dsarno: This description promises "a hint to run setup_{biom}.gd first", but the handler's error is just `"Main TileSet not found: <path>"` — and no `setup_{biom}.gd` mechanism exists in this repo. An agent hitting the missing-tileset case will go looking for a script/op that doesn't exist and stall on a documented recovery path that was never implemented. Please drop the hint (or implement it, but see the design comment in the review body — the `setup_{biom}.gd` reference looks like it leaked in from a specific game project).

---

dsarno (changes requested):
Thanks for this contribution — and for the quick follow-up commit addressing the earlier bot feedback (the `undoable` flags, rect-size guard, path validation, `duplicate(true)` before `add_source`, layer-definition copying, typed-load fix, `RESOURCE_NOT_FOUND` alignment, and the `hypothesis` dev dependency all check out; and placing `undoable` inside `data` matches the existing handlers' shape).

The capability itself is very welcome: tile-level editing, atlas inspection, and 2D viewport capture close a real gap — today an agent connected to this server can't place a tile or see a 2D scene at all, and these three pieces together form the full loop (inspect palette → paint → verify visually). We'd like to land this.

**Requesting changes on the following (details inline):**

1. **Undo of `tilemap_set_cells_rect` destroys all tiles outside the rect** — `_restore_cells_snapshot` starts with `node.clear()` but the snapshot only covers rect cells. This is user data loss and the main blocker.
2. **`source_id` is a positional index in the atlas ops but a raw TileSet source ID everywhere else** in this PR — agents will read one palette and paint from another on sparse-ID tilesets.
3. **VRAM-compressed atlas textures make `tileset_get_atlas_image` return success with an empty image** — needs `img.decompress()` and an empty-buffer error.
4. **`viewport_2d` is missing from the `editor_screenshot` docstring/contract** in `src/godot_ai/tools/editor.py`, and silently ignores `view_target`/`coverage` while metadata implies otherwise.
5. Smaller convention items on the tileset save path (`error_string(err)`, `"undoable": false` for file writes, reusing `McpResourceIO.save_to_disk` / `McpNodeValidator.resolve_or_error`).

**One design ask on `tileset_generate_specialized`:** as written it encodes a specific project's conventions into the shared server — the `{root_dir}/{biom}/{biom}.tres` folder layout, `{biom}_{layer}.tres` output naming, lava/volcano vocabulary in the tool descriptions, and a reference to a `setup_{biom}.gd` script that doesn't exist in this repo. Any project not organized that way can't reach its files through this parameter shape, and the (misspelled) `biom` param would become permanent public API. Could this op either be split out of this PR, or generalized to take an explicit source `tileset_path` plus an output mapping (e.g. `{output_path: [source_ids]}`)? The underlying "build a TileSet from a subset of another TileSet's sources" operation is useful — it's the baked-in naming convention we'd like to avoid. The tilemap/atlas/`viewport_2d` parts could land quickly once items 1–4 are fixed, so splitting may be the fastest path.

**Test coverage:** `tileset_get_atlas_image` currently has no tests on either side, and there are no `tests/integration/` cases for the new rollup tools (AGENTS.md asks for handler unit test + Python integration test + GDScript test per tool). A GDScript regression test for the rect-fill undo case (pre-existing tiles outside the rect survive Ctrl+Z) would have caught item 1.

---

dsarno:
CI note: the two failing checks (`Python 3.11 / ubuntu-latest` and `Python 3.13 / ubuntu-latest`) are a base-drift issue, not a bug in this PR's logic.

`main` recently merged #621 ("Error surfacing watermark"), which added a `surface_error_hints` keyword to `DirectRuntime.send_command` → `client.send()`. CI tests the PR merged with current `main`, so the stub clients in the new test files fail with:

```
TypeError: StubClient.send() got an unexpected keyword argument 'surface_error_hints'
```

11 failures, all in `tests/unit/test_tilemap_tileset_handlers.py` and `tests/unit/test_tileset_handler_properties.py` (the branch passes on its own merge-base — this only shows up against latest `main`).

Fix: rebase onto `main` and update the stub/fixture `send()` signatures to accept the new kwarg, e.g.

```python
async def send(self, command, params=None, session_id=None, timeout=5.0, surface_error_hints=True):
```

(or add `**kwargs` to stay resilient to future signature additions). All other jobs — including the Godot-side test suites on Linux/macOS/Windows and Godot 4.3 — are green on the current head.

---

dsarno:
Great turnaround on the review — the blocker fixes, the biome-op split, and the new coverage all look right, and the full Python suite passes locally on this head (1235 passed).

CI is nearly green now: the two failing checks are just the **Lint** step (`ruff check src/ tests/`), with 4 × `E501` line-too-long (>100) in `tests/unit/test_tilemap_tileset_handlers.py` at lines 14, 214, 271, and 281:

- 214/281 are the stub `send(...)` signatures — wrap the parameter list across lines.
- 271 is the base64 literal — split the string or add `# noqa: E501`.
- 14 just needs a wrap.

One last small thing: this repo commits the Godot-generated `.uid` sibling for every `.gd` file, and the four new scripts are missing theirs (`handlers/tilemap_handler.gd.uid`, `handlers/tileset_handler.gd.uid`, `tests/test_tilemap.gd.uid`, `tests/test_tileset_atlas.gd.uid`). Opening the project in the editor generates them — please commit the ones from your machine so checkouts don't churn fresh UIDs.

With those two, this should be green and ready to go.

---

dsarno:
Two more non-blocking nits to fold in while you're touching things up — neither affects the approval:

1. **`src/godot_ai/tools/editor.py` (~line 180):** the updated `Sources:` block landed at 16-space indent while the rest of the docstring uses 8, so the `"viewport"`/`"viewport_2d"` bullets are misaligned relative to `"cinematic"`/`"game"` below them. This docstring is the tool description MCP clients render, so worth re-indenting to match.

2. **`plugin/addons/godot_ai/handlers/tilemap_handler.gd` (`_resolve_layer`):** the empty-path check is a good addition, but the handler still hand-rolls node resolution. The shared `McpNodeValidator.resolve_or_error` (used by the other handlers) would also give the tilemap ops the edited-scene pinning (`EDITED_SCENE_MISMATCH`) protection and the richer not-found messages for free. Fine as a follow-up if you'd rather not touch it now.

---

dsarno (approved):
All review points are resolved — approving. ✅

- The rect-fill undo data-loss is fixed (`_restore_rect_snapshot` no longer clears the layer) with a regression test covering tiles outside the rect.
- The biome-specialized op was split out cleanly, leaving the generic tilemap/atlas/`viewport_2d` core.
- `source_id` now means raw TileSet source IDs consistently across the domain.
- Compressed atlas textures decompress (or error) instead of returning empty images, and the empty-PNG case errors.
- `viewport_2d` is documented in the tool contract, rejects unsupported params explicitly, and force-draws before capture.
- Both follow-up nits landed too: docstring indentation fixed and `_resolve_layer` now uses the shared `McpNodeValidator` (with test coverage), plus lint fixes and committed `.uid` files.

Verified locally on this head: `ruff check src/ tests/` clean and the full Python suite passing (1235 passed, 2 skipped); CI is fully green including the Godot-side suites on all platforms.

Thanks @saschiwy for the excellent, fast turnaround — this closes a real gap in 2D authoring support.

---

dsarno:
@saschiwy We merged this a few days ago and it'll be in the 3.0.0 release -- drop into our discord and give your username so we can give you a shoutout in release notes https://discord.gg/FDZ5fr2QkP

PR #637, opened 2026-07-04
scene_open: wait for the scene switch to land before replying (#633)

### LINKED ISSUE #633: scene_open returns success before the editor actually switches scenes; immediate follow-up writes hit the previous scene
Found during a live smoke of main @ 54f53cd (Godot 4.6.3, macOS).

## Symptom

`scene_open` replies with success as soon as it has *requested* the scene switch — `EditorInterface.open_scene_from_path()` completes the tab switch asynchronously. An agent that issues a follow-up write immediately can hit the **previous** scene.

Observed sequence (live editor, storm scene `res://_stormtest/storm.tscn` current):

1. `scene_open(path="res://_smoke/err_scene.tscn")` → success, `path: res://_smoke/err_scene.tscn`
2. `script_attach(path="/ErrRoot", ...)` immediately after → `NODE_NOT_FOUND: ... Scene root is "/Root"` (the storm scene's root — the switch hadn't landed)
3. `scene_save()` right after → **saved `res://_stormtest/storm.tscn`**, i.e. wrote the old scene to disk

Step 3 is the dangerous one: a write lands on the wrong scene with a success response. (In this case the old scene was gitignored scratch; with user scenes this could silently persist unintended state.)

A later `editor_state` call showed `current_scene` had switched by then — the window is a few frames wide, but at MCP call cadence it's easy to hit.

## Notes

- Readiness gating doesn't help: readiness stays `ready` during the pending switch.
- The `scene_file` guard (EDITED_SCENE_MISMATCH) already protects writes that pass it — but nothing sets it by default, and `scene_save` takes no target at all.

## Suggested fix

Have `open_scene` defer its response (dispatcher already supports `DEFERRED_RESPONSE`) until `EditorInterface.get_edited_scene_root()` reports `scene_file_path == path` (with a frame-budget/timeout fallback that returns `switched: false` if it never lands). Then a success reply means "the editor is now editing the scene you asked for", and follow-up writes are safe.

dsarno:
## Summary

Fixes #633 — `scene_open` replied with success as soon as it *requested* the tab switch, but `EditorInterface.open_scene_from_path()` completes the switch asynchronously. An agent issuing an immediate follow-up write could hit the **previous** scene. Observed live during a smoke: a `scene_save` issued right after `scene_open` saved the *old* scene to disk, and a `script_attach` resolved against the old scene's root.

## Fix

`open_scene` now defers its reply (via the dispatcher's existing `DEFERRED_RESPONSE` mechanism) until `EditorInterface.get_edited_scene_root().scene_file_path == path`. A success reply now means "the editor is now editing the scene you asked for", so follow-up writes are safe.

New response fields:
- `switched: true` / `settle: "settled"` — switch confirmed landed
- `settle: "already_current"` — requested scene was already edited (immediate, no async)
- `switched: false` / `settle: "timeout"` — switch not observed within 3s (re-check `editor_state` before writing)
- `switched: false` / `settle: "not_waited"` — synchronous fallback for `batch_execute` / no-connection contexts that can't await

The deferred coroutine is `static` (mirrors `FilesystemHandler._finish_scan_deferred`) so it survives the handler being freed by an `editor_reload_plugin` mid-await. Wait window (3s) stays under the dispatcher's 4500ms deferred default.

## Testing

Verified live on Godot 4.6.3 via MCP:
- Two throwaway scenes; `scene_open(B)` from A → `settle: settled, switched: true`; `editor_state` immediately after shows `current_scene` already switched.
- `scene_open` of the already-current scene → `settle: already_current`.
- `batch_execute([open_scene(A), create_node])` → open reports `not_waited`, and the node still lands on `/SceneA` (switch behavior preserved).
- New GDScript test `test_open_scene_already_current_reports_switched`.
- Full suite green via `test_run`: **0 failed, 1587 passed / 8 skipped, 57 suites**. `ci-check-gdscript`: all files OK.

## Summary by CodeRabbit

* **New Features**
  * Scene opening can now report completion after the editor finishes switching, including whether the target scene was already current, settled successfully, or timed out.

* **Bug Fixes**
  * Improved reliability by preventing follow-up actions from occurring before the editor actually switches scenes, with a clear fallback when waiting isn’t possible.

* **Tests**
  * Added a validation test to confirm that opening the already-current scene reports `switched=true` with `settle="already_current"` and the correct `previous_scene_path`.

* **Documentation**
  * Updated tool documentation to clarify response timing and how to interpret timeout results.

### DIFF
--- a/plugin/addons/godot_ai/handlers/scene_handler.gd
+++ b/plugin/addons/godot_ai/handlers/scene_handler.gd
@@ -143,6 +143,13 @@ func create_scene(params: Dictionary) -> Dictionary:
 	}
 
 
+## How long open_scene waits for the editor to actually switch to the
+## requested scene before replying switched=false. Tab switches normally land
+## within a few frames; keep this under the dispatcher's 4500 ms deferred
+## default so the coroutine always answers before DEFERRED_TIMEOUT fires.
+const _OPEN_SETTLE_MAX_MSEC := 3000
+
+
 ## Open an existing scene by file path.
 func open_scene(params: Dictionary) -> Dictionary:
 	var path: String = params.get("path", "")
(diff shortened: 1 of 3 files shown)


dsarno:
Addressed in 5c5b44a: the settle coroutine now requires a fresh edited-scene-root instance id (not just a path match), so force_reload waits for the actual rebuild instead of settling on the first frame. Verified live — force_reload now reports `settle: settled` only after the reload completes and correctly discards in-memory edits. Also documented the `settle: "not_waited"` sync-fallback case in the tool docstring.

Issue #635, opened 2026-07-04
Polish: in-run error scope labeling, input_mouse silent position fallback, test_run main.tscn precondition


dsarno:
Three small polish items from a live smoke of main @ 54f53cd (Godot 4.6.3). None are blockers; bundling them since each is a one-file change.

## 1. In-run push_error classified as `retained_recent` instead of run-scoped

A `push_error` fired 413 ms into a fresh `project_run` (in `_ready()` of the running scene) came back in the run response with `recent_errors_scope: "retained_recent"`, `recent_errors_may_predate_run: true`, and empty `current_run_errors` — even though the error demonstrably belongs to this run (the response's own `details.time` says `0:00:00:413`). Information is intact (the error is fully surfaced with stack), but the labeling tells an agent "this may be stale" about an error its own run just produced. Attribution should classify Errors-tab rows observed after run start as scope `run` → `current_run_errors`.

## 2. `game input_mouse` silently substitutes cursor position for malformed `position`

`game_manage(op="input_mouse", params={event: "motion", position: [100, 100]})` — an array instead of the expected `{x, y}` dict — is accepted, and `_dict_to_vector2` silently falls back to the **live mouse position** (response echoed `position: {x: -1.478, y: 0}`). Falling back when `position` is *omitted* is a reasonable default; doing the same for a present-but-wrong-shape value hides caller bugs. A present non-dict `position` should be rejected (or arrays `[x, y]` should be accepted as a coercion, matching the dict-or-array flexibility elsewhere in the surface).

## 3. `test_run` silently depends on main.tscn being the edited scene

Running the full GDScript suite with any other scene open produces ~79 phantom failures ("Scene root is /X", "Expected Main, got X", node-not-found cascades) because many suites assume `res://main.tscn` is the edited scene. Cost me a full debugging round during the smoke. Cheapest fix: `test_run` response (and/or McpTestRunner) includes the edited scene and a warning when it differs from `run/main_scene`; alternately test_handler could open the main scene before running when the project defines one.

dsarno:
All three items are now fixed on main: items 2 (`input_mouse` position validation with `[x,y]` coercion) and 3 (`test_run` edited-scene field + mismatch warning) landed in #638; item 1 (in-run error misclassified as `retained_recent` across an unscanned Errors-tab clear) landed in #657, merged today with tracker-level and end-to-end run-scoping tests. Closing as completed.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #518, opened 2026-06-03
editor_manage(game_eval): opaque INTERNAL_ERROR regressing since 2.5.10 (regression of #487)


dsarno:
## What telemetry is showing

The opaque `INTERNAL_ERROR` on `editor_manage(game_eval)` — the exact failure
mode fixed in #488 and closed as #487 — is **regressing in recent releases**.
It was driven down to ~1.4% of game_eval calls in the fix version (2.5.10),
but has climbed back over the two releases since, reaching **6.15% of
game_eval calls in 2.6.0** across **75 distinct installs**.

This is the same timeout-shaped, code-less `INTERNAL_ERROR` envelope #488
was meant to eliminate (distinct from the legitimate `EVAL_COMPILE_ERROR` /
`EVAL_RUNTIME_ERROR` caller-side codes added in #491, which are behaving as
intended, and distinct from the residual `TimeoutError` tail tracked in #500 —
that tail has effectively gone quiet in 2.6.0).

## Scope

- **Window**: last 7 days (cohort-compared back to 2.5.9 pre-fix baseline)
- **Metric**: `INTERNAL_ERROR` as a share of `editor_manage(game_eval)` calls, per version
- **Affected installs**: 75 (2.6.0 alone, last 7d)
- **Affected versions**: regression visible in 2.5.13 and 2.6.0; absent in 2.5.10/2.5.11
- **Platforms**: all three; most pronounced on Windows

## Per-version trend (exposure-normalized)

| Version | game_eval calls | `INTERNAL_ERROR` rate | Affected installs | Note |
|---|---|---|---|---|
| 2.5.9 | 9,853 | 13.55% | 166 | pre-fix baseline |
| **2.5.10** | 1,546 | **1.42%** | 9 | fix #488 — verified |
| 2.5.13 | 11,129 | 3.23% | 85 | regression begins |
| **2.6.0** | 4,604 | **6.15%** | 75 | regression continues |

## Per-platform (2.6.0, last 7d)

| Platform | game_eval calls | `INTERNAL_ERROR` | rate |
|---|---|---|---|
| Windows | 3,499 | 233 | 6.66% |
| Linux | 553 | 30 | 5.42% |
| Darwin | 568 | 23 | 4.05% |

All platforms are elevated relative to the 2.5.10 baseline (1.42%); Windows
is worst but the regression is not Windows-exclusive.

## Concentration

[...]

PR #638, opened 2026-07-04
Polish: reject malformed game mouse position; flag test_run scene mismatch (#635)

dsarno:
## Summary

Addresses two of the three items in #635 (both low-risk, found during a live smoke of main). The third is intentionally deferred — see below.

### 1. `game input_mouse` rejects malformed `position`

Previously a present-but-wrong-shaped `position` (an `[x, y]` array, or a bare number) silently fell through to the game's live cursor position, hiding caller bugs. Now:
- `{x, y}` object — accepted (unchanged)
- `[x, y]` array — accepted as a coercion (matches the dict-or-array flexibility elsewhere in the surface)
- omitted / null — falls back to the live cursor (a deliberate default, unchanged)
- present but any other shape (wrong-length array, scalar) — **rejected** with a clear error instead of silent substitution

### 2. `test_run` flags scene mismatch

Many GDScript suites assume the project main scene is the edited scene (they read `/Main/...` directly). Running with another scene open produces a flood of phantom failures that read like real regressions — this cost a full debugging round during the smoke. The response now includes:
- `edited_scene` — the scene currently open
- `scene_warning` — emitted only when `edited_scene` differs from `run/main_scene` **and** there are failures, pointing at `scene_open(main)` + re-run

## Deferred (item 3 of #635)

The in-run `push_error` labeled `recent_errors_scope: "retained_recent"` instead of run-scoped is **not** in this PR. It intersects the surfaced-error-tracker cursor bookkeeping that #518/#621 are built on, and a blind `retained_recent → run` flip risks regressing error surfacing. The error is already fully surfaced (stack + details); only the scope label is conservative. It warrants a focused change with its own verification. #635 stays open for it.

## Testing

Verified live on a freshly-restarted Godot 4.6.3 editor via MCP:
- 5 new `game_helper` tests for `_resolve_mouse_position` (dict, array coerce, absent-fallback, malformed-array reject, wrong-type reject) — all pass.
- 2 new `test_runner` tests for the scene annotation (`edited_scene` populated; warning suppressed when main scene is open) — all pass.
- Full suite green: **0 failed, 1594 passed / 7 skipped, 57 suites**; `edited_scene: res://main.tscn` present in the response.
- `pytest tests/unit -k "game or testing or tool"`: 77 passed. `ci-check-gdscript`: all files OK.

## Summary by CodeRabbit

* **New Features**
  * Test runs now include the currently open editor scene as `edited_scene`, and may add a `scene_warning` when failures occur while a non-main scene is open.
  * Mouse input can accept pointer positions as `{x, y}` objects or `[x, y]` arrays.
* **Bug Fixes**
  * Requests with malformed pointer position payloads are rejected with a clear error instead of silently using the current cursor position.
* **Documentation / Tests**
  * Updated tool documentation and added/expanded tests for scene annotation and pointer position resolution.

### DIFF
--- a/plugin/addons/godot_ai/handlers/test_handler.gd
+++ b/plugin/addons/godot_ai/handlers/test_handler.gd
@@ -30,7 +30,11 @@ func run_tests(params: Dictionary) -> Dictionary:
 				discovery.errors.size(),
 				", ".join(discovery.errors),
 			]
-		return {"data": {"error": msg, "total": 0, "load_errors": discovery.errors}}
+		var no_suites := {"error": msg, "total": 0, "load_errors": discovery.errors}
+		## Keep the edited_scene annotation on the no-suites error payload too,
+		## so the response contract is consistent across every return path.
+		_annotate_edited_scene(no_suites)
+		return {"data": no_suites}
 
 	var ctx := {
 		"undo_redo": _undo_redo,
--- a/plugin/addons/godot_ai/handlers/test_handler.gd
+++ b/plugin/addons/godot_ai/handlers/test_handler.gd
@@ -40,7 +40,30 @@ func run_tests(params: Dictionary) -> Dictionary:
 	var results := _runner.run_suites(suites, suite_filter, test_filter, ctx, verbose, exclude_test_filter)
 	if not discovery.errors.is_empty():
 		results["load_errors"] = discovery.errors
+	_annotate_edited_scene(results)
 	return {"data": results}
+
+
+## Many suites assume the project's main scene is the edited scene (they read
+## /Main/... nodes directly). Running with another scene open produces a flood
+## of phantom failures that look like real regressions. Surface the edited
+## scene and a warning when it differs from run/main_scene so the failures are
+## attributable at a glance instead of costing a debugging round (#635).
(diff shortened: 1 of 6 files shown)


dsarno:
Addressed in f9c9ab3: rewrote `test_annotate_scene_warns_on_mismatch_with_failures` to deterministically override `application/run/main_scene` (restored before asserting) so both the warn and no-warn branches run every time, not just the suppression branch. Also tightened the malformed-position tests to assert no fallback `position` is returned. Full suite green.

dsarno:
Addressed Copilot's two findings in e14bdaa:
- `_resolve_mouse_position` now rejects a non-empty dict lacking `x`/`y` (e.g. `{"foo":1}`) and non-numeric coordinates in either dict or array form (via a new `_is_number` guard), instead of silently falling back / coercing to 0.0. Empty `{}` still falls back (unspecified), and partial dicts with a numeric coordinate still work. Removed the now-dead `_dict_to_vector2`.
- `run_tests` now annotates the no-suites-found error payload with `edited_scene` too, so the response contract is consistent across every return path.

Verified live in play mode: `{"foo":1}` and `["a","b"]` are rejected with structured errors, `[120,45]` coerces. Full suite green (0 failed, 1598 passed).

PR #639, opened 2026-07-04
stormtest: fix stale params starving bind_event/logs_read; log every abort reason (#634)

### LINKED ISSUE #634: stormtest harness: stale params starve bind_event and logs_read of coverage; STOP reason never logged
Found while reconciling stormtest error buckets during a live smoke of main @ 54f53cd (default run 1371 calls + heavy run 6115 calls).

## 1. Stale param names starve two ops of real coverage

- `op_input_map` sends `{"event_type": "key", "key": "A"}` for `bind_event` — the op requires `keycode`, not `key`. Result: **26/26 bind_event calls failed** with MISSING_REQUIRED_PARAM in the heavy run. `bind_event` gets zero real stress coverage.
- `op_logs` draws `source` from `["plugin", "editor", "combined"]` — `"combined"` is not a valid source (valid: `plugin`, `game`, `editor`, `all`; `src/godot_ai/handlers/editor.py` raises ValueError). Result: **50/148 logs_read calls failed** as ToolError, and the `all` source path is never exercised.

Both make the storm report look noisier than it is and hide any real regressions in those paths.

## 2. STOP reason is not logged

The heavy run ended at 6115/~9000 calls with the summary saying `(STOP was set — see log above for why)` — but the silent STOP path prints nothing: in `worker_loop`, a worker whose reconnect attempt fails just sets `STOP[0] = True` and breaks. The only evidence is the missing calls. Every `STOP[0] = True` site should print a one-line reason (worker id + cause).

## Status

A fix for the two param names is already in progress in a separate local session (spun off as a task chip during the smoke). This issue tracks both items so whichever lands first can reference it; the silent-STOP logging may need a follow-up PR.

dsarno:
## Summary

Fixes #634. Two classes of problem in the stormtest harness, both found while reconciling error buckets during a live smoke.

### 1. Stale params starved two ops of coverage

- `op_input_map` sent `{"event_type": "key", "key": "A"}` for `bind_event`, but the op requires `keycode`. **26/26** bind_event calls failed `MISSING_REQUIRED_PARAM` in a heavy run — zero real coverage.
- `op_logs` drew `source` from `["plugin", "editor", "combined"]`; `"combined"` is not a valid source (`plugin|game|editor|all`). ~**50/148** logs_read calls failed `ToolError`, and the `all` path was never exercised.

Both fixed (`keycode`, `all`). Verified live: `bind_event` and `logs_read` now report **0 errors**.

### 2. Silent aborts

Several early-exit paths set `STOP[0]` with no reason logged, so a truncated run gave no cause — a heavy run ended at 6115/~9000 calls with the summary literally saying "see log above for why" while the log said nothing. Every genuine abort now routes through a new `_abort(reason)` helper that prints the cause and sets a distinct `ABORTED` flag. `STOP` alone is also set by the normal end-of-run teardown and by a user Ctrl-C (SIGINT), so the report/JSON now distinguish:
- clean finish → `aborted: false`, no "aborted early" line
- early failure → `aborted: true`, `!!! stormtest aborting: <reason>` printed, reason named in the summary

## Note on parallel work

A separate local session (spun off as a task chip during the smoke) began the two param-name edits but stopped without committing — its 2-line change sat uncommitted in a worktree. This PR includes those same two fixes plus the abort-logging (item 2 of #634), so it supersedes that unpushed work as the complete fix.

## Testing

- `python -m py_compile` + `ruff check`: clean.
- Live short run (4 workers × 3 waves, reload off) against Godot 4.6.3: `input_map_manage.bind_event` 1/0, `logs_read` 9/0, `final verdict EDITOR ALIVE`, JSON `stopped:false aborted:false`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of early test shutdowns so abort reasons are reported more clearly.
  * Final results now distinguish between a normal stop and an aborted run, with clearer messaging in the report.
  * JSON output now includes an explicit `aborted` status for easier result tracking.
  * Updated stress-test input values to better match current expectations (log source and event key name).

### DIFF
--- a/script/stormtest.py
+++ b/script/stormtest.py
@@ -117,8 +117,20 @@ REPORT_JSON = os.environ.get(
     "SS_REPORT", os.path.join(tempfile.gettempdir(), "stormtest_report.json")
 )
 START = [0.0]
-STOP = [False]  # set by signal handler / fatal server loss
+STOP = [False]  # set by signal handler / fatal server loss / normal teardown
+# ABORTED distinguishes an early failure abort from STOP being set by the
+# normal end-of-run teardown (the `finally` in main). Only the former should
+# print "aborted — see reason above"; a clean run also has STOP set.
+ABORTED = [False]
 ROOT_PATH = ["/Root"]  # resolved after scratch-scene creation
+
+
+def _abort(reason: str) -> None:
+    """Flip STOP + ABORTED and log why. Every early-exit failure path routes
+    through here so a truncated run always states its cause (#634)."""
+    print(f"  !!! stormtest aborting: {reason}")
+    STOP[0] = True
+    ABORTED[0] = True
 
 
 def _pct(vals: list[float], p: float) -> float:
@@ -157,6 +169,7 @@ def flush_report_json():
         "reload_recovery_s": [round(x, 1) for x in M["reload_recovery_s"]],
         "reconnects": M["reconnects"],
         "stopped": STOP[0],
+        "aborted": ABORTED[0],
         "overall_latency_ms": _lat_stats(all_lat),
         "err_codes": dict(M["err_codes"].most_common()),
         "per_op": {
@@ -324,7 +337,7 @@ async def op_logs(w: Worker):
         "logs_read",
         {
             "count": random.choice([10, 50]),
-            "source": random.choice(["plugin", "editor", "combined"]),
+            "source": random.choice(["plugin", "editor", "all"]),
         },
     )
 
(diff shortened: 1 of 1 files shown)


dsarno:
Addressed in e2fd055: the verdict now keys off the end-of-run liveness probe (`STOP[0]` at report time) instead of `M["ok"]`, so prior successes no longer mask a dead editor. Verified across all four (STOP, ABORTED, ok) combinations — prior-success + dead-final-probe now correctly reports DEAD/UNREACHABLE.

PR #640, opened 2026-07-05
Hermes Agent support (YAML config + working HTTP integration)

jb0gie:
## Summary

Adds first-class **Hermes Agent** (Nous Research) support to the Godot AI plugin, configuring it as a native HTTP MCP server from the Godot AI dock.

The previous revision of this PR assumed Hermes used JSON (`~/.hermes/mcp.json`, `mcpServers`, `type: streamable-http`). It doesn't — Hermes stores MCP config as **YAML** at `~/.hermes/config.yaml` under the snake_case `mcp_servers:` key, with HTTP entries that are just `url:` (transport inferred, no `type` field). Verified against the official docs: https://hermes-agent.nousresearch.com/docs/user-guide/features/mcp

This revision replaces that with a working implementation, verified **end-to-end** through the live godot-ai MCP server → editor plugin → real config file.

## Changes

- **`clients/hermes.gd`** — corrected descriptor: `config_type = "yaml"`, path `$LOCALAPPDATA/hermes/config.yaml`, `server_key_path = ["mcp_servers"]`, `entry_url_field = "url"`. No `type` field (transport inferred).
- **`clients/_yaml_strategy.gd`** (new) — minimal YAML upsert scoped to `mcp_servers`. No general YAML dependency (Godot has none in stdlib; Hermes only needs this one shape).
- **`client_configurator.gd` / `_base.gd` / `_manual_command.gd`** — wire the `"yaml"` config type into configure/remove/status/manual-command.
- **`clients/_registry.gd`** — register the `hermes` client.
- **`test_project/tests/test_clients.gd`** — Hermes tests assert the real YAML shape and round-trip (preserves unrelated keys like `model:`).

## Bugs fixed during live testing (commits on top of the original descriptor)

1. `PackedStringArray` (not `Array[String]`) for `String.split()` results — Godot 4 type mismatch crashed `configure`.
2. Sibling `mcp_servers` entries parsed by indent level, not just top-level keys (2-space-indented siblings were being nested under the first entry).
3. `$LOCALAPPDATA` (the token `_path_template.expand()` actually substitutes) instead of `%LOCALAPPDATA%` (left as a literal invalid path, so the write landed in a bogus location and the verify false-passed).
4. Space indentation — YAML forbids tabs.
5. `build_entry` emits the clean HTTP-native shape `{ url: ... }` instead of preserving stale `command/args/env` stdio-bridge keys.

## Verification

- `client_manage configure hermes` → `{"status": "ok"}`, writes `godot-ai: { url: http://127.0.0.1:8000/mcp }` to the real `~/.hermes/config.yaml`.
- Resulting YAML parses cleanly (PyYAML); `status` reports `configured`.
- **Hermes natively connects**: `hermes mcp test godot-ai` → `✓ Connected (110ms)`, `✓ Tools discovered: 43`.
- Full read/write/delete round-trip through the live editor (godot-ai 2.9.1 + Godot 4.7) confirmed working.

## Notes

- The GDScript `test_clients.gd` suite runs in CI (Godot 4.7); I validated the YAML logic via a faithful port + the live editor round-trip locally.
- `op_typo_hint.py` includes a small unrelated fix: the middleware now catches FastMCP's wrapped `ValidationError` so malformed tool calls return an actionable hint instead of crashing. Happy to split it into a separate PR if preferred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 [![Godot Asset Library](https://img.shields.io/badge/Godot-Asset%20Library-478cbf?logo=godotengine&logoColor=white)](https://godotengine.org/asset-library/asset/5050)
 [![Discord](https://img.shields.io/badge/Discord-Join%20chat-5865F2?logo=discord&logoColor=white)](https://discord.gg/FDZ5fr2QkP)
 
-**Connect MCP clients directly to a live Godot editor** via the [Model Context Protocol](https://modelcontextprotocol.io/introduction). Over **120 ops across ~41 MCP tools** ([full list](docs/TOOLS.md)) let AI assistants (Claude Code, Codex, Antigravity, etc.) build scenes, edit nodes and scripts, wire signals, and configure UI, materials, animations, particles, cameras, and environments.
+**Connect MCP clients directly to a live Godot editor** via the [Model Context Protocol](https://modelcontextprotocol.io/introduction). Over **120 ops across ~41 MCP tools** ([full list](docs/TOOLS.md)) let AI assistants (Claude Code, Codex, Antigravity, Hermes Agent, etc.) build scenes, edit nodes and scripts, wire signals, and configure UI, materials, animations, particles, cameras, and environments.
 
 > 🎉 **Now on the [Godot Asset Library](https://godotengine.org/asset-library/asset/5050) and the [new Godot Asset Store](https://store.godotengine.org/asset/dlight/godot-ai/)** — one-click install from Godot's **AssetLib** tab. You'll still need [uv](https://docs.astral.sh/uv/) for the Python server (see [Quick Start](#quick-start)).
 
@@ -77,7 +77,7 @@ The dock lists every supported client with a status dot and per-row
 **Configure** / **Remove** buttons, or press **Configure all**. Auto-configure
 covers:
 
-- **Claude Code**, **Claude Desktop**, **Antigravity**
+- **Claude Code**, **Claude Desktop**, **Antigravity**, **Hermes Agent**
 
 <details>
 <summary><strong>…and 16+ more clients</strong></summary>
--- a/plugin/addons/godot_ai/client_configurator.gd
+++ b/plugin/addons/godot_ai/client_configurator.gd
@@ -20,6 +20,7 @@ const Client := preload("res://addons/godot_ai/clients/_base.gd")
 const ClientRegistry := preload("res://addons/godot_ai/clients/_registry.gd")
 const JsonStrategy := preload("res://addons/godot_ai/clients/_json_strategy.gd")
 const TomlStrategy := preload("res://addons/godot_ai/clients/_toml_strategy.gd")
+const YamlStrategy := preload("res://addons/godot_ai/clients/_yaml_strategy.gd")
 const CliStrategy := preload("res://addons/godot_ai/clients/_cli_strategy.gd")
 const ManualCommand := preload("res://addons/godot_ai/clients/_manual_command.gd")
 const CliFinder := preload("res://addons/godot_ai/clients/_cli_finder.gd")
@@ -283,6 +284,8 @@ static func _dispatch_configure(client: Client, url: String) -> Dictionary:
 			return JsonStrategy.configure(client, SERVER_NAME, url)
 		"toml":
 			return TomlStrategy.configure(client, SERVER_NAME, url)
+		"yaml":
+			return YamlStrategy.configure(client, SERVER_NAME, url)
 		"cli":
 			# #463: fall back to writing the config file directly when the CLI
 			# binary isn't on PATH (Claude Code as a VS Code/Cursor extension).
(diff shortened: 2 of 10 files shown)


Clubhouse1661:
@jb0gie Thanks for putting this together. The registry wiring and test coverage are in good shape, but I'd like to hold merging until the config target matches how Hermes actually stores MCP config.

Hermes reads MCP config from `~/.hermes/config.yaml` under a `mcp_servers` key (YAML, snake_case), not `~/.hermes/mcp.json` under `mcpServers`:
https://hermes-agent.nousresearch.com/docs/user-guide/features/mcp

Godot AI's client configurator currently only has `json`, `toml`, and `cli` strategies, and HTTP entries in Hermes use `url` + `headers` with transport inferred (no `type` field). So this can't be fixed by adjusting descriptor fields. Making Hermes first-class needs a YAML strategy (or a Hermes-specific strategy) that reads/writes the shape Hermes expects. Happy to help scope that.

jb0gie:
awesome yeah this is really helpful feedback ☺️ yeah sure thing tbh I saw green and got excited this is very inciteful!!

jb0gie:
Updated with a working implementation. The original descriptor assumed JSON (`~/.hermes/mcp.json`, `mcpServers`, `type: streamable-http`), but Hermes Agent actually stores MCP config as **YAML** at `~/.hermes/config.yaml` under the snake_case `mcp_servers:` key, with HTTP entries that are just `url:` (transport inferred, no `type` field). Verified against the official docs: https://hermes-agent.nousresearch.com/docs/user-guide/features/mcp

The corrected code (branch `hermes-support` → https://github.com/jb0gie/godot-ai/tree/hermes-support) now:
- adds a minimal YAML strategy (`clients/_yaml_strategy.gd`) scoped to `mcp_servers` — no general YAML dep (Godot has none in stdlib)
- writes the correct Hermes shape: `godot-ai: { url: http://127.0.0.1:8000/mcp }`
- fixed several bugs found while testing live through the godot-ai MCP server → editor plugin → real config file:
  - `PackedStringArray` (not `Array[String]`) for `String.split()` results — Godot 4 type mismatch crashed `configure`
  - sibling `mcp_servers` entries must be parsed by indent level, not just top-level keys
  - `$LOCALAPPDATA` (the token `_path_template.expand()` actually substitutes) instead of `%LOCALAPPDATA%` (left as a literal invalid path, so the write landed in a bogus location and the verify false-passed)
  - space indentation (YAML forbids tabs)

Verified end-to-end against godot-ai 2.9.1 + Godot 4.7 (ProjectBoost): `client_manage configure hermes` produces the correct entry and `status` reports `configured`; the resulting `config.yaml` parses as valid YAML and Hermes loads `godot-ai` as a native HTTP sub-MCP. Full read/write/delete round-trip through the live editor also confirmed working.

Note: the GDScript `test_clients.gd` suite runs in CI (Godot 4.7); I validated the YAML logic via a faithful port + the live editor round-trip since there's no headless Godot in my local env.

@jb0gie the branch is on your `jb0gie/godot-ai` fork (`hermes-support`) — push it to `hi-godot/hermes-support` (or re-target this PR) whenever you want it on this PR. The PR currently points at the older `0eb174d` json-based commit; the new commits sit on top of it.

dsarno (changes requested):
The overall approach is sound (scoped YAML upsert rather than a general parser, verified end-to-end against a live Hermes), and `"unix"` as a path_template key is supported by `_path_template.resolve()`. But there are two definite breakages in `test_clients.gd` and a couple of smaller items:

**1. `test_hermes_windows_path_template_uses_appdata` contradicts the descriptor — it will fail.**
The test asserts the windows template contains `%APPDATA%`, but `hermes.gd` deliberately uses `$LOCALAPPDATA/hermes/config.yaml` (and its own comment explains why `%VAR%` syntax is wrong and why it's LocalAppData, not Roaming). This looks like a stale test from the pre-fix revision. It should assert `$LOCALAPPDATA` and `config.yaml` instead.

**2. The `test_opencode_client_uses_home_config_on_windows` function header was deleted, orphaning its body.**
The last hunk replaces the line `func test_opencode_client_uses_home_config_on_windows() -> void:` with the new Hermes block, but leaves the OpenCode test's docstring and body in place — they now dangle inside/after `test_hermes_yaml_roundtrips_through_configure`. That either breaks the parse or silently swallows the OpenCode regression coverage. Please restore the function header.

**3. Formatting/comment loss inside and around the `mcp_servers` block.**
`_extract_block` uses `text.split("\n", false)` — `allow_empty=false` drops *every* blank line in the file (prefix and suffix included), and comments inside the `mcp_servers` block are discarded on re-emit. The file-level docstring promises the rest of config.yaml is "preserved verbatim by round-tripping the raw lines" — with `split(..., false)` that's not true. Switching to `split("\n", true)` (and keeping the existing trailing-blank trim in `_assemble`) would fix the blank-line loss cheaply; comment loss within the block is probably acceptable if documented honestly.

**4. Minor:** `_assemble` re-checks `_text.contains("mcp_servers:")` — a commented-out `# mcp_servers:` line would falsely take the "block existed" path. You already computed `header_idx` in `_extract_block`; passing that through would be more robust than re-scanning the raw text.

On the bundled `op_typo_hint.py` fix: it's fine functionally, but as you offered — it's unrelated to Hermes and would be cleaner as its own small PR with a test for the FastMCP-wrapped path.

Happy to re-review quickly once 1 and 2 are fixed; those are the blockers.

---

dsarno:
Thanks @jb0gie — Hermes Agent support is merged for the upcoming release! 🎉 If you'd like a shoutout in the release announcement, drop into our Discord (https://discord.gg/FDZ5fr2QkP) and share your username.

jb0gie:
joined the discord 
also let's good I'm glad its working

PR #643, opened 2026-07-06
logs_read: forward since_run_id so retained prior-run game logs are reachable

dsarno:
## Problem

`logs_read` in `src/godot_ai/handlers/editor.py` built the `get_logs` params as `{count, offset, source}` and never forwarded `since_run_id`. The plugin side has supported it since ~#572 (`_get_game_logs` → `get_run_page(target_run_id, ...)`), and the tool docstring promises "pass `since_run_id` from an earlier response to read that prior run" — but because the param was dropped, the plugin always served the current run, the python `run_id != since_run_id` guard always fired, and every prior-run read came back as the empty `stale_run_id=true` shape. Retained prior-run reads were unreachable through MCP.

Found during live smoke of #642; pre-existing on `main`.

## Fix

- Forward `since_run_id` in the `get_logs` params for `source="game"`.
- Propagate the plugin's `stale_run_id` instead of hardcoding `false` — a retained-run read now returns the lines **with** `stale_run_id=true`, as documented.
- Pass `current_run_id` through (already promised by the tool docstring).
- Keep the `run_id != since_run_id` guard, repurposed: the new plugin echoes the requested id back as `run_id`, so a mismatch now only means an older plugin ignored the param and served the current run — the guard prevents those lines from being mislabeled as the requested run. Genuinely unknown/evicted run ids flow through normally as an empty page with `stale_run_id=true` (the plugin's own shape for them).

No plugin-side changes — the GDScript side was already correct.

## Tests

- Unit (`tests/unit/test_runtime_handlers.py`): StubClient now mirrors the plugin's retained-ring semantics; new cases cover the retained-run read (asserting the param is actually on the wire, since dropping it was the bug), `since_run_id` matching the current run, an unknown/evicted run id, and the old-plugin fallback.
- Integration (`tests/integration/test_mcp_tools.py`): new test asserting `since_run_id` reaches the plugin over the real MCP stack and the retained run round-trips with `current_run_id` and `stale_run_id=true`. The existing stale-shape test survives unchanged as the old-plugin case.

Full suite: 1245 passed, 4 skipped.

## Relation to #642

Independent fixes with one trivial textual overlap: both add `current_run_id` to the passthrough tuple in `logs_read` (#642 also adds `helper_live`/`session_active`/`game_status`/`editor_errors_count`/`editor_errors_hint`), and #642 inserts a test adjacent to ones touched here. Whichever merges second needs a one-line conflict resolution (dedupe the tuple entry).

## Summary by CodeRabbit

* **New Features**
  * Game log requests now honor a run boundary parameter to resume or inspect retained logs from prior runs.
  * Log responses can include additional run metadata (including the current run identifier) for clearer paging context.

* **Bug Fixes**
  * Improved stale detection behavior: stale results are returned with empty lines and `stale_run_id` correctly flagged.
  * Enhanced older-plugin compatibility and ensured stale flags are consistently represented.

* **Tests**
  * Added/expanded integration and unit tests covering run-boundary forwarding, unknown-run handling, and stale/current run metadata behavior.

### DIFF
--- a/src/godot_ai/handlers/editor.py
+++ b/src/godot_ai/handlers/editor.py
@@ -234,6 +234,11 @@ async def logs_read(
     ## ring buffer's run_id, dropped_count, and is_running stay
     ## authoritative on the editor side.
     params = {"count": count, "offset": offset, "source": source}
+    if source == "game" and since_run_id:
+        ## The plugin resolves since_run_id against its retained ring
+        ## (get_run_page), returning the prior run's lines with
+        ## stale_run_id=true when the id is no longer current.
+        params["since_run_id"] = since_run_id
     if source == "editor" and since_cursor is not None:
         params["since_cursor"] = since_cursor
     if include_details:
@@ -244,9 +249,10 @@ async def logs_read(
     )
     run_id = result.get("run_id", "")
     if since_run_id and run_id and run_id != since_run_id:
-        ## A new game run has started since the caller's last poll —
-        ## tell them to reset their cursor instead of returning stale
-        ## lines from the previous play session.
+        ## The plugin echoes the requested run id back as `run_id` when it
+        ## honors since_run_id, so a mismatch means an older plugin ignored
+        ## the param and served the current run. Return the empty stale
+        ## shape rather than mislabeling current-run lines as the old run.
         return {
             "source": source,
             "lines": [],
@@ -273,9 +279,10 @@ async def logs_read(
         "run_id": run_id,
         "is_running": result.get("is_running", False),
         "dropped_count": result.get("dropped_count", 0),
-        "stale_run_id": False,
+        "stale_run_id": bool(result.get("stale_run_id", False)),
     }
     for key in (
+        "current_run_id",
         "cursor",
         "oldest_cursor",
         "next_cursor",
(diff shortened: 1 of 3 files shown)


dsarno:
The two failing Python CI jobs are unrelated to this diff: CI now resolves fastmcp 3.4.3 (released after main's last green run on 3.4.2), which wraps pydantic's `ValidationError` in fastmcp's own (fastmcp #4128) and breaks the op-typo "Did you mean" middleware hint — `test_op_typo_surfaces_did_you_mean_message` fails on any fresh-dependency run, including a re-run of `main` today. #642 already carries the middleware fix (`op_typo_hint.py` unwraps `__cause__`) and its CI is green. Plan: merge #642 first, then rebase this branch (one-line dedupe of `current_run_id` in the `logs_read` passthrough tuple) and re-run CI.

dsarno:
Taking this over to land it: #642 just merged (883db7c), which both fixes the two failing Python checks here (the fastmcp 3.4.3 op-typo regression) and rewrote the same `logs_read` paths this PR touches. I'll merge main in, reconcile, address the CodeRabbit note about `current_run_id` in the stale fallback, and merge once CI is green.

dsarno:
Heads-up before you reconcile: the CodeRabbit stale-fallback note is already addressed on this branch — `0bc0638` preserves `current_run_id` in the mismatch/stale shape when the plugin sends it, with unit tests for both the field-present and old-plugin (field-absent) cases. Remaining work should just be the main merge itself: #642 and this branch both add `current_run_id` to the `logs_read` passthrough tuple (dedupe to one entry, keep #642's extra keys) and both add tests adjacent to `test_logs_read_handler_since_run_id_*` in `tests/unit/test_runtime_handlers.py`. Holding off on further pushes from my side.

Issue #641, opened 2026-07-05
Agents don't see Stack Trace Parse Error


Pti-Tza:
Sometimes, when launching the project, parse errors appear in the stack trace, but the agent does not detect them because they are not present in the logs.

dsarno:
@Pti-Tza Thanks for flagging.  Fixed in latest 2.9.1 👍🏼

PR #646, opened 2026-07-06
Surface boot-time debugger breaks that were invisible to every MCP surface (#645)

### LINKED ISSUE #645: Launch parse errors that break into the debugger are invisible to every MCP surface
## Symptom

Follow-up to #641 / #642, found during the live smoke of #642 (Godot 4.6.3-stable, macOS).

When an editor-launched game hits a GDScript parse error during boot — a broken script attached to a scene node, a broken autoload, root or non-root — the game process calls `GDScriptLanguage::debug_break_parse` and freezes in a remote-debugger break **before the game helper goes live** (confirmed by sampling the hung process: parked in `RemoteDebugger::debug` under `GDScript::reload`).

In that state there is **no record anywhere** for the #642 machinery to surface:

- no Debugger Errors-tab row (the break happens before the error message is flushed to the editor),
- no editor Logger entry,
- no game-log line (helper logger never attached),
- `project_run` resolves not_live with the generic "may still be booting or may have failed silently" message (it names a parse error only if *retained pre-run* errors happen to correlate, labeled `may_predate_run`).

The error is visible **only** in the Debugger's Stack Trace panel — which matches the original #641 report wording ("parse errors appear in the stack trace") and is scraped by nothing. The #642 mechanisms (live-message errors, game-log hint, timer-driven watermark) all work, but only when an error *record* exists; this path produces none.

## Repro

1. Attach a script containing `func f(:` to any node in the main scene (or register it as an autoload).
2. `project_run` via MCP.
3. Game hangs pre-live in the debugger break; every MCP surface reads clean. `sample <game-pid>` shows the `debug_break_parse` stack.

## Proposed direction

Capture the break editor-side in `McpDebuggerPlugin`: a `debug_enter`/breaked notification arriving while the run is not yet live is almost certainly a boot-time script failure — record it (message + stack when available) into the surfaced-error tracker so `project_run`'s liveness decision can report a run-scoped parse failure instead of the generic not-responding message, and consider auto-continuing or flagging the frozen-at-break state explicitly in `game_status` (e.g. `status: "break"`), since agents currently can't distinguish it from a slow boot.

Full smoke evidence: PR #642 discussion / smoke report (2026-07-05).

dsarno:
Fixes #645.

## Problem

On Godot 4.6.3, an editor-launched game that hits a GDScript parse error during boot (broken scene-node script or autoload) calls `GDScriptLanguage::debug_break_parse` and freezes in a remote-debugger break **before the game helper goes live**. No record reaches any MCP surface — no Errors-tab row, no editor Logger entry, no game-log line — so `project_run` resolved `not_live` with the generic "may still be booting or may have failed silently" message. The error text existed only in the Debugger's Stack Trace panel. (Found in the #642 live smoke; this was the core #641 flavor that remained unreachable.)

## Feasibility spike (phase 1)

Empirically established on 4.6.3, driving the break with the issue's repro (`func f(:` attached to a `main.tscn` node):

- `EditorDebuggerSession.breaked(can_debug)` carries **no reason text** (`can_debug=false` for parse errors), but the underlying `ScriptEditorDebugger` node's own `breaked(reallydid, can_debug, reason, has_stackdump)` signal — located by walking the editor UI, same approach as the existing Errors-tab scrape — delivers the full `"Parser Error: Expected parameter name."` ~4 ms later. So: signal, not scrape, for the reason.
- Stack frames are not exposed by any session API, but the Stack Trace `Tree` rows carry locale-independent metadata dicts `{frame, file, function, line}`. The panel is **empty at signal time** and populated by ~0.5 s (the editor requests the stack dump separately), so frames are scraped on short timers.
- The break fires at ~0.7 s with `game_ready=false` — well before the helper hello, and the game reuses the debugger session created at plugin load, so signal connections made in `_setup_session` persist across runs.

## Implementation (phase 2)

**`McpDebuggerPlugin`**
- Connects `session.breaked`/`continued` per session plus `ScriptEditorDebugger.breaked` for the reason text; tracks break state (`reason`, `can_debug`, `pre_live`), cleared on continue/run-end/new-run.
- On a **pre-live** break, scrapes the Stack Trace panel at 0.5 s / 2 s and synthesizes an Errors-tab-shaped record via the new `McpSurfacedErrorTracker.record_synthetic_error`, joining the same promoted sequence as scraped Debugger rows — so run-scoping, `logs_read(source="editor")`, the retained fallback, and the response watermark all surface it with no extra plumbing.
- `game_status` gains **`status="break"`** (takes precedence over `"live"`; `helper_live=false`, `session_active=true`) with `break={reason, can_debug, pre_live}`.
- Game-tool ready-wait loops bail out early on a break — a parked game never beacons. `game_manage(op="get_scene_tree")` against a parked game went from an opaque **15 s timeout to ~65 ms** with the actionable break error.

**`project_handler`**
- The liveness decision resolves a pre-live break once the synthesized record correlates (ready window as fallback), naming the failing script and instructing `project_manage(op="stop")`. Policy per the issue: **report + leave the game parked**. Post-live breaks (user breakpoints) get resumable wording instead of boot-failure framing.

**Docs**: `docs/TOOLS.md` status list, `project_run`/`editor_state` docstrings.

## Two bugs the live smoke caught that unit tests could not

1. The synthesized record was missing from `logs_read(source="editor")` — `collect_editor_log_entries` read only the Logger ring + live Errors tab. Fixed by merging synthetic promoted entries (promotion bookkeeping keys stripped from responses).
2. The break early-exit made the not-ready error reply **synchronous**, but the dispatcher registers a deferred request only *after* the handler returns `DEFERRED_RESPONSE` — the reply was dropped as `expired` and the server timed out at 15 s. Fixed by yielding one `process_frame` before any reply in the three wait coroutines.

## Live verification (Godot 4.6.3, macOS)

Repro loop run end-to-end twice (fresh editor per plugin build, raw streamable-http driver, session verified via `session_manage`):

- `project_run` resolves at ~1.2 s with `game_status.status="break"`, `break.reason="Parser Error: Expected parameter name."`, run-scoped `recent_errors` carrying `res://smoke_broken.gd:2` + frames, and the stop instruction in `reason`.
- `editor_state` and `logs_read(source="game")` report `status="break"` (game hint names the parse error); `logs_read(source="editor", include_details=true)` shows the synthesized record with frames and no `_debugger_*` leaks.
- `project_manage(op="stop")` recovers cleanly (`stopped=true`, status → `"stopped"`).

## Tests

- 17 new GDScript tests (break state machine, precedence over live, signal-notice merge, record synthesis ± frames, post-live non-synthesis, clear-on-continue/run-end, `_explain_not_live` wording, tracker promote/re-promote/refresh-survival, `logs_read` inclusion, liveness-decision branches, already-running message). Full suite: **1631 passed, 0 failed**.
- Python: **1255 passed** (docstring-only changes; `project_run`/`editor_state`/`logs_read` responses pass plugin fields through).
- Headless `--import` parse check clean.

## Summary by CodeRabbit

* **New Features**
  * Added clearer handling for projects paused by a debugger break, including more detailed status metadata and targeted stop/resume/relaunch guidance.
* **Bug Fixes**
  * Prevented screenshot, eval, and command requests from hanging while the game is paused at a break.
  * Improved break-state precedence and synthetic error promotion so break-related issues surface reliably.
* **Documentation**
  * Updated tool help/docs to explain the new `break` status and recommended next steps.
* **Tests**
  * Expanded coverage for boot-time debugger breaks, synthetic error promotion, and break-state lifecycle/clearing.

### DIFF
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -52,7 +52,16 @@ if a run-scoped parse/load error appeared, the response names it and points to
 game launched but this project has no `_mcp_game_helper` autoload, as with some
 headless/custom-main-loop setups: `helper_live=false` while
 `session_active=true`. `"launching"` is a soft "not live yet" state and can
-reconcile on a later `editor_state` poll.
+reconcile on a later `editor_state` poll. `"break"` means the game process is
+parked in a remote-debugger break: during boot this is a GDScript parse/load
+error (`GDScriptLanguage::debug_break_parse`) that freezes the game before the
+helper can register and before any error record reaches the Errors tab, the
+editor Logger, or the game log (#645). The plugin synthesizes an error record
+from the debugger break (reason + Stack Trace frames) so the `project_run`
+response names the failing script, and `game_status.break` carries `{reason,
+can_debug, pre_live}`. The run cannot continue on its own — call
+`project_manage(op="stop")`, fix the error, and relaunch. `helper_live=false`,
+`session_active=true` while at a break.
 
 `editor_state` includes the same `game_status` object in addition to the legacy
 `is_playing` boolean and `game_capture_ready`. It also mirrors
@@ -84,7 +93,7 @@ Game and combined log responses also include `game_status`, `helper_live`, and
 `game_status.session_active`. For compatibility, `is_running` is retained as an
 alias of `session_active`; it is no longer raw editor play-state. Both are `false` for
 `game_status.status` of `"not_live"` or `"stopped"`, and `true` for `"live"`,
-`"launching"`, or `"no_helper"`. This lets a parse/load failure that leaves the
+`"launching"`, `"break"`, or `"no_helper"`. This lets a parse/load failure that leaves the
 editor play button active report as not running, while a legitimate headless or
 custom-main-loop project without `_mcp_game_helper` remains active with
 `helper_live=false`, `session_active=true`, and
--- a/plugin/addons/godot_ai/debugger/mcp_debugger_plugin.gd
+++ b/plugin/addons/godot_ai/debugger/mcp_debugger_plugin.gd
@@ -87,6 +87,28 @@ var _game_run_started_msec := 0
 var _game_run_started_editor_cursor := 0
 var _game_run_started_debugger_cursor := 0
 var _game_helper_expected := true
+
+## #645: a GDScript parse error hit while an editor-launched game boots calls
+## GDScriptLanguage::debug_break_parse — the game parks in a remote-debugger
+## break BEFORE the helper's mcp:hello and before any record reaches the
+## Errors tab, the editor Logger, or the game log. The only editor-side traces
+## are the debugger break signals; the stack frames land in the Stack Trace
+## panel a few frames later. Track the break here so game_status can report
+## status="break" and a synthesized error record can name the failure.
+var _break_active := false
+var _break_can_debug := false
+var _break_reason := ""
+var _break_pre_live := false
+var _break_run_token := -1
+var _break_record_synthesized := false
+
+## #645: how long after the break signal to scrape the Stack Trace panel for
+## frames. The editor requests the stack dump from the game separately, so the
+## panel is empty at signal time; ~0.5s has it populated. The late tick
+## synthesizes with whatever is available so a scrape failure still yields a
+## record carrying the break reason.
+const BREAK_FRAME_SCRAPE_DELAYS_SEC: Array[float] = [0.5, 2.0]
+
 signal game_ready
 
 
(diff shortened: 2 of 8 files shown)


dsarno:
Both review nitpicks addressed in 1ea0799:

- **resolve flow**: moved the `correlated_error or elapsed_msec >= ready_wait_msec` computation into the `pre_live` branch, so the post-live branch's unconditional `resolve = true` is no longer an override.
- **retained errors in the already-running break message**: a `retained_recent` error is now named with the "may predate this run" caveat (parity with the `not_live` case) instead of falling through to the generic message; covered by `test_run_project_already_running_break_message_labels_retained_errors`.

Full GDScript suite re-run locally: 1632 passed, 0 failed.

PR #664, opened 2026-07-11
Don't tear down the editor session on a single malformed WebSocket frame (#526)

### LINKED ISSUE #526: [robustness] A single malformed WebSocket frame tears down the whole editor session
**Severity:** Medium · From pre-release robustness review (2026-06-09, `be40e1c`, v2.6.0)

**Location:** `src/godot_ai/transport/websocket.py:210,218,234`

**Risk:** Inside the per-message loop, a non-JSON frame or a command response that fails Pydantic validation raises out of the loop, is caught by the catch-all at `:234`, and the `finally` unregisters the session and drops the connection. The event path handles this correctly (local `ValidationError` catch at `:287`); the command path does not, so one garbled frame drops the editor↔server bridge instead of skipping the frame.

**Fix:** Wrap the per-message parse/validate in a local `try/except (json.JSONDecodeError, ValidationError)` that logs and `continue`s, mirroring the event path.

dsarno:
Fixes #526.

## Problem

In the per-message loop of `_handle_connection`, a non-JSON frame or a command response failing Pydantic validation raised out of the loop into the catch-all, and the `finally` block unregistered the session and dropped the connection — one garbled frame took down the whole editor↔server bridge. The event path already handled this correctly (local `ValidationError` catch); the command path did not.

## Fix

- `json.loads` and `CommandResponse.model_validate` each get a local catch (`json.JSONDecodeError` / `ValidationError`) that logs a warning (same `exc.errors(include_url=False, ...)` idiom as the event path) and skips the frame. **Only those two exception types are swallowed** — everything else still propagates to the existing catch-all/finally.
- Pending-future handling: if the malformed frame carries a string `request_id` matching a pending command, that future is failed immediately with a clear `ConnectionError` instead of leaving the caller to eat the 5s timeout. `send_command`'s `finally`-pop makes this race-safe (pop is idempotent). No usable `request_id` → normal timeout, the safe default.

## Tests

New `TestMalformedFrameResilience` (3 tests) on the existing harness: non-JSON frame skipped + session survives + subsequent command round-trips; invalid CommandResponse skipped + session survives; malformed frame with a valid pending `request_id` raises `ConnectionError` fast, session survives.

Full suite: **1271 passed, 2 pre-existing skips**; `ruff` clean. Python-transport-only — no GDScript surface.

https://claude.ai/code/session_017bCB9GDsZBp9wk7hjCGeu2

---
_Generated by [Claude Code](https://claude.ai/code/session_017bCB9GDsZBp9wk7hjCGeu2)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved WebSocket session stability when malformed or invalid frames are received.
  - Invalid messages are now ignored without disconnecting the editor session.
  - Malformed command responses with a request ID now fail immediately with a clear connection error instead of timing out.
  - Valid command and response exchanges continue working after invalid frames.

### DIFF
--- a/src/godot_ai/transport/websocket.py
+++ b/src/godot_ai/transport/websocket.py
@@ -207,7 +207,21 @@ class GodotWebSocketServer:
                 if live is not None:
                     live.touch()
 
-                data = json.loads(raw_msg)
+                ## Parse/validation failures of a single frame must not tear
+                ## down the whole editor session (#526): before this guard,
+                ## one garbled frame raised out of the loop into the
+                ## catch-all, and the finally block unregistered the session
+                ## and dropped the editor↔server bridge. Skip the bad frame
+                ## with a warning instead, mirroring the event path's
+                ## ValidationError handling. Only parse/validation errors are
+                ## swallowed — everything else still propagates.
+                try:
+                    data = json.loads(raw_msg)
+                except json.JSONDecodeError as exc:
+                    logger.warning(
+                        "Dropping non-JSON frame from session %s: %s", session_id[:8], exc
+                    )
+                    continue
 
                 # Handle state events from the plugin
                 if data.get("type") == "event":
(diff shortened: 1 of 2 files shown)


dsarno:
@coderabbitai review

---

Issue #680, opened 2026-07-11
Dock: 'Run this manually' fallback panel appears below the fold — no auto-scroll


dsarno:
UX nit from the 2026-07-11 live verification (Windows 11, Godot 4.6.2, #663 Settings-tab walkthrough).

In **Clients & Tools → Clients**, clicking Configure on a client whose CLI can't be found (e.g. 'Kimi Code (not detected)') correctly fails and expands the row's 'Run this manually:' panel (`mcp_dock.gd::_show_manual_command_for`). But when the row is low in the list, the panel renders below the visible scroll area and nothing scrolls it into view — the user sees the Configure click 'do nothing' unless they happen to scroll down.

Fix sketch: after making `manual_panel` visible, call `ensure_control_visible(manual_panel)` on the enclosing ScrollContainer (deferred one frame so layout has settled).

Issue #679, opened 2026-07-11
Windows: foreign server bound to 0.0.0.0 coexists with our 127.0.0.1 bind — foreign-port conflict undetected (and foreign-smoke repro nuance)


dsarno:
Found during the 2026-07-11 live verification (Windows 11, Godot 4.6.2).

**Observation:** with a foreign occupant bound to the wildcard address (`python -m http.server 18130` → `0.0.0.0:18130`), the plugin's spawned server still successfully bound `127.0.0.1:18130` — Windows permits a specific-address bind alongside another process's wildcard bind (no `SO_EXCLUSIVEADDRUSE`). Result: no conflict was diagnosed, startup proceeded normally, and both listeners coexisted (netstat showed `0.0.0.0:18130` = squatter and `127.0.0.1:18130` = godot-ai simultaneously).

**Consequences:**
1. Mostly harmless — the server works, loopback clients reach godot-ai because the specific bind wins routing for 127.0.0.1. But an agent probing the port *before* the spawn (or an external client hitting a LAN address) can reach the foreign server instead, and the pre-spawn probe path could misclassify.
2. **Repro/CI nuance:** `ci-stale-server-smoke --mode foreign` (and any manual foreign-occupant repro) only reproduces the conflict on Windows when the squatter binds `127.0.0.1` explicitly. Worth (a) a note in the smoke script/docs, and (b) checking whether the smoke's planted listener binds loopback on Windows — if it binds wildcard, the Windows row may be exercising the coexistence path, not the conflict path.

**Possible hardening:** bind with `SO_EXCLUSIVEADDRUSE` on Windows so a wildcard squatter produces an honest bind failure → the existing #659 diagnosis paths take over; or have the pre-spawn in-use check also try a wildcard-bind test.

dsarno:
Partial resolution in #682: confirmed the smoke's planted listeners already bind `127.0.0.1` explicitly (so the Windows CI rows do exercise the conflict path, answering point 2b), and that being load-bearing is now documented in the smoke's module docstring and at both bind sites. Still open here: the `SO_EXCLUSIVEADDRUSE` bind hardening (or a wildcard-bind pre-spawn probe) so a wildcard squatter produces an honest bind failure on Windows.

---
_Generated by [Claude Code](https://claude.ai/code)_

dsarno:
[...]

Issue #678, opened 2026-07-11
Editor semi-freezes during plugin init/reload when the server port is contended


dsarno:
Observed twice during the 2026-07-11 live verification of the #655–#666 merge train (Windows 11, Godot 4.6.2, dev venv, ports 18130/19630):

1. **Editor open with a foreign occupant on the HTTP port** (python http.server bound to 127.0.0.1:18130): the editor semi-froze during startup ('Project initialization / Initializing plugins…' hung around 80%), recovering after a noticeable delay, then correctly showed the INCOMPATIBLE dock panel.
2. **Dock 'Reload Plugin' click** right after a WS-port squatter was killed: a long semi-freeze before the reload completed. It did recover on its own — session came back `ready` and the server respawned cleanly.

Both hangs coincide with plugin init/teardown while port state is unusual, which points at synchronous work on the main thread in the startup/reload path — plausibly the pre-spawn port probe / live-server status probe (`_probe_live_server_status_for_port`) or the spawn-watch wait. Same family as #238/#239 (sync CLI probes blocking the main thread), which were fixed by moving the work off-thread.

Repro sketch:
- `python -m http.server <http_port> --bind 127.0.0.1`, then open the editor → freeze during plugin init.
- Or: squat the WS port, open editor, kill squatter, click Reload Plugin → freeze during reload.

Expected: probes/spawn waits happen off the main thread (or with tight timeouts + deferred continuation) so the editor stays responsive; the dock panel can arrive a beat later.

Severity: annoying but self-recovering; no data loss observed.

dsarno:
Fixed by #682 — `start_server` and `recover_strong_port_occupant` are coroutines in production: the port scrapes, per-PID brand shells, HTTP status probes, and kill-drain waits run on a dedicated worker thread while the main thread keeps pumping frames, so both repro scenarios (editor open with a foreign HTTP-port occupant; dock Reload right after a squatter was killed) no longer freeze the editor. Guard rails: generation-based cancellation on teardown and the one-shot restart paths, a re-entrancy guard, main-thread-only EditorSettings access, and the recovery await-chain through to the dock. `stop_server` stays synchronous (bounded 2s) since `_exit_tree` teardown must complete. Verified via both `ci-stale-server-smoke` modes with the async path active.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #672, opened 2026-07-11
Claude Code web session hook pins Godot 4.6.2 while the repo targets 4.7.0


dsarno:
## Summary

`.claude/hooks/session-start.sh` pins `GODOT_VERSION="4.6.2"` for Claude Code on the web sessions, but the repo has moved to 4.7: `test_project/project.godot` declares `config/features=PackedStringArray("4.7")` and all three OS rows in `ci.yml` pin `godot-version: "4.7.0"`. Remote smoke/test sessions therefore run the editor gauntlet on an engine one minor behind what CI certifies, and 4.7-specific behavior (e.g. the 4.7 Vulkan game-subprocess handling from #586) can't be exercised in those sessions.

This surfaced during the queue-sweep integration smoke (report: `QUEUE_SWEEP_SMOKE_REPORT.md` on `integration/queue-sweep-smoke`, §0), where the whole run had to be caveated as 4.6.2-only.

## Details

- `session-start.sh:27` — `GODOT_VERSION="4.6.2"`; the download URL template appends `-stable`, so the bump is `GODOT_VERSION="4.7"` (verify the release asset name `Godot_v4.7-stable_linux.x86_64.zip` matches).
- Cached binary at `~/.cache/godot-ai/godot` is reused across sessions, so existing environments keep serving 4.6.2 even after the hook is bumped — the hook should invalidate the cache on version change (e.g. cache path keyed by version, or a version check against `"$GODOT_BIN" --version`).
- Environment network policy: in at least some Claude Code web environments, `github.com/godotengine/godot` release downloads are blocked by the session egress proxy (403, out-of-scope repo). The hook's `curl -fsSL` would fail on a cold container there, leaving no Godot at all. Worth either documenting the required network-policy allowance or adding a clear failure message pointing at it.

## Suggested fix

1. Bump the pin to 4.7 and key the cache by version so stale 4.6.2 binaries are replaced.
2. Fail with an actionable message when the download is blocked by the environment's network policy.

dsarno:
Fixed by #682 — hook pinned to Godot 4.7 with a version-keyed cache (stale binaries pruned), an actionable failure message when the download is blocked by the environment network policy, and a bonus fix: the editor launch now sets `GODOT_AI_ALLOW_HEADLESS=1`, without which the plugin disabled itself and the promised `:8000` server never spawned in web sessions.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #682, opened 2026-07-12
Issue sweep: CI hardening, mandatory update verification, typed Array[T] writes, non-blocking startup probes

dsarno:
Sweeps eight open issues, one commit each, ordered smallest to largest.

## CI / tooling

- **#668 — `ci-godot-tests` aborts instead of retrying.** The session-probe pipeline (`curl | grep | awk`) killed the whole script under `set -euo pipefail` on exactly the conditions the retry loop targets (server down → curl exit 7; header missing → grep exit 1). `|| true` neutralizes the pipeline status so the empty-`SESSION_ID` retry branch is reachable; the two other in-loop curl calls are hardened the same way.
- **#671 — `ci-stale-server-smoke` back-to-back cross-contamination.** Both modes now run a bounded pre-plant drain: wait (never kill) until the HTTP and WS ports are genuinely free before planting the sim/listener, so a previous run's server in its orphan-reap window can't get adopted in place of the planted occupant. The probe **binds** rather than connects — GitHub's Windows runners silently drop SYNs to closed loopback ports, so connect-based probing reads every free port as occupied there (found via this PR's first two CI rounds); binding is also the exact operation the plant performs, mirroring `preflight_check_port`. Only `EADDRINUSE` counts as occupied; on drain timeout the harness fails naming the occupant.
- **#679 (docs half) — Windows wildcard-bind coexistence.** Confirmed the planted listeners already bind `127.0.0.1` explicitly — the Windows CI rows do exercise the conflict path. That being load-bearing is now documented in the module docstring and at both bind sites (a `0.0.0.0` squatter coexists with the plugin's loopback bind on Windows and never conflicts). The `SO_EXCLUSIVEADDRUSE` hardening idea remains open on the issue.
- **#672 — web-session hook pinned Godot 4.6.2.** Bumped to 4.7, cache keyed by version (`godot-4.7`) so warm containers replace stale engines on a bump, superseded cache entries pruned, and a blocked download now fails with a message pointing at the environment network policy (verified live in this environment, where the download 403s). Plus an adjacent bug found while testing: the hook launched the editor without `GODOT_AI_ALLOW_HEADLESS=1`, so the plugin disabled itself and the promised `:8000` server never spawned; fixed and verified live.

## Docs / contract

- **#669 — adoption-teardown contract.** Stated in AGENTS.md (server lifecycle section), the `--owner-pid` help text, and `ci-start-server`'s header: an editor that adopts an externally started server owns its end-of-life and stops it on clean editor exit; a long-lived `--reload` dev server should run on a non-default port.

## Plugin

- **#599 — self-update verification now mandatory.** The verify-if-present bypass is removed: a release without a `godot-ai-plugin.zip.sha256` sidecar refuses to install and drops the staged ZIP, per the maintainer's analysis that no population breaks (self-update only verifies the *next* download, and every release since #523 ships the sidecar). The URL-scoping follow-up had already shipped, so this completes the issue. New test covers the refusal (and that the retry button stays enabled); the existing trusted/untrusted/match tests still pass.
- **#612 stage 1 — typed `Array[T]` values silently dropped.** A JSON list into a typed `Array[T]` `@export` slot used to pass the element-type-agnostic `TYPE_ARRAY` passthrough as an untyped Array, get rejected wholesale by Godot's typed setter, and leave the slot empty with success reported. New `NodeHandler._coerce_typed_array` detects the typed slot from the property's current value (the getter returns the empty typed container — no `hint_string` parsing), coerces each element through the existing `_coerce_value`/`_check_coerced` pair, and bulk-moves via `Array.assign()` with a post-assign size check. Wrong/mixed elements clean-error naming the element index; nothing partially writes. Both `set_property` and `_apply_resource_properties` dispatch ahead of the generic passthrough. Object elements (`Array[Texture2D]`, …) are stage 2 and refuse loudly instead of dropping silently — except an empty list, which clears the slot (review finding, regression-tested); typed `Dictionary[K,V]` remains a follow-up per the maintainer's PR-shape decision on the issue.
- **#678 — editor semi-freeze during plugin init/reload on contended ports.** `start_server`'s contended branch chained port scrapes (lsof/netstat/PowerShell), per-PID cmdline brand shells, an 800ms HTTP status-probe busy-wait, and up to 3s of `_wait_for_port_free` — synchronously on the main thread (same family as #238/#239). `start_server` and `recover_strong_port_occupant` are now coroutines: blocking sections run on `WorkerThreadPool` via a `_run_blocking` primitive while the main thread keeps pumping frames. Safety rails: `defer_blocking_work` defaults off so the synchronous characterization tests keep call-then-assert behavior (the real plugin enables it in `_enter_tree`, which test fixtures never reach); an async-generation counter bumped by `stop_server` cancels in-flight startup on teardown; a re-entrancy guard blocks double-starts; EditorSettings reads/writes stay main-thread-only (the managed record is injected into the proof helper via a new `record_override` param); the trace counters bumped off-thread are mutex-guarded. `stop_server` stays synchronous — `_exit_tree` teardown must complete, and its wait is bounded at 2s.

## Verification

- Full GDScript suite live against the editor: **1709/1728 passed, 0 failed** (includes 15 typed-array tests, 2 self-update tests, and 4 async-startup tests; lifecycle characterization suites all green through the synchronous inline path).
- Both `ci-stale-server-smoke` modes pass back-to-back locally **with the async startup path active** — the two #678 repro scenarios (foreign-occupant INCOMPATIBLE diagnosis and stale kill+respawn) work end-to-end, and the #671 drain absorbs the between-runs reap window.
- `pytest` green; `ruff check` clean; hook exercised live (blocked-download error path, version-keyed reuse, pruning, headless editor + server bring-up).

Not addressed here: #651 stage 2 / #670 (gated on post-release telemetry), #612 stages 2–3 (per maintainer's staging), #518 (needs fleet data on 2.9.2+), #574 (needs a Windows machine), #498 (deferred by design discussion), #680 (open PR #681).

https://claude.ai/code/session_01AZkpsWbMrFMHxfVm42T2ks

## Summary by CodeRabbit

- **New Features**
  - Added support for setting typed arrays with automatic conversion for values such as numbers, colors, vectors, and nested arrays.
  - Improved asynchronous startup and recovery to prevent stale server processes from reappearing during shutdown or restart.

- **Bug Fixes**
  - Updates now require checksum verification before installation.
  - Improved handling of unavailable servers, port conflicts, and startup retries.
  - Godot sessions now use version-specific caching and Godot 4.7.

- **Documentation**
  - Clarified server ownership, adoption, and lifecycle behavior during editor restarts.

### DIFF
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -24,9 +24,14 @@ fi
 
 cd "$CLAUDE_PROJECT_DIR"
 
-GODOT_VERSION="4.6.2"
+# Keep in step with ci.yml (godot-version: "4.7.0" — the chickensoft action
+# wants 3-part semver; the Godot release tag/asset uses the 2-part form) and
+# test_project/project.godot (config/features "4.7"). See #672.
+GODOT_VERSION="4.7"
 GODOT_CACHE="$HOME/.cache/godot-ai"
-GODOT_BIN="$GODOT_CACHE/godot"
+# Version-keyed binary path: bumping GODOT_VERSION invalidates the cache, so
+# warm containers can't keep serving a stale engine after a version bump.
+GODOT_BIN="$GODOT_CACHE/godot-${GODOT_VERSION}"
 GODOT_ZIP="Godot_v${GODOT_VERSION}-stable_linux.x86_64.zip"
 GODOT_EXE="Godot_v${GODOT_VERSION}-stable_linux.x86_64"
 GODOT_URL="https://github.com/godotengine/godot/releases/download/${GODOT_VERSION}-stable/${GODOT_ZIP}"
@@ -37,7 +42,17 @@ script/setup-dev
 if [ ! -x "$GODOT_BIN" ]; then
   echo "[godot-ai session-start] downloading Godot ${GODOT_VERSION}..."
   mkdir -p "$GODOT_CACHE"
-  curl -fsSL -o "$GODOT_CACHE/godot.zip" "$GODOT_URL"
+  if ! curl -fsSL -o "$GODOT_CACHE/godot.zip" "$GODOT_URL"; then
+    cat >&2 <<EOF
+[godot-ai session-start] ERROR: could not download Godot ${GODOT_VERSION} from
+  $GODOT_URL
+If this is a 403, the session's environment network policy is blocking
+github.com/godotengine release downloads (out-of-scope repo). Allow that
+domain/repo in the environment's network policy, or pre-seed the binary at
+$GODOT_BIN. Godot-dependent MCP tools will be unavailable this session.
+EOF
+    exit 1
+  fi
   unzip -q -o "$GODOT_CACHE/godot.zip" -d "$GODOT_CACHE"
   mv "$GODOT_CACHE/${GODOT_EXE}" "$GODOT_BIN"
   chmod +x "$GODOT_BIN"
@@ -45,6 +60,11 @@ if [ ! -x "$GODOT_BIN" ]; then
 else
   echo "[godot-ai session-start] reusing cached Godot at $GODOT_BIN"
 fi
+
+# Drop superseded cache entries: the pre-#672 unversioned path and any
+# godot-<other-version> binaries left behind by earlier pins.
+find "$GODOT_CACHE" -maxdepth 1 -type f \( -name 'godot' -o -name 'godot-*' \) \
+  ! -name "godot-${GODOT_VERSION}" -delete 2>/dev/null || true
 
 mkdir -p "$HOME/.local/bin"
 ln -sf "$GODOT_BIN" "$HOME/.local/bin/godot"
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -84,7 +84,11 @@ if curl -sf -o /dev/null -X POST \
   echo "[godot-ai session-start] MCP server already running on :8000"
 else
   echo "[godot-ai session-start] launching headless Godot editor (plugin auto-spawns MCP server on :8000)..."
-  nohup "$GODOT_BIN" --headless --path test_project --editor \
+  # GODOT_AI_ALLOW_HEADLESS=1 is required: without it the plugin disables
+  # itself in --headless launches (plugin.gd::_mcp_disabled_for_headless)
+  # and the promised MCP server never spawns. A web session driving MCP
+  # tools is exactly the intentional-headless case AGENTS.md carves out.
+  nohup env GODOT_AI_ALLOW_HEADLESS=1 "$GODOT_BIN" --headless --path test_project --editor \
     > /tmp/godot-editor.log 2>&1 &
   disown
 fi
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -61,9 +61,10 @@ else
   echo "[godot-ai session-start] reusing cached Godot at $GODOT_BIN"
 fi
 
-# Drop superseded cache entries: the pre-#672 unversioned path and any
-# godot-<other-version> binaries left behind by earlier pins.
-find "$GODOT_CACHE" -maxdepth 1 -type f \( -name 'godot' -o -name 'godot-*' \) \
+# Drop superseded cache entries: the pre-#672 unversioned path, any
+# godot-<other-version> binaries left behind by earlier pins, and a
+# leftover godot.zip from an interrupted download.
+find "$GODOT_CACHE" -maxdepth 1 -type f \( -name 'godot' -o -name 'godot-*' -o -name 'godot.zip' \) \
   ! -name "godot-${GODOT_VERSION}" -delete 2>/dev/null || true
 
 mkdir -p "$HOME/.local/bin"
--- a/plugin/addons/godot_ai/handlers/node_handler.gd
+++ b/plugin/addons/godot_ai/handlers/node_handler.gd
@@ -922,7 +922,9 @@ static func _coerce_typed_array(value: Variant, slot_value: Array, prefix: Strin
 		)
 		return ErrorCodes.prefix_message(err, prefix)
 	var elem_type := slot_value.get_typed_builtin()
-	if elem_type == TYPE_OBJECT:
+	## An empty list has no elements to coerce, so it may clear the slot even
+	## when the element type is stage-2 territory (PR #682 review).
+	if elem_type == TYPE_OBJECT and not (value as Array).is_empty():
 		var obj_err := ErrorCodes.make(
 			ErrorCodes.WRONG_TYPE,
 			"Writing elements into a typed Array[%s] property is not supported yet (#612 stage 2 covers Object elements); assign the elements individually in the editor for now" % elem_label,
--- a/plugin/addons/godot_ai/handlers/node_handler.gd
+++ b/plugin/addons/godot_ai/handlers/node_handler.gd
@@ -939,11 +939,14 @@ static func _coerce_typed_array(value: Variant, slot_value: Array, prefix: Strin
 		if elem_err != null:
 			return elem_err
 		if coerced == null:
+			## Prefix with `elem_prefix` (which already folds in `prefix`),
+			## not `prefix` again — the latter double-stamped the property
+			## context (PR #682 review).
 			var null_err := ErrorCodes.make(
 				ErrorCodes.WRONG_TYPE,
-				"%s: cannot store null in Array[%s]" % [elem_prefix, elem_label],
+				"cannot store null in Array[%s]" % elem_label,
 			)
-			return ErrorCodes.prefix_message(null_err, prefix)
+			return ErrorCodes.prefix_message(null_err, elem_prefix)
 		staging.append(coerced)
 	var out := slot_value.duplicate()
 	out.clear()
--- a/plugin/addons/godot_ai/handlers/node_handler.gd
+++ b/plugin/addons/godot_ai/handlers/node_handler.gd
@@ -251,6 +251,16 @@ func set_property(params: Dictionary) -> Dictionary:
 				return apply_err
 		value = res
 		instantiated_resource = true
+	elif target_type == TYPE_ARRAY and old_value is Array and (old_value as Array).is_typed():
+		## Typed Array[T] slot (#612): the generic TYPE_ARRAY passthrough
+		## hands an untyped Array to Godot's typed setter, which rejects it
+		## wholesale and leaves the slot at its default — with success still
+		## reported. Route through the element-aware coercer instead; errors
+		## name the offending element index.
+		var typed_out: Variant = _coerce_typed_array(value, old_value)
+		if typed_out is Dictionary:
+			return typed_out
+		value = typed_out
 	else:
 		value = _coerce_value(value, target_type)
 		## Refuse any value that didn't land as the target compound Variant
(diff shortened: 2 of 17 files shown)


plugin/addons/godot_ai/utils/server_lifecycle.gd
```diff
@@ -435,7 +485,21 @@ func _set_plugin_spawned_env() -> void:
 ##   port in use, record matches + live ok   -> adopt port owner (heals PID)
 ##   port in use, record drifts              -> kill owner + respawn
 ##   port in use, no verified live match     -> block adoption + warn
+##
+## #678: this is a coroutine in production (`defer_blocking_work`) — the
```
dsarno, line 503: **Race: one-shot restart paths can silently no-op against an in-flight async startup.**

Only `stop_server` bumps `_async_generation`. Two production paths kill the port occupant and then call `start_server()` without cancelling an in-flight walk:

- `force_restart_server()` (line ~990, outside this diff) — dock "Restart Server" button
- the incompatible-recovery path ending in `start_server()` (line ~927, outside this diff)

Startup is now a multi-second suspended coroutine on a contended port, and the editor stays *responsive* during it (the point of #678) — so these dock buttons are clickable exactly while a walk is in flight, a window the old synchronous freeze made unreachable. If clicked mid-flight:

1. the recovery path kills the occupant, resets state, calls `start_server()` → no-op via this guard;
2. the suspended walk resumes with its generation still valid and continues against pre-kill data (e.g. a live-status snapshot of the now-dead occupant), potentially latching INCOMPATIBLE or adopting a dead server;
3. net: the user's explicit restart is silently dropped.

The 1s health-watch timer is fine (it re-fires), but these two paths are one-shot.

Suggestion: bump `_async_generation` in `reset_for_force_restart()` (and before the kill in the `:927` path) to cancel the stale walk, and make sure the fresh `start_server()` isn't swallowed by `_start_in_flight` (await the stale walk's unwind, or re-arm). A test mirroring `test_stop_server_invalidates_async_generation` for the force-restart path would lock it in.

Related, minor: both paths now return before the respawned walk completes (fire-and-forget coroutine call), so dock code reading state immediately after may see a pre-spawn state for a few frames.

---

plugin/addons/godot_ai/utils/server_lifecycle.gd
```diff
@@ -72,11 +72,61 @@ var _startup_path: String = McpStartupPathScript.UNSET
...
+## Falls back to inline execution when `defer_blocking_work` is off, or
+## when no SceneTree is available to pump frames against.
+##
+## Uses a dedicated Thread (the dock's #238/#239 worker pattern) rather
+## than WorkerThreadPool: `wait_to_finish()` hands the return value back
+## without a shared mutable container, and this plugin has already seen
+## WorkerThreadPool tasks SIGABRT under concurrency (see the notes in
+## script_handler.gd / filesystem_handler.gd). `wait_to_finish` after
+## `is_alive()` goes false joins an already-dead thread, so it never
+## blocks the main thread.
```
dsarno, line 111: **Worker-thread lifetime (minor):** the task lambdas capture `_host` (the EditorPlugin). If the plugin object is freed while a task is still running a bounded shell probe (~800ms–3s), the worker thread hits "previously freed instance" errors — the generation check protects state *after* resume, but not calls *inside* the task. Non-fatal (errors, not crashes) and the window is small, but an `is_instance_valid(_host)` guard at the top of the deferred closures would quiet it.

---

plugin/addons/godot_ai/handlers/node_handler.gd
```diff
@@ -887,6 +897,86 @@ static func _coerce_value(value: Variant, target_type: int) -> Variant:
...
+		var coerced: Variant = _coerce_value(in_list[i], elem_type)
+		var elem_err := _check_coerced(coerced, elem_type, elem_prefix)
+		if elem_err != null:
+			return elem_err
+		if coerced == null:
+			## Prefix with `elem_prefix` (which already folds in `prefix`),
+			## not `prefix` again — the latter double-stamped the property
+			## context (PR #682 review).
+			var null_err := ErrorCodes.make(
+				ErrorCodes.WRONG_TYPE,
```
dsarno, line 946: **Nit: double prefix on the null-element arm.** `null_err`'s message is built with `elem_prefix` — which already embeds the outer `prefix` on the resource path — and is then wrapped in `prefix_message(null_err, prefix)` again, yielding e.g. `Property 'items': Property 'items' element 1: cannot store null in Array[int]`. The `_check_coerced` arm above returns its error raw, which is the right shape; returning `null_err` unwrapped here matches it. (Node path is unaffected — `prefix` is empty there.)

---

plugin/addons/godot_ai/handlers/node_handler.gd
```diff
@@ -251,6 +251,16 @@ func set_property(params: Dictionary) -> Dictionary:
 				return apply_err
 		value = res
 		instantiated_resource = true
+	elif target_type == TYPE_ARRAY and old_value is Array and (old_value as Array).is_typed():
```
dsarno, line 254: **Note (no action needed for stage 1):** this dispatch relies on the getter returning the live typed container. A scripted property whose custom getter returns an *untyped* copy still falls through to the old generic passthrough and its silent-drop behavior. Probably worth a line on #612 so it's a known residual rather than a surprise.

---

script/ci-stale-server-smoke
```diff
@@ -188,6 +203,80 @@ class SmokeError(RuntimeError):
...
+
+    Mirrors `preflight_check_port` in src/godot_ai/__init__.py: SO_REUSEADDR
+    on POSIX so a just-freed port's TIME_WAIT connections don't read as an
+    occupant (the sim's HTTPServer sets allow_reuse_address too), skipped on
+    Windows where SO_REUSEADDR means "hijack the active listener". Only
+    EADDRINUSE counts as occupied — any other bind failure (EACCES, winnat
+    exclusion ranges) is not a drainable occupant, and the plant will surface
+    its own honest failure for those."""
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        if os.name != "nt":
```
dsarno, line 227: **Watch item, not a blocker:** skipping `SO_REUSEADDR` on Windows is right for the hijack reason given, but it also means server-side TIME_WAIT sockets on the port (connections the previous run's server closed first) bind-fail with `EADDRINUSE` and read as an occupant. Windows' TIME_WAIT default (~4 min) exceeds the 150s drain timeout, so a spurious `SmokeError` (occupant "unknown") is possible in the worst case. CI is green now and the failure mode is loud and named — just flagging this as the first place to look if the Windows smoke ever flakes on "port(s) still occupied".

---

.claude/hooks/session-start.sh
```diff
@@ -46,6 +61,11 @@ else
   echo "[godot-ai session-start] reusing cached Godot at $GODOT_BIN"
 fi
 
+# Drop superseded cache entries: the pre-#672 unversioned path and any
+# godot-<other-version> binaries left behind by earlier pins.
+find "$GODOT_CACHE" -maxdepth 1 -type f \( -name 'godot' -o -name 'godot-*' \) \
+  ! -name "godot-${GODOT_VERSION}" -delete 2>/dev/null || true
+
```
dsarno, line 69: **Nit:** `godot.zip` matches neither prune pattern (`godot` exact, `godot-*`), so the last downloaded archive lingers in the cache forever. Pre-existing behavior (it's overwritten on each download), but since this change adds the prune anyway, `-o -name 'godot.zip'` would finish the job.

---

plugin/addons/godot_ai/plugin.gd
```diff
@@ -206,6 +209,12 @@ func _enter_tree() -> void:
...
 	## and reset to noisy on every editor restart (#626).
 	_log_buffer.enabled = McpSettings.mcp_logging_enabled()
+	## #678: in the real editor, run the startup path's blocking probes and
+	## kill-drain waits off the main thread so a contended port can't freeze
+	## plugin init/reload. Set here (not _init) so test fixtures — which
+	## extend this plugin but never enter the tree — keep the synchronous
+	## default and can call-then-assert.
+	_lifecycle.defer_blocking_work = true
 	_start_server()
 	_startup_trace_phase("server_start")
```
dsarno, line 219: **Nit:** with `start_server` now a coroutine, `_startup_trace_phase("server_start")` stamps when the walk first *suspends*, not when it completes — the contended-port cost largely disappears from the startup trace. Worth remembering if trace numbers are ever compared across the #678 boundary; a phase mark at walk completion would keep the old meaning.

---

dsarno (commented):
Thorough pass over all eight commits, cross-checked against the base code (`_fail_verification`, `_read_managed_server_record` sentinel, `ci.yml` version pin, `Array.assign` commit-on-success semantics) and current CI (all completed checks green on 9fe1622; Codecov reports all modified coverable lines covered).

**Summary of findings** (details in the inline comments):

- **Main finding** — the #678 re-entrancy guard + generation scheme misses the two one-shot kill-and-restart paths (`force_restart_server`, incompatible recovery), which neither bump `_async_generation` nor get past `_start_in_flight`; a dock restart clicked during a contended startup can be silently dropped while the stale walk resumes against post-kill reality. Small, contained fix (generation bump in `reset_for_force_restart()` + not swallowing the follow-up start); would be good before merge since #678 makes the race newly reachable. *(The affected call sites — server_lifecycle.gd ~927 and ~990 — are outside the diff, so the comment is anchored on the new guard.)*
- One-line cosmetic fix: double-prefixed null-element error in `_coerce_typed_array` (resource path only).
- Minor/watch items: worker-thread `_host` lifetime during teardown; Windows TIME_WAIT vs. the bind-probe drain; `godot.zip` never pruned; startup-trace phase timing semantics; scripted-getter untyped-copy residual for #612.

**What looks solid:** the typed-array design (typing detected from the getter's container, per-element coercion, `assign()` size backstop — correct given `Array.assign` only commits on full success — explicit null rejection, empty-list clear with undo coverage); mandatory verification for #599 with a thorough refusal test; the `|| true` scoping under `set -euo pipefail` for #668; the bind-vs-connect probe rationale and wait-don't-kill drain for #671; and the #669/#679 docs placed where people will actually trip over the behavior. Test coverage across all three plugin changes is excellent.

Verdict: approve-with-comments — one requested change (the restart race), everything else nits or watch items.

---

plugin/addons/godot_ai/utils/server_lifecycle.gd
dsarno: Fixed in aef2a44. `_invalidate_async_startup()` (generation bump + guard release) is now called by `stop_server`, `reset_for_force_restart`, and the incompatible-recovery path once it's committed to the kill; the `start_server` wrapper releases the guard only when its own generation is still current, so the cancelled walk's unwind can't clobber a newer walk's guard. Test mirroring `test_stop_server_invalidates_async_generation` added for the force-restart path (also asserts the guard release, since without it the follow-up start is the part that gets swallowed).

On the fire-and-forget note: agreed it's benign for the dock today (it re-reads state on its refresh cadence), so I left both paths returning before the respawned walk completes rather than making their bool returns async.

---

plugin/addons/godot_ai/handlers/node_handler.gd
dsarno: This one was fixed in 9fe1622 (pushed just before your review snapshot): the null arm now builds the bare message and stamps once via `prefix_message(null_err, elem_prefix)`, matching the `_check_coerced` arm's shape — `Property 'items' element 1: cannot store null in Array[int]`. Regression test asserts the property context appears exactly once.

---

plugin/addons/godot_ai/utils/server_lifecycle.gd
dsarno: Follow-up in 680bfd0: CodeRabbit's delayed pass caught that `recover_incompatible_server()` returned true while its trailing walk was still suspended, so `_resume_connection_after_recovery` (which gates on SPAWNING/READY) ran against STOPPED and left the connection blocked. The await now propagates manager → plugin wrapper → dock dispatch → both dock callers, so recovery reports success only after the walk completed. That supersedes the fire-and-forget stance I stated above for the *recovery* path — `force_restart_server` alone stays fire-and-forget (nothing state-gated downstream). Also moved the startup-trace 'done' stamp to walk completion and mutex-guarded the netsh counter write in `_startup_trace_finish`.

---

Issue #687, opened 2026-07-12
[S3] Self-update integrity: self-signed checksum + unverified install writes


dsarno:
Surfaced by the codebase audit. The self-update integrity chain has one design gap and two mechanical write-verification gaps that can install a corrupt tree over live plugin code.

### 1. Checksum is self-signed (design) — `update_manager.gd:~404`
`download_url` and `checksum_url` are both read from the same GitHub Releases API response (`parse_releases_response:274-281`) and fetched over the same channel. Anyone who can modify the release's assets (leaked repo token, compromised release workflow) regenerates the `.sha256` sidecar to match a tampered zip. The docstring (405-409) overclaims: "a tampered asset … can't be installed over live plugin code" — true only for single-object substitution / in-transit corruption, **not** a tampered release.

**Fix (design):** embed a static ed25519/minisign public key in the plugin, sign the sidecar in `release.yml` (ideally with a key outside the release workflow's default token scope), and verify the signature in `_verify_then_install` before trusting the digest. **Immediate mechanical tightening:** rewrite the docstring to state the actual guarantee (integrity vs. corruption + object-substitution, not vs. a compromised release).

### 2. Install write verification checks the wrong signal — `update_reload_runner.gd:~377`
`_install_zip_file` does `f.store_buffer(content)` (ignoring the bool return), then checks `f.get_error()`. In Godot, `store_buffer` reports a short write **only via its return value** — it doesn't set the last-error state `get_error()` reads — and the write is stdio-buffered, so disk-full surfaces at flush/close. `f.close()` is unchecked; there is no post-write size verification before the temp file is renamed over the live target.

**Failure:** disk-full during self-update → buffered write "succeeds", `get_error()` returns OK, a truncated file is renamed over e.g. `plugin.gd`, backups are deleted by `_finalize_install_success`, and the plugin re-enables against a corrupt tree → parse-error cascade with no rollback left.

**Fix:** capture `store_buffer`'s return; after `close()`, verify `stored and write_error == OK and FileAccess.get_file_as_bytes(temp_path).size() == content.size()`, else remove the temp and return `{}` so the existing rollback path engages.

### 3. `McpAtomicWrite` rename path skips the size check its copy path enforces — `_atomic_write.gd:~88`
[...]

PR #702, opened 2026-07-13
Verify self-update/atomic-write bytes actually landed before trusting them

dsarno:
## Summary

Fixes findings 2 and 3 of #687 — two mechanical write-verification gaps in the self-update/atomic-write paths that could install or persist a truncated file with no rollback path left:

- **`update_reload_runner.gd::_install_zip_file`** ignored `store_buffer`'s bool return (confirmed via a scratch script against this engine's Godot 4.7 — `store_buffer` returns `bool`, unlike `store_string`, and does **not** set the `get_error()` state on a short write) and never verified the temp file's actual on-disk size before renaming it over a live plugin script. A disk-full mid-update could rename a truncated file over e.g. `plugin.gd`, delete backups via `_finalize_install_success`, and leave the plugin to re-enable against a corrupt tree. Now checks `stored`, `write_error`, and the real on-disk byte count before trusting the temp file.
- **`_atomic_write.gd::write()`** staged its temp file via `store_string`/`flush`/`close` with no size check, then committed by rename — while the copy-fallback path immediately below already enforced exactly this via `_written_size_matches` (the #297 data-loss guard the module header cites). Both commit paths now share the same integrity gate.

## Not included — needs a maintainer decision

Finding 1 of #687 (self-signed checksum: `download_url` and `checksum_url` both come from the same GitHub Releases API response, so a compromised release workflow/token can regenerate a matching `.sha256`) is a release-pipeline design change — `docs/audit-tier2-plan.md` flags "#687 release signing (ed25519/minisign + key management)" as a maintainer infra decision. Raised as a comment on #687; not attempted here.

## ⚠️ Verification gap — needs maintainer sign-off before merge

This PR touches `update_reload_runner.gd` and `_atomic_write.gd`, which per `AGENTS.md` requires running `script/local-self-update-smoke` — an **interactive** harness where "the operator still performs the one step that matters: click Update in the dock." This session is headless with no GUI, so I have no way to drive that click. I did not skip this by choice; I'm flagging it explicitly rather than claiming a check I couldn't run. Everything else in the gauntlet is green (see below). Please run `script/local-self-update-smoke` before merging.

## Test plan

- [x] `ruff check src/ tests/`
- [x] `pytest -q` — 1307 passed, 4 skipped
- [x] `bash script/ci-check-gdscript` — all GDScript OK
- [x] Live editor test via `script/ci-start-server` + fresh headless editor + `script/ci-godot-tests` — 0 failed (1724/1742 passed); `update_reload_runner` suite 16/16, `clients` suite 115/115
- [x] New GDScript tests in `test_clients.gd`: `test_written_size_matches_true_for_exact_content`, `test_written_size_matches_false_for_truncated_content`, `test_written_size_matches_false_for_missing_file` — pin `_written_size_matches`'s exact contract directly, since forcing a genuine disk-full/quota failure through the real `write()`/`_install_zip_file` call path isn't portable or deterministic in this CI environment (same limitation as the existing, similarly-untested `update_reload_runner.gd:378` pattern this PR extends)
- [ ] `script/local-self-update-smoke` — **not run**, see above
- [x] `git checkout -- test_project/project.godot` (no editor churn was staged)

https://claude.ai/code/session_01MYSH45sEiTHKWx8rupx4vU

---
_Generated by [Claude Code](https://claude.ai/code/session_01MYSH45sEiTHKWx8rupx4vU)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented incomplete or truncated temporary files from being committed during atomic writes.
  - Added validation to ensure extracted update files are fully written before installation.
  - Improved write failure reporting with details about stored and expected file sizes.

- **Tests**
  - Added coverage for complete, truncated, and missing temporary files.

### DIFF
--- a/plugin/addons/godot_ai/clients/_atomic_write.gd
+++ b/plugin/addons/godot_ai/clients/_atomic_write.gd
@@ -66,6 +66,16 @@ static func write(path: String, content: String) -> bool:
 	# which preserves it.
 	_apply_mode(tmp_path, target_mode)
 
+	# Verify the staged temp actually landed on disk before either commit path
+	# (rename or the copy-fallback below) trusts it. The copy-fallback path
+	# already guards this (_written_size_matches at line ~95 below) — under
+	# disk-full/quota a truncated rename would otherwise "succeed" and the
+	# caller would treat a corrupt config as correctly written while the
+	# `.backup` (not yet taken) holds nothing to recover from.
+	if not _written_size_matches(tmp_path, content):
+		DirAccess.remove_absolute(tmp_path)
+		return false
+
 	# Best-effort: snapshot the prior file before we touch the target so we
 	# can restore on a failed swap. The backup is also kept on success as a
 	# one-shot rollback aid for the user — give it the same (preserved) mode
--- a/plugin/addons/godot_ai/update_reload_runner.gd
+++ b/plugin/addons/godot_ai/update_reload_runner.gd
@@ -374,13 +374,25 @@ func _install_zip_file(
 			FileAccess.get_open_error(),
 		])
 		return {}
-	f.store_buffer(content)
+	## store_buffer only reports a short write via its bool return — it does
+	## NOT set the last-error state get_error() reads — and the write is
+	## stdio-buffered, so disk-full can surface only at close(). Check both
+	## signals, then verify the bytes actually landed on disk before trusting
+	## this temp file: a truncated file renamed over a live plugin script (or
+	## project.godot) parse-error-cascades the next enable with no rollback.
+	var stored := f.store_buffer(content)
 	var write_error := f.get_error()
 	f.close()
-	if write_error != OK:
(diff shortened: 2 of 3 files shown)


dsarno:
Closing as a duplicate of #695, which was opened ~1h earlier from the Phase-1 worker's patch handoff and covers the same two #687 findings (the `_install_zip_file` store_buffer/size verification and the `_atomic_write` rename-path `_written_size_matches` gate) plus the `_verify_then_install` docstring correction in one PR. I cross-checked the two implementations — they land the same integrity guarantees; #695 stays canonical because it's the one the maintainer is already tracking for the required manual `script/local-self-update-smoke` sign-off, and one smoke run should gate one PR, not two.

(Context: this duplication came from multiple worker sessions being spawned against the same brief during an environment reconfiguration — not a fault of this PR's content, which was sound.)

---

PR #743, opened 2026-07-14
McpScenePath.from_node: single upward walk instead of two O(depth) passes

eth0net:
## Motivation

`McpScenePath.from_node` paid **two** O(depth) walks per node: `is_ancestor_of()` (node → root) to guard against `get_path_to()` emitting a `"../"`-style path for a non-descendant, then `get_path_to()` (root → node). On hot, unbounded reads like `find_nodes` / `get_children` over a deep tree that's O(n·depth) with a 2× constant.

## Change

Walk up from the node to `scene_root` once, collecting names: reaching `scene_root` proves ancestry **and** yields the relative segments in a single pass; walking off the tree top means it isn't a descendant (return `""`). Output and the empty-string guard are unchanged.

## Tests

Behavior-preserving — validated by the existing exhaustive `test_scene_path.gd` (root / child / nested / null node / null root / orphan / foreign tree / ancestor-of-root). Rebased onto v3.0.2 (clean — `from_node` was untouched upstream; the file's other functions gained the new `resolve("/")` alias); parse-clean. Full Godot `test_run` runs in CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scene path generation for nodes within a scene hierarchy.
  * Invalid or unrelated nodes now correctly return an empty path.
  * Preserved correct handling for missing inputs and the scene root itself.

### DIFF
--- a/plugin/addons/godot_ai/utils/scene_path.gd
+++ b/plugin/addons/godot_ai/utils/scene_path.gd
@@ -9,18 +9,29 @@ const ErrorCodes := preload("res://addons/godot_ai/utils/error_codes.gd")
 
 
 ## Return a clean path relative to the scene root (e.g. /Main/Camera3D).
-## Returns "" when `node` is not the scene root or a descendant of it —
-## without the ancestry guard, get_path_to() returns an empty NodePath that
-## concatenates into a plausible-looking but invalid "/Main/".
+## Returns "" when `node` is not the scene root or a descendant of it.
+##
+## Single upward walk from `node` to `scene_root`, collecting names as we go.
+## The pre-PR version paid two O(depth) walks per node — `is_ancestor_of()`
+## (node -> root) to guard against get_path_to() emitting a "../"-style path
+## for non-descendants, then `get_path_to()` (root -> node) again. Walking up
+## once both validates ancestry (did we reach scene_root before the tree top?)
+## and builds the relative segments, halving the per-node cost on hot reads
+## like find_nodes / get_children with no change to the output or the guard.
 static func from_node(node: Node, scene_root: Node) -> String:
 	if scene_root == null or node == null:
 		return ""
 	if node == scene_root:
 		return "/" + scene_root.name
-	if not scene_root.is_ancestor_of(node):
-		return ""
-	var relative := scene_root.get_path_to(node)
-	return "/" + scene_root.name + "/" + str(relative)
+	var segments: Array[String] = []
+	var current := node
+	while current != null and current != scene_root:
+		segments.append(str(current.name))
+		current = current.get_parent()
+	if current != scene_root:
+		return ""  # walked off the top without hitting scene_root: not a descendant
+	segments.reverse()
+	return "/" + scene_root.name + "/" + "/".join(segments)
 
 
 ## Resolve a clean scene path like "/Main/Camera3D" to the actual node.


dsarno:
@eth0net These are awesome, thank you! Reviewing now.  

Can you drop into our discord so I can give you a shoutout for contributions? I added your name to the big 3.0 announcement post but didn't have your discord handle.  https://discord.com/invite/FDZ5fr2QkP

dsarno (changes requested):
Requesting changes — I benchmarked this, and as measured it's a slowdown, not an optimization. Correctness is fine (all eight `test_scene_path.gd` edge cases agree between the two implementations), so the PR is purely a perf trade, and it loses:

| chain depth | main (2 native walks) | this PR (GDScript loop) | ratio |
|---|---|---|---|
| 3 | 1.05 µs/call | 1.50 µs/call | **1.44× slower** |
| 10 | 1.53 µs/call | 3.42 µs/call | **2.24× slower** |
| 30 | 2.66 µs/call | 8.82 µs/call | **3.31× slower** |

Godot 4.6.3 headless, macOS ARM, 100k calls per implementation per depth; the three variants (main, this PR, and a third — see below) were first cross-checked to agree on root / child / deep / null-node / null-root / orphan / foreign-tree / ancestor-of-root.

The complexity argument — "two O(depth) walks → one" — counts the walks but not where they run. `is_ancestor_of()` and `get_path_to()` are each a *single* GDScript→C++ boundary crossing that loops natively; the replacement loop pays roughly three interpreted operations per ancestor level (`get_parent()`, `str(name)`, `append`) plus a `reverse()` and a `join()`, and the boundary/interpreter overhead dwarfs the native per-level cost. The gap *widens* with depth — the exact scenario the PR targets.

Two related notes so the idea doesn't resurface in another shape:

- I also benchmarked the obvious middle ground — drop `is_ancestor_of`, call `get_path_to` alone, and reject empty or `".."`-prefixed results. It's no faster than main (~1.06 / 1.54 / 2.77 µs — the ancestry pre-walk is essentially free next to the call overhead), **and** Godot's C++ prints `ERROR: Parameter "common_parent" is null` for orphan/foreign nodes, so the existing guard is also load-bearing for clean editor logs. Main's current implementation is the best of the three variants.
- With #742 merged, `from_node` only runs for the ≤`limit` windowed nodes per call, which shrinks the hot path this PR was targeting.

Happy to reconsider with a benchmark showing a win on a real editor build, but absent that this should close. (Minor and moot if so: the new doc comment narrates pre-PR history — "The pre-PR version paid two O(depth) walks…" — which is changelog content rather than a code invariant.)

<details>
<summary>Benchmark script (headless, <code>godot --headless --script bench_from_node.gd</code>)</summary>

```gdscript
extends SceneTree
## Microbenchmark: McpScenePath.from_node — old (is_ancestor_of + get_path_to)
## vs PR 743 (single GDScript upward walk) vs alt (get_path_to + ".." check).


static func from_node_old(node: Node, scene_root: Node) -> String:
	if scene_root == null or node == null:
		return ""
	if node == scene_root:
		return "/" + scene_root.name
	if not scene_root.is_ancestor_of(node):
		return ""
	var relative := scene_root.get_path_to(node)
	return "/" + scene_root.name + "/" + str(relative)


static func from_node_pr(node: Node, scene_root: Node) -> String:
	if scene_root == null or node == null:
		return ""
	if node == scene_root:
		return "/" + scene_root.name
	var segments: Array[String] = []
	var current := node
	while current != null and current != scene_root:
		segments.append(str(current.name))
		current = current.get_parent()
	if current != scene_root:
		return ""
	segments.reverse()
	return "/" + scene_root.name + "/" + "/".join(segments)


static func from_node_alt(node: Node, scene_root: Node) -> String:
	if scene_root == null or node == null:
		return ""
	if node == scene_root:
		return "/" + scene_root.name
	var relative := str(scene_root.get_path_to(node))
	if relative.is_empty() or relative.begins_with(".."):
		return ""
	return "/" + scene_root.name + "/" + relative


func _make_chain(depth: int) -> Array[Node]:
	## Returns [root, deepest]
	var root := Node.new()
	root.name = "Main"
	var cur := root
	for i in depth:
		var child := Node.new()
		child.name = "Child%d" % i
		cur.add_child(child)
		cur = child
	var out: Array[Node] = [root, cur]
	return out


func _initialize() -> void:
	# --- correctness cross-check on edge cases ---
	var pair := _make_chain(3)
	var root: Node = pair[0]
	var deep: Node = pair[1]
	var orphan := Node.new()
	orphan.name = "Orphan"
	var foreign_root := Node.new()
	foreign_root.name = "Other"
	var foreign_child := Node.new()
	foreign_child.name = "FChild"
	foreign_root.add_child(foreign_child)
	var super_root := Node.new()
	super_root.name = "Super"
	super_root.add_child(root)  # root now has a parent; test ancestor-of-root

	var cases := {
		"root": [root, root],
		"child": [root.get_child(0), root],
		"deep": [deep, root],
		"null_node": [null, root],
		"null_root": [deep, null],
		"orphan": [orphan, root],
		"foreign": [foreign_child, root],
		"ancestor_of_root": [super_root, root],
	}
	var all_ok := true
	for key in cases:
		var n: Node = cases[key][0]
		var r: Node = cases[key][1]
		var o := from_node_old(n, r)
		var p := from_node_pr(n, r)
		var a := from_node_alt(n, r)
		var ok := o == p and p == a
		if not ok:
			all_ok = false
		print("case=%s old=%s pr=%s alt=%s %s" % [key, var_to_str(o), var_to_str(p), var_to_str(a), "OK" if ok else "MISMATCH"])
	print("correctness: %s" % ("ALL AGREE" if all_ok else "MISMATCH FOUND"))

	# --- timing ---
	const ITERS := 100_000
	for depth in [3, 10, 30]:
		var p2 := _make_chain(depth)
		var broot: Node = p2[0]
		var bdeep: Node = p2[1]
		var s := ""

		var t0 := Time.get_ticks_usec()
		for i in ITERS:
			s = from_node_old(bdeep, broot)
		var t_old := Time.get_ticks_usec() - t0

		t0 = Time.get_ticks_usec()
		for i in ITERS:
			s = from_node_pr(bdeep, broot)
		var t_pr := Time.get_ticks_usec() - t0

		t0 = Time.get_ticks_usec()
		for i in ITERS:
			s = from_node_alt(bdeep, broot)
		var t_alt := Time.get_ticks_usec() - t0

		print("depth=%2d  old=%6.2f us/call  pr743=%6.2f us/call  alt=%6.2f us/call  (pr/old ratio %.2fx)" % [
			depth,
			float(t_old) / ITERS,
			float(t_pr) / ITERS,
			float(t_alt) / ITERS,
			float(t_pr) / float(t_old),
		])
		broot.free()

	super_root.free()
	orphan.free()
	foreign_root.free()
	quit()
```

Raw output:

```
case=root old="/Main" pr="/Main" alt="/Main" OK
case=child old="/Main/Child0" pr="/Main/Child0" alt="/Main/Child0" OK
case=deep old="/Main/Child0/Child1/Child2" pr="/Main/Child0/Child1/Child2" alt="/Main/Child0/Child1/Child2" OK
case=null_node old="" pr="" alt="" OK
case=null_root old="" pr="" alt="" OK
case=orphan old="" pr="" alt="" OK    (alt printed ERROR: Parameter "common_parent" is null)
case=foreign old="" pr="" alt="" OK   (alt printed ERROR: Parameter "common_parent" is null)
case=ancestor_of_root old="" pr="" alt="" OK
correctness: ALL AGREE
depth= 3  old=  1.05 us/call  pr743=  1.50 us/call  alt=  1.06 us/call  (pr/old ratio 1.44x)
depth=10  old=  1.53 us/call  pr743=  3.42 us/call  alt=  1.54 us/call  (pr/old ratio 2.24x)
depth=30  old=  2.66 us/call  pr743=  8.82 us/call  alt=  2.77 us/call  (pr/old ratio 3.31x)
```

</details>

🤖 Review drafted with [Claude Code](https://claude.com/claude-code)

eth0net:
That's a great catch on the GDScript vs C++ performance boundary! 

I'll drop this and leave it as it is, at least for now!

I'm already in the Discord, my name is eth0net there too 🚀

PR #751, opened 2026-07-16
Guard null lifecycle in _process during script reload

shelbykb2:
## Motivation

During dual-plugin enable or script-reload races, `_lifecycle` can be null while `_process` is still armed. That floods the Output dock with null errors every frame.

## Change

- Guard `_lifecycle == null` in `_update_process_enabled` and `_process`.
- Disable process and return early when lifecycle is unavailable.

- [x] Branched from main
- [x] Godot-side tests for new plugin behavior (test_plugin_lifecycle.gd):
  - test_process_null_lifecycle_disarms_without_crash
  - test_update_process_enabled_null_lifecycle_disarms
- [x] ruff check src/ tests/ (no Python surface change)

## Scope

GDScript safety fix + tests. No version bump.

### DIFF
--- a/plugin/addons/godot_ai/plugin.gd
+++ b/plugin/addons/godot_ai/plugin.gd
@@ -947,6 +947,9 @@ func _arm_server_version_check() -> void:
 
 
 func _update_process_enabled() -> void:
+	if _lifecycle == null:
+		set_process(false)
+		return
 	set_process(
 		_lifecycle.get_adoption_watch_deadline_ms() > 0
 		or _lifecycle.is_awaiting_server_version()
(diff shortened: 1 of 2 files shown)


dsarno (approved):
Thanks for this fix — tight, correct, and well-tested.

The guard is right: during dual-plugin enable / script-reload races `_lifecycle` can be null while `_process` is still armed, and disarming via `set_process(false)` + early return is the correct recovery (re-arming happens naturally through the existing `_update_process_enabled` call sites once the lifecycle is reconstructed). Both tests follow the suite's established `GodotAiPlugin.new()` pattern and free the instance before asserting.

Verified beyond CI: parse-clean under `godot --headless --import`, and the full in-editor GDScript suite passes on this branch against a live headless editor (1871/1891, 0 failed, 20 environment skips).

---

PR #752, opened 2026-07-16
Resolve addon junction for dev venv; strip local tags from uvx pins

shelbykb2:
## Motivation

When res://addons/godot_ai is a directory junction or symlink into a source checkout, ProjectSettings.globalize_path only walks the logical project path and never finds the checkout .venv. The plugin then falls through to uvx with a package version that may include PEP 440 local metadata (e.g. 3.0.2+local.1), which is not on PyPI.

## Change

- Resolve addon real path via DirAccess.read_link and walk up from that target for .venv + src/godot_ai.
- Optional GODOT_AI_VENV_PYTHON override.
- Strip +local metadata for uvx pins (_pypi_pin_version); keep pre-release tags like -rc1.
- Clearer dock crash copy for local-build metadata.

- [x] Branched from main
- [x] Godot-side (test_clients.gd): pypi pin, resolve_addons_realpath, junction-depth venv walk
- [x] Python: self-update smoke harness accepts pypi_version pin form
- [x] ruff check src/ tests/

## Scope

Infra only. No new MCP tools. No package version bump.

### DIFF
--- a/plugin/addons/godot_ai/client_configurator.gd
+++ b/plugin/addons/godot_ai/client_configurator.gd
@@ -471,14 +471,13 @@ static func get_plugin_version() -> String:
 	return "0.0.1"
 
 
-## Strip local build metadata for PyPI pins: `3.0.2+grok.1` → `3.0.2`.
+## Strip PEP 440 local build metadata for PyPI pins: `3.0.2+local.1` → `3.0.2`.
+## Pre-release segments (`3.1.0-rc1`) are preserved — only `+…` is removed.
 static func _pypi_pin_version(version: String) -> String:
 	var v := version.strip_edges()
 	var plus := v.find("+")
 	if plus >= 0:
 		v = v.substr(0, plus)
-	## Also strip pre-release style local tags after a second hyphen segment
-	## that is clearly not a numeric patch (keep 3.0.2 as-is).
 	return v
 
 
--- a/plugin/addons/godot_ai/client_configurator.gd
+++ b/plugin/addons/godot_ai/client_configurator.gd
@@ -471,6 +471,17 @@ static func get_plugin_version() -> String:
 	return "0.0.1"
 
 
+## Strip local build metadata for PyPI pins: `3.0.2+grok.1` → `3.0.2`.
+static func _pypi_pin_version(version: String) -> String:
+	var v := version.strip_edges()
+	var plus := v.find("+")
+	if plus >= 0:
+		v = v.substr(0, plus)
+	## Also strip pre-release style local tags after a second hyphen segment
+	## that is clearly not a numeric patch (keep 3.0.2 as-is).
+	return v
+
+
 ## Override for the dev-vs-user heuristic. Accepted values:
 ##   "dev"   — force dev-checkout mode (skip update check + self-install)
 ##   "user"  — force user-install mode (run update check, allow self-install)
@@ -564,6 +575,11 @@ static func get_server_command(refresh: bool = false) -> Array[String]:
 	var uvx := find_uvx()
 	if not uvx.is_empty():
 		var version := get_plugin_version()
+		## PEP 440 local build tags (e.g. 3.0.2+local.1) are not on PyPI.
+		## Pin uvx to the public base version so the server still boots;
+		## checkout-local extras need the dev_venv tier above
+		## (symlink/junction → repo .venv).
+		var pypi_version := _pypi_pin_version(version)
 		## Pin to the EXACT plugin version rather than `~=<minor>`. Under the
 		## tilde form, uvx was happy to reuse a cached tool env that matched
 		## the minor constraint — so an install that first spawned 1.2.0 kept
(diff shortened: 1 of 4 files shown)


dsarno (changes requested):
Thanks for this one — the core is solid and clearly fixes a real pain point. `resolve_addons_realpath()` via `DirAccess.read_link`, the project-first-then-addon-realpath venv walk, the 8-hop depth with the drive-root guard, the `GODOT_AI_VENV_PYTHON` override, and `_pypi_pin_version` (stripping `+local` while keeping pre-release tags) are all sensible and well-tested. Reusing the new helper in the dock's `_resolve_plugin_symlink_target` is a nice dedup.

**One confirmed runtime bug blocks merge**, in `mcp_dock.gd::_crash_body_for_state`, the new CRASHED copy for local builds. In GDScript, `%` binds tighter than `+`, so in:

```gdscript
return (
    "The server exited before the WebSocket handshake. "
    + "Local plugin version is %s (PEP 440 local build metadata) — uvx pins PyPI godot-ai==%s. "
    + ...
    + "python.exe, then Reload Plugin. Log should show 'MCP | using dev venv: ...'."
    % [version, pin]
)
```

the format applies **only to the final fragment**, which has no placeholders. Reproduced against a real Godot binary:

```
ERROR: String formatting error: not all arguments converted during string formatting.
RESULT: [A %s and %s. B tail with no placeholders.]
```

So the exact audience this message was written for (local-build + uvx crash) sees literal `%s` placeholders in the dock, plus a pushed formatting error. The stock-release branch below it only works by accident — its lone `%s` happens to sit in the last fragment.

Fix: format the fully-parenthesized concatenation, e.g. `("…" + "…" + "…") % [version, pin]`, or build the string in a `var` first (and parenthesize the second branch too, for robustness against future edits). Please also add a small GDScript test asserting the rendered CRASHED message contains the actual version strings and no `%s` residue — the current suite passes 1872/1892 with this bug in place precisely because nothing renders that branch.

Minor, non-blocking:
- The dock now calls `ClientConfigurator._pypi_pin_version` — since it's cross-file API, consider dropping the underscore.
- The message says "python.exe" on all platforms.
- Missing second blank line around the newly inserted tests in `test_clients.gd` (suite convention is two).
- Heads-up: this and #753 insert tests at the same spot in `test_clients.gd`; #753 is landing first, so this branch will need a trivial rebase.

Everything else verified: Python smoke-harness tests pass, parse-clean under `godot --headless --import`, and the full in-editor GDScript suite is green on this branch. Fix the format string and this is good to go.

---

dsarno (changes requested):
The direction is right and wanted — junction/symlink installs falling through to a nonexistent PyPI pin is a real failure mode, `_pypi_pin_version` is correctly scoped (strips only `+local`, keeps pre-release tags), and the scratch-dir walk tests are appreciated. Two blocking issues before merge:

**1. Confirmed: the CodeRabbit `%`-precedence finding in `_crash_body_for_state` is a real bug.** GDScript binds `%` tighter than `+`, so in the new concatenated messages the format args apply to the **last fragment only**. In the local-metadata branch (`pin != version`) that last fragment has no placeholders — the `% [version, pin]` errors at runtime and the user sees literal `%s` in the dock, in exactly the scenario this PR exists to explain. (The stock branch works only by luck: its lone `%s` happens to sit in the final fragment.) Fix by parenthesizing the full concatenation before applying `%`, and please add a `test_dock.gd` case that exercises the local-metadata branch — the existing `test_crashed_body_mentions_pypi_propagation_on_uvx_tier` can't reach it, which is why this slipped through.

**2. The `GODOT_AI_VENV_PYTHON` override is unreliable in the primary spawn path.** `McpPathTemplate.env_lookup` intentionally serves **worker threads** only from a main-thread-warmed snapshot and returns `""` for un-warmed keys — and venv discovery runs on the startup walk's worker (`get_server_command` inside `_run_blocking`, the #678 pattern). Since this PR doesn't add `GODOT_AI_VENV_PYTHON` to the warm set, the documented override silently no-ops on a cold auto-spawn (whether it works depends on whether some main-thread caller happened to touch the key first). Add it to `_BASE_ENV_VARS` (or an explicit `warm_env_snapshot(extra)` before the walk dispatch) and it'll behave deterministically.

One question, non-blocking: does Godot's `DirAccess.is_link`/`read_link` actually report Windows **junctions** (vs. symlinks) on the engine versions we support? If not, the Windows half of `resolve_addons_realpath` silently falls back to the logical path — which is safe, but worth a comment either way since junctions are the setup-dev default on Windows.

Ran the GDScript suite over this branch's test additions locally — the new `_pypi_pin_version` and walk tests pass; the broken message branch has no coverage (point 1).

---

dsarno:
Thanks @shelbykb2 — this was a solid diagnosis and fix. Since the branch had drifted behind `main` and we haven't been able to reach you, we've adopted it into #804 with your four commits and authorship preserved, rebased onto current `main`, plus one review-fix commit (a `%`-vs-`+` precedence bug in the new dock crash message, a warning for a dead `GODOT_AI_VENV_PYTHON` override, and conflict/style cleanup in `test_clients.gd`).

Closing in favor of #804. If you're back and want to pick it up again, comment there — happy to hand it back.

---

Issue #736, opened 2026-07-13
Editor boot: ~1.7s+ GDScript compile of the eager preload graph stalls 'Initializing plugins' on every start


dsarno:
Measured on Windows 11, Godot 4.6.2, main post-#735 (2026-07-13), during a user report of the editor sticking for several seconds at **Project initialization → "Initializing plugins…"**.

## Evidence

Two timestamped captures (every console line wall-clock-stamped; plugin startup trace enabled via `GODOT_AI_STARTUP_TRACE=1`):

**1. Headless `--verbose` run** — isolates the compile cost:

```
+3.5s  [ 66% ] first_scan_filesystem | Initializing plugins...
+3.5s  Loading resource: res://addons/godot_ai/plugin.gd
       (1.67s of silence — no loads, no output)
+5.2s  MCP startup trace | begin platform=Windows http_port=18130 ws_port=19630
```

The 1.67s gap sits between Godot loading `plugin.gd` and the **first instruction of plugin code** (`_enter_tree` → `_startup_trace_begin`). Nothing else runs in that window: it is the GDScript parse/compile of `plugin.gd`'s preload closure — effectively the whole addon (~60 scripts: handlers, clients registry, utils), pulled in eagerly through `const X := preload(...)` chains. GDScript has no compiled cache across editor restarts, so this cost recurs on **every** editor boot and every plugin re-enable.

**2. GUI run** (same machine): the pre-`_enter_tree` gap was ~5s — same signature, inflated by GUI-side work (dock scene, themes) and cold caches. This is the stall visible in the "Initializing plugins… 80%" progress dialog.

For contrast, the plugin's own synchronous init is **not** the problem — the startup trace shows 280ms total (settings 4ms, async server-walk kickoff 27ms, core objects 24ms, handlers 30ms, dock UI 195ms), and the expensive Windows probes (powershell/netstat/netsh/discovery) all run off-thread post-#678/#729. Editor baseline before our plugin (CA certs, first scan, theme/doc cache) adds ~2.5s that is not attributable to the addon.

## Proposed direction (needs design, not a quick patch)

Break the eager preload graph so rarely-hot scripts load on demand:

- `plugin.gd` / `dispatcher.gd` currently preload every handler; handlers could `load()` lazily at first dispatch (or at registration time on a deferred frame) instead.
- The clients registry (`clients/_registry.gd`) preloads all 21 client descriptors at plugin load; the dock only needs them on refresh, which already runs on a worker.

[...]

PR #754, opened 2026-07-16
Add workflow_manage helpers for modeling and visual QA

shelbykb2:
## Motivation

Agents repeatedly need modeling guidance, asset inventory, screenshot checklists, multi-shot visual QA, and MCP install hints. An excludable workflow domain keeps tool caps manageable (--exclude-domains=workflow) without new WebSocket commands.

## Change

- New domain workflow / tool workflow_manage (modeling_guidance, asset_pipeline, screenshot_verify, visual_qa, install_hints)
- Wired in domains.py, server.py, tool_catalog.gd, docs/TOOLS.md
- Composition-first handlers; pure-guidance ops do not require write lock

- [x] Branched from main
- [x] Python unit tests: tests/unit/test_workflow_handlers.py
- [x] Domain catalog parity: tests/unit/test_tool_domains.py (Python to tool_catalog.gd)
- [x] ruff check / format on new modules
- [x] No new GDScript WebSocket commands; catalog row covered by parity tests

## Scope

New excludable tool domain only. Independent of the Grok client descriptor PR.

### DIFF
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -181,6 +181,7 @@ Calls take the form:
 | `resource_manage` | `search`, `load`, `assign`, `get_info`, `create`, `curve_set_points`, `environment_create`, `physics_shape_autofit`, `gradient_texture_create`, `noise_texture_create` |
 | `api_manage` | `get_class` |
 | `client_manage` | `status`, `configure`, `remove` |
+| `workflow_manage` | `modeling_guidance`, `asset_pipeline`, `screenshot_verify`, `visual_qa`, `install_hints` |
 | `tilemap_manage` | `tilemap_set_cell`, `tilemap_set_cells_rect`, `tilemap_clear`, `tilemap_get_cells` |
 | `tileset_manage` | `tileset_get_atlas_tiles`, `tileset_get_atlas_image` |
 
--- a/plugin/addons/godot_ai/tool_catalog.gd
+++ b/plugin/addons/godot_ai/tool_catalog.gd
@@ -47,6 +47,7 @@ const DOMAINS := [
 	{"id": "editor", "label": "editor", "count": 4, "tools": ["editor_manage", "editor_reload_plugin", "editor_screenshot", "logs_read"]},
 	{"id": "filesystem", "label": "filesystem", "count": 1, "tools": ["filesystem_manage"]},
 	{"id": "game", "label": "game", "count": 1, "tools": ["game_manage"]},
+	{"id": "workflow", "label": "workflow", "count": 1, "tools": ["workflow_manage"]},
 	{"id": "input_map", "label": "input_map", "count": 1, "tools": ["input_map_manage"]},
 	{"id": "material", "label": "material", "count": 1, "tools": ["material_manage"]},
 	{"id": "node", "label": "node", "count": 4, "tools": ["node_create", "node_find", "node_manage", "node_set_property"]},
--- a/src/godot_ai/handlers/workflow.py
+++ b/src/godot_ai/handlers/workflow.py
@@ -1,0 +1,409 @@
+"""Shared handlers for agent workflow helpers (modeling, assets, visual QA).
+
+These ops are intentionally composition-first: they use existing editor
+commands (filesystem search, screenshots, editor state) rather than new
+plugin WebSocket commands. Pure-guidance ops ignore the runtime connection
+when no Godot session is required.
+
+Exposed via the ``workflow_manage`` MCP rollup (domain id: ``workflow``).
+"""
+
+from __future__ import annotations
+
+import asyncio
+from typing import Any
+
+from godot_ai.handlers import editor as editor_handlers
+from godot_ai.handlers import filesystem as filesystem_handlers
+from godot_ai.handlers import project as project_handlers
+from godot_ai.runtime.direct import DirectRuntime
+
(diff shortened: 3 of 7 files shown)


dsarno (commented):
Thanks for the contribution — this was reviewed carefully, and to be clear up front: **the implementation is not the problem.** It was verified end-to-end and everything is green:

- Full Python suite on this branch: 1383 passed (including the new workflow handler tests, domain-catalog parity, and the resource-form lint).
- Full in-editor GDScript suite against a live headless editor: 1869/1889 passed, 0 failed.
- The wiring is done exactly right: excludable domain, no new WebSocket commands, `tool_catalog.gd` parity, composition over existing handlers.

The recommendation is to **close this PR** anyway, on product fit rather than quality:

1. **Three of the five ops return static prose, not editor capability.** `modeling_guidance` and `install_hints` are hardcoded strings — documentation delivered through a tool call. Guidance like poly budgets and glTF export tips is something capable agents already know, and once baked into the server it drifts silently with no review pressure. `install_hints` also duplicates the README and overlaps the existing `client` domain.
2. **It spends the resource this repo works hardest to conserve.** A lot of engineering here (rollups, `defer_loading`, `--exclude-domains`) exists specifically to keep the tool surface small because real clients choke past ~40 tools. A new domain whose main payload is prose sets the wrong precedent for that budget, even as an excludable one.
3. **Cross-PR text coupling.** `install_hints` hard-codes port 8000 (the port is configurable) and describes `GODOT_AI_VENV_PYTHON` and the `+local` uvx-pin behavior from #752, which hasn't merged — if #752 changes or is rejected, this ships stale instructions.
4. **`visual_qa(run_if_needed=true)` runs the project with `autosave=True`**, which persists any in-memory MCP scene mutations to disk mid-"read-only" QA pass — exactly the autosave-pollution trap AGENTS.md warns about. A QA helper should default `autosave=False`.
5. The `stylized_ethereal` style guide is very specific to one game's art direction (mint flora, restore loops, theme-color emission states) — great content for that project, but not general enough to ship as product surface.

**What would be welcome instead:**

- If the guidance content is valuable, move it to `docs/` (e.g. `docs/ASSET_GUIDELINES.md`) or the server's `instructions=` blurb — reviewable, zero tool-slot cost, and agents still see it.
- `visual_qa` is the one genuinely useful composition here. If multi-shot visual QA proves to be a recurring need, it would be worth revisiting as a single op on an existing rollup (it's editor/game territory rather than a new domain), with `autosave=False` and without the hardcoded checklist text.

Appreciate the care that went into the tests and the domain wiring — the same rigor applied to #751/#752 is exactly what those PRs needed.

---

dsarno (changes requested):
Thanks — the mechanical integration is done right (domain registered in `domains.py` + `server.py` + `tool_catalog.gd` + `docs/TOOLS.md`, excludable via `--exclude-domains`, handler unit tests present). Three concrete problems, then a scope question that needs a maintainer decision before this merges in any form.

**Blocking, concrete:**

1. **`workflow_visual_qa` hardcodes `project_run(mode="main", autosave=True)`.** Autosave-on-run persists any in-memory MCP scene mutations to the `.tscn` on disk — the exact footgun AGENTS.md's "Live-smoke scene hygiene" section exists to prevent, triggered here by a tool whose purpose is *inspection*. A QA helper must not silently commit the scene it's inspecting. Default `autosave=False` (or expose the parameter and default it off).

2. **`workflow_install_hints` documents features from an unmerged PR.** The `dev_checkout_junction` option tells users to set `GODOT_AI_VENV_PYTHON` and describes the `+local`-tag uvx pinning — both of which only exist in #752, which this PR's description says it's independent of. If #752 changes or doesn't land, this tool ships instructions for behavior the server doesn't have.

3. **`workflow_asset_pipeline` ignores its `path` argument for 2 of 5 searches** (CodeRabbit's finding, confirmed): `pbr` scans a fixed `res://assets` and `terrain` a fixed `res://addons` regardless of the caller's `root`. A scoped scan silently reports project-wide results. (The sequential-awaits and catalog-ordering nits from CodeRabbit are also both valid.)

**Scope question (maintainer call, tracked in the review-batch issue):** `modeling_guidance` and `install_hints` are static prose — style guides and install docs served as MCP tools. That's a new *kind* of surface for this server: it can't reflect the actual project, it goes stale independently of the code, it duplicates the README, and the `stylized_ethereal` palette ("mint/teal leaves", "restore loops") reads as one specific game's art bible baked into every install. The screenshot ops are different — `screenshot_verify`/`visual_qa` compose real editor capabilities and could stand on their own. My suggestion: split the PR — the two capture/QA ops (with fix 1) as one reviewable unit, and pitch the guidance-text ops separately as docs or a client-side skill rather than tools.

---

dsarno:
Thanks @shelbykb2 for the proposal and the work that went into it. After review, the maintainers have decided not to take the repo in this direction: `workflow_manage`'s ops are largely canned guidance text shipped as MCP tools, which cuts against the current push to trim tool-surface token cost and editor-boot overhead (#736, #792), and the branch now conflicts with the reworked domain-exclusion surface in `server.py`.

This is a scope decision rather than a code-quality one, so we're closing the PR rather than asking for a rebase. If you (or anyone) want to pursue the underlying idea — agent-facing modeling/QA guidance — the right first step is a design issue discussing where that content should live (docs, skills, or resources rather than tools), and whether it belongs in this repo. Closing. Thanks again!

---

Issue #525, opened 2026-06-09
[robustness] managed_record kill-proof trusts a recorded PID without a cmdline brand check


dsarno:
**Severity:** Low · From pre-release robustness review (2026-06-09, `be40e1c`, v2.6.0)

**Location:** `plugin/addons/godot_ai/plugin.gd:1091-1093`

**Risk:** `_evaluate_strong_port_occupant_proof` returns `record_pid` as a kill target when it is an alive listener on the port, but does not brand-verify it via `_pid_cmdline_is_godot_ai` (the sibling `pidfile_listener`/`status_matches_record` branches are stricter). If the recorded PID is recycled to an unrelated process that binds the configured port, that process could be killed. Narrow window (PID reuse + exact port).

**Fix:** Add the brand gate to the `managed_record` branch so all automatic kill targets clear the same bar.

dsarno:
Tracked in #573 (self-update hardening umbrella).

---
_Generated by [Claude Code](https://claude.ai/code)_

dsarno:
Fixed by #575 (merged). The `managed_record` kill branch in `_evaluate_strong_port_occupant_proof` now requires the `_pid_cmdline_is_godot_ai_for_proof` brand check, matching the sibling `pidfile_listener` branch, so a recycled PID bound to the same port can't be selected as a kill target. Covered by a new regression test (`test_strong_proof_rejects_unbranded_managed_record_pid`). Closing as completed.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #757, opened 2026-07-16
dock: don't show the Install-uv setup section while the server launch is settling

dsarno:
## Motivation

Closes #744.

While the plugin is launching, the dock shows the Setup section with a red "uv: not found" row and an **Install uv** button. The visibility check (`_apply_dev_mode_visibility`) only consulted the cached uv probe, which mid-launch is often wrong or irrelevant:

- the startup probe can fail transiently (the #739 shape — contended spawn, cold Defender scan, stale PATH), and it only self-corrects when the server connects;
- the launch may be succeeding via the `.venv` or system-CLI tiers, where uv isn't needed at all.

Either way the user watches "Starting server…" next to install guidance they don't need.

## Change

- The uv-missing branch of the Setup section's visibility is now gated on the server launch having **settled**: connected, a terminal diagnosis (`CRASHED` / `NO_COMMAND` / `PORT_EXCLUDED` / `INCOMPATIBLE` / `FOREIGN_PORT` — `NO_COMMAND` is exactly when "Install uv" is the right guidance), or startup-grace expiry (the same window that renders "Starting server…", so the amber status text and the Setup section agree on what "still launching" means).
- `_update_status` re-applies visibility on every status transition (`changed` only, so nothing new runs per-frame), so the section appears the moment the launch outcome makes it relevant.
- The decision is factored into a static `_setup_section_should_show(dev, uv_missing, launch_pending)` plus an instance `_server_launch_pending()`, so the truth table is testable without faking the uv probe or a dev checkout. Dev-toggle behavior is unchanged (always shows the section).

## Tests

- New GDScript tests in `test_project/tests/test_dock.gd`: the visibility truth table, and `_server_launch_pending` across spawning/terminal/connected/grace-expired states (using the existing `_RestartDispatchPlugin` stub).
- Full GDScript suite against a live headless editor (Godot 4.6.2, CI-style `ci-start-server` + `ci-godot-tests` flow): **1873/1891 passed, 0 failed, 18 skipped** (the skips are the pre-existing environment-gated ones).
- `ci-check-gdscript` on the import log: all GDScript files OK.
- No Python surface change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PiPyzz94bmAjjDYEhzc8JY

---
_Generated by [Claude Code](https://claude.ai/code/session_01PiPyzz94bmAjjDYEhzc8JY)_

## Summary by CodeRabbit

* **Bug Fixes**
[...]

Issue #498, opened 2026-05-30
Reaper altitude: session-idle self-terminate vs. owner-PID polling


dsarno:
Follow-up to #495. The orphan reaper keys on the **owning editor's PID**: poll `pid_alive(owner)` and, once it's dead with zero sessions, self-terminate. This works on POSIX, but the PID channel has costs worth weighing against an alternative.

**Limitations of the owner-PID approach:**
- **PID reuse** can cause a *permanent* missed reap (bare-pid check, no identity proof — unlike the plugin's `_pid_cmdline_is_godot_ai` gating).
- **Windows** `os.kill` semantics forced gating it off entirely (see companion issue).
- Adds plumbing: a `GODOT_AI_OWNER_PID` env channel + a per-spawn set/unset dance in two spawn paths.
- Adoption correctness leans on a session-count guard + a grace re-check to avoid wrong reaps during reconnect blips.

**Alternative: session-idle self-terminate.** The server exits after a grace window during which **zero editor sessions** have been connected (idle-since-last-disconnect, plus a boot grace).
- **Pros:** no owner PID needed (drop the env channel); covers Windows for free; sidesteps PID reuse entirely; single mechanism.
- **Risks to design around:** boot window before the first editor connects; the reconnect gap during a plugin reload (grace must exceed it); legitimately-idle dev servers (dock "Start Dev Server" / `--reload`) that may run with no editor — likely needs an opt-out.

**Decision needed:** keep owner-PID (works today, POSIX-only), migrate to idle-self-terminate (cleaner, cross-platform), or do both (idle as a cross-platform backstop, owner-PID as a fast path).

Refs: #495 review (altitude finding), `src/godot_ai/orphan_reaper.py`.

dsarno:
Queue-decision sweep — calling this decision so implementation can proceed: **both, staged, with idle as the destination.**

1. **Land session-idle self-terminate as a cross-platform backstop** (opt-out env var for legitimately-idle dev servers — dock "Start Dev Server" and `--reload` should set it automatically). Grace window must exceed the worst observed plugin-reload reconnect gap with margin (the stormtest data showed ~6s recovery; a 60-120s idle grace is comfortably safe), plus a boot grace before the first session.
2. **Keep owner-PID as the POSIX fast path for now** — it works today and bounds the idle-grace exposure window.
[...]

PR #758, opened 2026-07-16
startup: adopt a crash-surviving server the bind probe missed

### LINKED ISSUE #745: MCP is not detect after crash or new window open
In very rare cases the window of Godot crashes or another window is opened. In that cases the MCP server continues to run in the background. Instead of checking if the MCP server is running the add-on checks for the port availability only. When the port is not available the MCP server might already be running and should be able to connect it. If the MCP server do not reply properly through the port call the it is likely another application. I can also see the pid file in the project cache that might be used but it is not enough.

dsarno:
## Motivation

Closes #745.

After an editor crash (or a second editor window), the managed MCP server keeps running in the background. But the startup walk's first gate is only "can I bind the HTTP port?" (`_is_port_in_use` → `can_bind_local_port`). Two ways that reports **free** while a live server is listening:

- On **Windows**, a `SO_REUSEADDR` bind can succeed straight over an existing listener that didn't set `SO_EXCLUSIVEADDRUSE` (uvicorn doesn't) — the Windows branch of `_is_port_in_use` returns "free" on a successful bind without a scrape.
- The netstat/lsof scrape fallback can fail transiently (stripped installs, locale issues, contended spawn).

When that happens, the walk never reaches the existing (well-tested) probe/adopt/recover branch and blind-spawns a duplicate server — the reporter's "MCP is not detected" symptom.

## Change

Exactly what the issue asks for: check whether the server is *running*, not just whether the port is bindable, and use the pid-file the reporter pointed at as the surviving evidence.

When the bind probe says free, the walk now runs one extra evidence check (on the same off-main-thread worker as the rest of the walk):

1. **Pid-file** (`user://godot_ai_server.pid`, written by the server via `--pid-file`, survives a crash) names a PID,
2. that PID is **alive**, and
3. its cmdline carries the **godot-ai brand** (the #525 gate, so a kernel-recycled PID can never redirect startup), and
4. the **HTTP status probe** on our port actually answers as a godot-ai server ("if the MCP server does not reply properly through the port call it is likely another application" — per the issue),

→ then the port is treated as occupied and the *existing* adopt/recover branch runs: compatible server → adopted (`READY`), version drift → the established branded kill+respawn recovery. No new kill paths and no new terminal states: any tier failing (no pid-file, dead/stale/foreign pid, unresponsive port) falls through to the unchanged spawn path, and the check costs nothing when no pid-file exists.

## Tests

- `test_managed_server_evidence_requires_live_branded_pidfile` — the gate's truth table (no pid-file / dead pid / live-but-unbranded / live+branded).
- `test_start_adopts_crash_survivor_when_bind_probe_reports_port_free` — full startup walk with a stubbed "bind says free, pid-file + status probe say godot-ai" environment lands in `READY` with the survivor adopted as managed, confirming the HTTP probe ran.
- Full GDScript suite against a live headless editor (Godot 4.6.2, CI-style flow): **1873/1891 passed, 0 failed, 18 skipped** (+2 over baseline = the new tests).
- `ci-check-gdscript`: all GDScript files OK. No Python surface change.

The Windows `SO_REUSEADDR` reproduction itself can't run in this Linux environment — the stale/foreign-server CI smoke jobs cover the cross-OS recovery machinery this change routes into, and the new unit tests pin the gate's behavior on all tiers.

https://claude.ai/code/session_01PiPyzz94bmAjjDYEhzc8JY

---
_Generated by [Claude Code](https://claude.ai/code/session_01PiPyzz94bmAjjDYEhzc8JY)_

### DIFF
--- a/plugin/addons/godot_ai/plugin.gd
+++ b/plugin/addons/godot_ai/plugin.gd
@@ -1165,6 +1165,21 @@ func _find_managed_pid(port: int) -> int:
 	return _find_pid_on_port(port)
 
 
+## #745: after an editor crash the managed server keeps running, but the
+## bind probe can still report the HTTP port as free — Windows lets a
+## SO_REUSEADDR bind succeed straight over a live listener, and the OS
+## scrape fallback can fail transiently. The pid-file the server writes
+## via `--pid-file` survives the crash; when it names a live process
+## whose cmdline carries the godot-ai brand, a server is likely still
+## up. Liveness + brand only — the startup walk confirms with the HTTP
+## status probe before changing behavior, so a kernel-recycled PID can
+## never redirect startup on its own. Uses the `_for_proof` seams so
+## lifecycle tests can stub it without touching real processes.
+func _managed_server_evidence_alive() -> bool:
+	var pid := _read_pid_file_for_proof()
+	return pid > 0 and _pid_alive_for_proof(pid) and _pid_cmdline_is_godot_ai(pid)
+
+
 ## `live` is the result of a prior `_probe_live_server_status_for_port`
 ## call that the caller already has on hand. When non-empty it short-
 ## circuits the internal probe at the bottom of this helper, so a single
--- a/plugin/addons/godot_ai/plugin.gd
+++ b/plugin/addons/godot_ai/plugin.gd
@@ -1177,7 +1177,7 @@ func _find_managed_pid(port: int) -> int:
 ## lifecycle tests can stub it without touching real processes.
 func _managed_server_evidence_alive() -> bool:
 	var pid := _read_pid_file_for_proof()
-	return pid > 0 and _pid_alive_for_proof(pid) and _pid_cmdline_is_godot_ai(pid)
+	return pid > 0 and _pid_alive_for_proof(pid) and _pid_cmdline_is_godot_ai_for_proof(pid)
 
 
 ## `live` is the result of a prior `_probe_live_server_status_for_port`
--- a/plugin/addons/godot_ai/utils/server_lifecycle.gd
+++ b/plugin/addons/godot_ai/utils/server_lifecycle.gd
@@ -636,6 +636,29 @@ func _start_server_impl(async_gen: int) -> void:
 	))
 	if _async_stale(async_gen):
 		return
+	if not port_in_use:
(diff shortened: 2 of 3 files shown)


dsarno:
## Windows hand-test — the #760 checklist item — PASS ✅

Environment: Windows 11, Godot 4.6.2 (GUI editor), plugin + server from this branch merged with #757 on top of `main` (`aa2b598`), dev-venv server on HTTP 18130 / WS 19630.

**Reporter scenario, natural run**: launched the editor (managed server spawned, pid-file written), hard-killed the editor à la Task Manager (`TerminateProcess`), server survived, relaunched the project →

```
MCP | plugin loaded
MCP | adopted managed server (PID 5100, live v3.0.2, WS 19630, plugin v3.0.2)
MCP | connected to server
```

Single listener on 18130 (same PID, no duplicate spawn), 43-tool MCP round-trip healthy after adoption.

**One honest nuance**: on this box the natural run reaches adoption via the *pre-existing* branch, not the new gate — Godot's bind probe actually **fails** over the live uvicorn listener (measured directly: a `SO_REUSEADDR` bind over it returns `WinError 10013`; asyncio doesn't set `SO_REUSEADDR` on Windows), and the netstat scrape works, so `_is_port_in_use` correctly says busy. So to pin the **new evidence gate** specifically, I re-ran with an env-gated local patch forcing `_is_port_in_use → false` — the exact pre-fix blind-spawn condition (bind-says-free / scrape-fails). Result: **adoption still landed**, with every evidence tier running for real — pid-file read, `tasklist` liveness, PowerShell cmdline brand walk (the pid-file names the real interpreter under the uv shim; its cmdline carries `-m godot_ai` so the brand hits on hop 1), and the HTTP status probe. The negative tier also got a natural exercise: a stale pid-file naming a dead PID fell through to a clean normal spawn, no error.

**Suite**: full in-editor GDScript run on Windows: **1900 total, 1881 passed, 0 failed, 19 skipped** (includes this PR's 2 new tests + #757's).

Two environment notes for posterity (neither affects the fix): killing the *console-wrapper* exe (`_console.exe`) takes the whole process tree down including the server — the crash-survivor scenario needs the GUI editor killed, which is what real users run; and the surviving server is eventually reaped by the #498 idle backstop if no editor returns, so the adoption window isn't indefinite.

PR #779, opened 2026-07-21
project_run: stop leaking stale errors from past runs into launch responses

ViZAlice:
Before the fix: _recent_editor_errors_since() falls back to stale errors from the buffer when the current run produces no new errors (scope = "retained_recent"). These could be issues fixed 10 minutes ago, cluttering the response with outdated context and confusing the AI agent.

After the fix: stale errors are cleared only when the game launches successfully. If the game fails to start (crash, script error, timeout, etc.), previous errors are preserved for debugging. This mirrors Unity's behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented retained editor errors from carrying over into success messaging when the project run state is confirmed as live.
  * Added logic to clear retained-error data and normalize related scope/flags after the live-state decision.
  * Expanded automated coverage with a new unit test to verify the retained-error reset behavior.

### DIFF
--- a/plugin/addons/godot_ai/handlers/project_handler.gd
+++ b/plugin/addons/godot_ai/handlers/project_handler.gd
@@ -296,9 +296,14 @@ static func _run_project_liveness_decision(status: Dictionary, errors_info: Dict
 	var recent_errors: Array = errors_info.get("errors", [])
 	var errors_scope := str(errors_info.get("scope", "none"))
 	var truncated := bool(errors_info.get("truncated", false))
-	## When the game launches successfully, don't carry forward retained errors
-	## from past runs — they were already reported in earlier run responses and
-	## only add noise to the current response.
+		## When the game launches successfully, clear errors that predate this
+		## run's window — they don't belong in a successful launch response and
+		## only add noise. They remain reachable via logs_read(source='editor') and
+		## the retained buffer so failed-run debugging is unaffected.
+		## #635 tradeoff: a genuine in-run error whose Errors-tab row carries an
+		## empty or byte-identical time text can be misclassified as
+		## retained_recent and will be dropped here. It is still reachable via
+		## logs_read.
 	if state == "live" and errors_scope == "retained_recent":
 		recent_errors = []
 		errors_scope = "none"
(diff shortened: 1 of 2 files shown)


dsarno (changes requested):
Thanks — this is a well-scoped fix and a nice first PR. The change is correct: the message path already ignored retained errors on `live` (`test_run_project_liveness_decision_live_ignores_retained_errors_in_message`), and this extends that to the data payload, which is what actually confuses agents. I verified that downstream `split_errors_by_scope` handles the `"none"` scope cleanly, that the retained buffer itself is untouched (so failed runs still get the fallback), and that the `truncated` flag can't go inconsistent (the retained-scope branch always returns `truncated: false` in `mcp_debugger_plugin.gd`). The new test pins exactly the right contract.

Two changes requested before merge — both small:

1. **Record the #635 tradeoff in the new comment** (`project_handler.gd:297-302`). `surfaced_error_tracker.gd:122` documents a boundary case where a genuine *in-run* error can be misclassified as `retained_recent` (empty or byte-identical row time text). Before this PR such an error still appeared in the live payload flagged `may_predate_run`; now it's dropped from the response entirely (it remains reachable via `logs_read(source='editor')`). That's an acceptable tradeoff, but it should be on record at the clearing site so the next person touching #635 knows this path swallows it. Please add a sentence to the comment.

2. **Fix the comment's justification.** "they were already reported in earlier run responses" isn't guaranteed — retained errors can originate from editor activity *between* runs and may never have appeared in any run response. Please reword to the actual invariant: these errors predate this run's window, so they don't belong in a successful launch response, and they stay available via `logs_read` and the retained buffer.

---

dsarno (commented):
Thanks for the quick turnaround — 716d7a4 addresses both review points: the justification now states the real invariant, and the #635 misclassification tradeoff is on record at the clearing site. Content-wise this is exactly what was asked.

One issue introduced by the fixup: the comment block is now indented with **two tabs** while the `if` it documents (and the surrounding statements) sit at one (`project_handler.gd:299-306`). Godot's parser ignores comment lines, so CI's parse check won't catch it, but the block visually attaches to the wrong nesting level. Please dedent the eight `##` lines to a single tab so the comment sits at the same depth as the code it describes.

Also, please make the commit message English to match the repo's history (`git commit --amend` is enough) — or if the maintainers squash on merge, just make sure the squash title is English.

With the indentation fixed this is good to go from my side.

---

dsarno (approved):
8c56202 resolves the last point — the comment block sits at one tab with the code it documents, the justification states the real invariant, and the #635 tradeoff is recorded at the clearing site. All review points addressed; approving.

(If this repo merges without squashing, please reword the three commit messages to English first — with a squash-merge the English PR title covers it.)

---

ViZAlice:
Sorry, my agent pushed these automatically. I'll take care of it.

Issue #778, opened 2026-07-20
filesystem_manage.reimport should distinguish non-imported files such as .gd scripts


Clubhouse1661:
Follow-up to #765. The reimport handler currently calls EditorFileSystem.update_file(path) and appends every existing path to reimported, including .gd scripts. Scripts are not imported resources, so the success response can be misread as proof that parsing or diagnostics ran. The #765 server-side work documents this limitation. The behavioral fix should report non-imported file types distinctly, for example with a skipped_non_imported list/count, while preserving normal reimport behavior for imported assets such as textures, models, and audio. Add GDScript coverage for mixed imported and non-imported path batches.

PR #780, opened 2026-07-21
Accept flat params for manage tools

### LINKED ISSUE #765: Agent UX: rollup tools' nested `params` contract causes repeated "unexpected keyword argument" retry loops (and pushes weak agents toward destructive workarounds)
## What happened

Across a 22-session agent-behavior eval (see #761 — Sonnet 5 and Haiku 4.5 driving the MCP tools on a scaffolding task), agents repeatedly passed rollup arguments flat instead of nested under `params`:

```
script_manage {"op": "read", "path": "res://src/sim/sim_clock.gd"}
-> INVALID_PARAMS: script_manage.read: ... Unexpected param(s): 'path'. Accepted params for op 'read': ...
```

The structured error names the accepted params, but agents frequently did not recover from it. Observed patterns, roughly ascending in cost:

1. Retry the same flat shape once or twice, then give up on that read entirely (several Sonnet + Haiku sessions).
2. Pivot to a different tool for the same job (e.g. `filesystem_manage` read_text — hitting the same wall).
3. Conclude the *tooling* is broken and report a "tool-environment bug" to the user in their completion report (two sessions reported exactly that, listing `script_manage.read`, `filesystem_manage.read_text`, `api_manage.get_class`, `project_manage.settings_get`, `node_manage.delete`, `batch_execute` as "rejecting their own documented parameters").
4. Reach for `editor_reload_plugin` as a repair hammer — which in headless mode permanently bricks the session (#764). 5 of 12 Haiku sessions did this, most immediately after a params-validation error.

At least 3 of 10 Sonnet sessions and effectively all Haiku sessions burned calls on this; one session's flailing pushed it from ~20 calls to 57.

## Why the current mitigation isn't landing

The error text is good — it names the op's accepted params. But it says *what* is accepted, not *where* it goes: the actual fix is moving `path` inside a nested `params` dict (`{"op": "read", "params": {"path": ...}}`), and nothing in the message says that. Agents read "Accepted params for op 'read': 'path'", see they passed `path`, and conclude the validator is contradicting itself — verbatim from one report: *"tools reject their own documented parameters with Unexpected keyword argument even though the same tools report those exact parameter names as required/accepted when omitted."* That reading is wrong but understandable, and it's what escalates to the destructive workarounds above.

Precedent: #88 (flat-kwarg naming mismatch, fixed by rename/alias). This is the same friction class, one level up — the nesting contract itself.

## Possible fixes (any one would help)

1. **Accept top-level aliases**: when an unexpected top-level key exactly matches an accepted param for the given op, fold it into `params` (optionally with a soft deprecation note in the response). Kills the whole failure class.
2. **Sharpen the error text** to state the shape, e.g.: `'path' must be nested under 'params': {"op": "read", "params": {"path": ...}}`. Cheapest change; directly contradicts the "validator is broken" misreading.
3. Both — alias silently, and keep the sharpened message for genuinely unknown keys.

## Related smaller observation

`filesystem_manage(op="reimport")` on a `.gd` path succeeds (`"reimported": [...]`) but is a no-op for parse/diagnostic purposes — scripts aren't imported resources. Agents used it as a "revalidate this script" button and trusted the success response. Either rejecting `.gd` paths with a pointer to the actual idiom (a no-op `script_patch`/`script_create` re-save, which returns fresh diagnostics) or documenting that idiom in TOOLS.md would match the intuition agents demonstrably have.

Clubhouse1661:
## Summary

- Accept flat top-level op parameters on registered `*_manage` tools by folding them into the canonical `params` object before Pydantic validation.
- Add precise diagnostics for conflicting duplicates, non-dict `params`, missing required parameters, and incorrectly nested wrapper fields.
- Preserve complete multi-error validation diagnoses and existing op-typo suggestions.
- Document the canonical call shape and clarify that `.gd` files are not imported resources.

Closes #765.

## Behavior

The canonical nested form remains unchanged:

```json
{"op": "read", "params": {"path": "res://player.gd"}}
```

When transmitted by the client, this compatibility form now works too:

```json
{"op": "read", "path": "res://player.gd"}
```

The middleware:

- only applies to tools registered in `MANAGE_TOOL_OPS`;
- keeps `op`, `params`, and `session_id` top-level;
- validates all duplicates before mutating the request;
- deduplicates equal JSON values without conflating booleans and numbers;
- reports conflicting top-level/nested values explicitly;
- rewrites only pure top-level-extra validation failures.

Missing required arguments now include a concrete nested-call example. Already-nested invalid keys retain the existing accepted/unexpected-parameter diagnosis. Incorrectly nested `op` or `session_id` fields receive specific top-level-sibling guidance.

One secondary behavior change: unknown, non-allowlisted client-wrapper kwargs on manage tools are now folded and surface through the dispatcher as `Unexpected param(s)` rather than as schema-boundary errors. This is more actionable, but worth noting for future wrapper-compatibility triage.

## Documentation

- Documented the flat compatibility alias while retaining the nested form as canonical.
- Clarified that `filesystem_manage.reimport` is intended for imported assets. A successful `.gd` entry only refreshes the editor filesystem cache and does not prove parsing or diagnostics occurred.
- The behavioral `.gd` reimport improvement is tracked in #778.

## Testing

- `ruff check src tests` — passed.
- Focused final-branch suite — 83 passed.
- Full Python suite — 1,404 passed and 13 skipped; two port-preflight tests were initially blocked by the active Godot server on port 9500, then both passed after the editor was closed.
- Godot suite — 1,887 passed, 19 skipped, 0 failed across 63 suites.
- Fresh-server live smoke:
  - flat `script_manage` read succeeded;
  - bogus flat key reached the dispatcher and returned `Unexpected param(s)` without a false nesting claim;
  - missing `path` returned the canonical nested-`params` example.

## Summary by CodeRabbit

* **New Features**
  * Manage-tool calls now accept operation-specific parameters in a flat form, which is folded into the canonical nested `params` structure before validation.
* **Bug Fixes**
  * Added clearer, more targeted error details (required/missing/unexpected, examples, and contract guidance) for misplaced or conflicting parameters, including `op`/`session_id` sibling misuse.
  * Improved operation typo hints by unwrapping wrapped validation errors more reliably.
* **Documentation**
  * Updated tool documentation, middleware order guidance, and `filesystem_manage.reimport` behavior notes.
* **Tests**
  * Expanded integration/unit coverage for folding, conflicts, and the improved error messaging.


dsarno (changes requested):
Well-engineered. What I verified:

- Middleware position is right — inside `ParseStringifiedParams` (so a stringified `params` is decoded before the non-dict check) and outside `HintOpTypoOnManage` — and the order docstring, `AGENTS.md`, `docs/TOOLS.md`, and the order-lock test all moved in lockstep.
- Mutation discipline is careful: all duplicates are validated before any fold, so a late conflict can't leave the call half-rewritten, and `_same_json_value` correctly refuses to conflate `true`/`1`.
- The error-rewrite path is strictly gated (every Pydantic error must be a single-string-loc extra error), so unrelated validation failures keep their raw diagnostics.
- The shared `_find_pydantic_validation_error` (cause/context walk with a cycle guard) is a genuine improvement over the four hand-rolled unwrap branches it replaces in `op_typo_hint`.
- The one behavior change — non-allowlisted client-wrapper kwargs now folding and surfacing as dispatcher `Unexpected param(s)` — is called out in the PR body rather than slipped in. Reasonable trade.

Changes requested:

1. **Truncate caller values echoed into error messages** (`fold_flat_manage_params.py`): `_call_example` embeds `arguments.get(key)` verbatim, and the duplicate-conflict error interpolates both values with `!r`. A flat `content` key carrying a whole script file turns the error into a multi-KB string that gets shipped back to the client (and logged). Please cap embedded values (e.g. repr truncated to ~120 chars with an ellipsis) in both `_call_example` and the conflict message. The `dispatch_manage_op` example builder is safe (it uses a literal `...`), so this is contained to the middleware.

2. **Promote `_find_pydantic_validation_error` to a shared helper.** `fold_flat_manage_params.py` importing a private, underscore-prefixed function from `op_typo_hint` makes the dependency invisible to anyone refactoring `op_typo_hint`. Now that it has two consumers, move it to a shared middleware module (or drop the underscore and export it) so the contract is explicit.

3. **Fix the `_call_example` fallback quoting**: the non-serializable fallback builds a JSON-shaped example with Python `!r` quoting (single quotes), so the "correct shape" the client is told to copy isn't valid JSON. Build the fallback with `json.dumps` on the key and a placeholder value instead, so the example is always copy-pasteable.

---

dsarno (approved):
The three response commits resolve all review points: echoed caller values are now bounded by `_truncate` (120-char cap) in both the call example and the duplicate-conflict message, `find_pydantic_validation_error` moved to the shared `middleware/_validation_errors.py` with both consumers updated, and the `_call_example` fallback emits valid JSON. The tightened `_extra_params_hint` gating (only rewriting keys the caller actually sent flat, via the pre-fold snapshot) is a further improvement over what was asked, with branch coverage added. All checks green. Approving.

---

Issue #647, opened 2026-07-06
No helpful Message if TCP Port is in use


Fox-Alpha:
Wenn i start Godot with the Plugin an another service is using the Port 8000, there comes no helpful Message to hint to this point.

I'm sure the TCP Stack provide a status that says port_in_use or something like this.

The Server should handle this. The UI should show a proper Message.

<img width="640" height="360" alt="Image" src="https://github.com/user-attachments/assets/f70e2bc2-9815-4b04-94c9-dfc44e23491d" />

After Stop my docker Container ...

<img width="262" height="264" alt="Image" src="https://github.com/user-attachments/assets/d08d4afc-87c5-4e0d-b713-be9a8b91a8dc" />

dsarno:
@Fox-Alpha Thanks, fixed in latest release (2.9.2).  Let us know if any issues.

PR #796, opened 2026-07-21
tests: allocate free WebSocket ports so concurrent pytest runs don't collide

dsarno:
## Problem

The Python test suite hardcoded its WebSocket ports, so two concurrent `pytest` runs (common: two Claude Code sessions in different worktrees of the same clone) collided — the second run's `mcp_stack` either failed to bind ("WebSocket port 19502 already in use") or its mock plugin connected to the *other* run's server and died with "received 4001 session id already registered", producing dozens of spurious failures/errors.

Hardcoded sites: `tests/conftest.py` (`mcp_stack` → 19502, `harness` → 19500), the two direct-dial call sites in `tests/integration/test_mcp_tools.py`, and one more of the same class found while fixing: `tests/integration/test_websocket.py::test_env_token_reaches_enforcement_end_to_end` → 19533.

## Fix

- `tests/conftest.py`: new `allocate_free_port()` helper (bind an ephemeral socket to port 0 on loopback, note the port, release it) and a session-scoped `mcp_ws_port` fixture. `mcp_stack` builds its server on `mcp_ws_port`; `harness` allocates per-instance (its old comment claiming a fixed port was required was stale).
- `tests/integration/test_mcp_tools.py`: `test_reload_cycle` and `TestPerCallSessionRouting._connect_second_plugin` take the port from the `mcp_ws_port` fixture instead of hardcoding 19502.
- `tests/integration/test_websocket.py`: the env-token end-to-end test allocates its port instead of pinning 19533.
- `tests/unit/test_cli_reload.py`: the two `main()` HTTP-bind tests now pass an explicit free `--ws-port`. This surfaced during verification — without the flag, `main()`'s #647 preflight probes the **default 9500** and exits 98 whenever a live dev server (a running Godot editor stack) holds it, i.e. the same fixed-port collision class against the production port.

No `src/` changes; the diff is tests-only.

## Verification

- `ruff check src/ tests/` — clean (`ruff format --check` drift in 9 untouched files pre-exists on main under ruff 0.15.10 and is left alone).
- Full `pytest -q` — **1399 passed, 4 skipped** (including the two `test_cli_reload` tests that fail on main while a live dev server holds 9500).
- Two simultaneous `pytest tests/integration -q` runs in different worktrees of the same clone (the exact collision scenario): both **332 passed, 2 skipped**, fully overlapping in time (~55 s each).
- GDScript `test_run` smoke skipped: the diff touches only Python tests, no plugin or server code.

## Summary by CodeRabbit

* **Tests**
  * Improved WebSocket reliability by dynamically allocating free ports for both integration and end-to-end scenarios.
  * Reduced port conflicts in parallel/repeated runs by removing fixed WebSocket port usage and threading the chosen port through reconnect/routing flows.
  * Expanded CLI reload coverage to use explicitly allocated HTTP/WS ports when validating `--allow-host` binding behavior.
  * Updated dev script compile checks to write generated bytecode into a per-test temp directory.


dsarno:
Branch updated onto current main (`117856e`) — was 13 commits behind with conflicts in `test_cli_reload.py`.

**Conflict resolution:** main had independently fixed the same CLI bind-test collisions with a local `_free_ports(2)` helper. Kept main's atomic multi-bind semantics (all probe sockets held open until every port is allocated, so HTTP/WS can never get the same port) but centralized it as `allocate_free_ports(count)` in `tests/conftest.py` per this PR's intent; `allocate_free_port()` remains as the single-port form; the local duplicate is gone.

**Verification on Windows 11:** `ruff` clean; single run 1429 passed / 13 skipped; then **two full `pytest` suites run concurrently in one worktree** — the scenario this PR exists for. First attempt surfaced one adjacent bug of the same fixed-resource class: `test_dev_scripts_compile` races on the shared `script/__pycache__/*.pyc` atomic replace (WinError 5). Fixed by compiling to the per-test `tmp_path` (`0fa0086`). After the fix: both concurrent runs **1429 passed / 0 failed**.

Issue #146, opened 2026-04-21
Server-crash output is invisible in the dock (WinError 10013 on Windows is the canonical case)


dsarno:
> **🤖 Note to the issue bot — READ THIS FIRST.**
>
> An open PR already addresses this issue: **#148** (chosen as the canonical PR after a duplicate was filed against this issue).
>
> **Do not write any more PRs for this issue.** If you scan this issue and are tempted to open another one, stop.
>
> If you have no other open issues you can productively work on, the preferred idle behaviors are, in order:
> 1. **Go to sleep.** Do nothing. Another duplicate PR is worse than no PR.
> 2. **Audit a core system of the repo** (protocol, dispatcher, session registry, handler/runtime layer, undo/redo correctness, test-runner guardrails, client configurator) and file findings as an issue — not as a PR.

---

## Problem

When the plugin spawns the MCP server via `OS.create_process` ([plugin.gd:345](https://github.com/hi-godot/godot-ai/blob/main/plugin/addons/godot_ai/plugin.gd#L345)), stdout and stderr are dropped on the floor. If the server exits seconds after spawn, the dock shows nothing useful — just `reconnecting in 10s (attempt N)` climbing forever, with the setup section still green (uvx ✓, godot-ai ✓).

A user can't tell the difference between "still starting up", "WebSocket port firewalled", and "server crashed at startup with a clear error message that would have told you the answer". The only way to recover is to know to re-run the spawn command manually in a terminal — that's a high diagnostic floor for a turn-key plugin.

## Motivating case: Windows port reservation (WinError 10013)

A user on Windows 11 / Godot 4.6.2 / Codex hit this. Setup detection was green; dock spun forever. Running the server by hand revealed:

```
godot_ai.transport.websocket | Starting WebSocket server on port 9500
INFO:     Started server process [29096]
websockets.server | server listening on 127.0.0.1:9500
ERROR:    [Errno 13] error while attempting to bind on address ('127.0.0.1', 8000):
          [winerror 10013] an attempt was made to access a socket in a way forbidden by its access permissions
INFO:     Application shutdown complete.
```

WinError 10013 is not "port in use" (that's 10048). It means Windows has *reserved* the port range — typically because Hyper-V's `winnat` service grabbed it at boot to hand out to WSL2 / Docker Desktop / Windows Sandbox. `netstat` shows nothing because no process owns the port; the range is just excluded.

Confirmation:

```
netsh interface ipv4 show excludedportrange protocol=tcp
```

[...]

PR #799, opened 2026-07-21
Windows startup walk: trust healthy netstat answers, cache netsh per session

dsarno:
## Summary

The deferred startup walk on Windows measured **7.2–7.4s**, dominated by powershell.exe spawns (~1180ms each). Root cause: an empty per-port parse of `netstat -ano` was treated as a possible parse failure and "confirmed" through the `Get-NetTCPConnection` PowerShell fallback — but for a free port, empty **is** the answer, so every free-port probe burned ~1.2s (~40x the netstat cost) to find nothing. The in-use scrape had the same trap, with its fallback spawns hidden under the `netstat` trace counter.

- **`port_resolver.gd`**: new `windows_netstat_dump_parseable` health check — a dump with at least one parseable TCP row (locale-independent: the proto column is never localized; `netstat -an`-shaped output without a PID column is rejected) proves netstat works, so its empty per-port answer is trusted and the PowerShell probe is skipped. The fallback still fires when netstat is absent or unparseable (stripped installs, locale-odd shapes).
- **`is_port_in_use_via_scrape`** now takes the same optional trace hook as `find_all_pids_on_port`, so a genuine PowerShell fallback is counted as `powershell` instead of hiding under `netstat`; the `plugin.gd` shim delegates counting to the scraper that actually ran.
- **`windows_port_reservation.gd`**: the netsh excludedportrange cache is now session-lifetime instead of a 2s TTL — winnat sets the table at boot, a netsh spawn costs ~250ms, and the TTL could even expire mid-walk when server-command discovery ran long between the two netsh consumers. Staleness risk is bounded to pre-#146 behavior for the rest of the session, and only if a spawn happens after a mid-session winnat change (e.g. Docker starting).

Startup-trace contract (`godot_ai/log_startup_timing`) is unchanged: same counter names, same `done` line format.

## Measurements (Windows 11, same machine/procedure as baseline, 3 boots each, spawn path)

| metric | baseline (711ecdd) | this PR |
|---|---|---|
| walk `total_ms` | 7200–7400 | **1525 / 1321 / 1325** |
| `powershell` counter | 2 | **0** |
| `netstat` counter | 2 | 2 |
| `netsh` counter | 1 | 1 |

## Testing

- Full GDScript suite via `test_run`: **1931 passed / 0 failed / 19 skipped** (1922 baseline + 9 new tests)
- New tests: netstat dump-health matrix (realistic / localized / established-only accepted; empty / garbage / header-only / no-PID-column rejected), Windows live smoke proving free-port and held-port probes never spawn PowerShell on a healthy host, session-lifetime netsh cache incl. `is_port_excluded` + `suggest_non_excluded_port` served with zero netsh spawns
- Targeted suites individually: `port_resolver` 18/0, `windows_port_reservation` 13/0, `server_lifecycle` 40/0, `plugin_lifecycle` 83/0
- `pytest`: 1429 passed / 13 skipped; `ruff check` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows port detection by using reliable `netstat` results before falling back to PowerShell.
  * Improved handling of localized and malformed Windows network output.
  * Reduced repeated Windows port-exclusion queries by caching results for the editor session.

* **Tests**
  * Added coverage for Windows port detection, fallback behavior, localized output, and session-level caching.

* **Documentation**
  * Updated the Tier-1 backlog to mark the Windows port-occupancy item as resolved.


dsarno:
@coderabbitai review

PR #802, opened 2026-07-21
Add opt-in keep_server_on_exit: leave the managed server running on editor exit

### LINKED ISSUE #800: Option to keep the MCP server running after the editor closes
Right now the server dies whenever the editor closes, because the plugin kills it in `_exit_tree`. That's a problem on the client side: I use Claude Code over HTTP, and once the server is gone the connection gets marked as failed and I have to reconnect manually with `/mcp`. Same thing when I start Claude Code before opening Godot — wrong order, no connection.

I worked around it with a systemd user service that runs the server with `Restart=always`. Adoption handles it fine (the plugin picks up the running server on start, #758/#774), but it feels like fighting the plugin: it tries to kill the server on exit, systemd brings it back, and I have to keep the version pin in the service file in sync with the addon by hand.

Would you accept a setting like `godot_ai/keep_server_on_exit`, off by default? When enabled, `_exit_tree` would only disconnect the websocket and leave the process running, and keep the managed-server record so the next editor session adopts it again. The reaper (#497) would need to skip servers with that flag, otherwise it kills them anyway.

I get that killing the server on exit is intentional (that's what the reaper exists for), so this would stay purely opt-in. Version upgrades should be fine too: a newer plugin finding the old server on the port already hits the existing version-mismatch / restart flow.

Can send a PR if you're open to it.

yigitcankzl:
Closes #800.

Adds an opt-in `godot_ai/keep_server_on_exit` EditorSetting (default off, registered next to the port settings). With it enabled:

- `_exit_tree` calls a new `detach_server()` instead of `_stop_server()`: async startup is cancelled and the watch stops like a normal teardown, but the process, managed-server record and pid-file are left alone. The next editor session picks the server up through the existing adoption walk, and HTTP MCP clients (Claude Code etc.) never lose the endpoint.
- Spawns skip `GODOT_AI_OWNER_PID` and stage `GODOT_AI_NO_IDLE_EXIT=1` (the escape hatch that already exists in `orphan_reaper.py`), so neither the owner-PID watchdog nor the idle backstop reaps a server that sits at zero sessions between editor runs on purpose. `NO_IDLE_EXIT` is only unset after spawn when we set it, so a user's own env value isn't stripped.

Explicit stops are untouched: dock Restart and `prepare_for_update_reload()` still go through `stop_server()` and kill, so restarts and self-update keep working. Version upgrades also keep working — a newer plugin finding the old survivor on the port hits the existing version-mismatch / restart flow. The setting is read at spawn time, so toggling it applies on the next server start.

No server-side changes.

Tests: three new cases in `test_server_lifecycle.gd` (detach kills nothing / preserves record + pid-file, detach bumps the async generation like stop_server, env staging honors the setting). Full suite: 1940/1965 passed, 0 failed, 25 skipped (same skips as before).

Also verified end-to-end on Linux with a headless editor: enable setting → clean quit via `editor_manage(op=quit)` → server stays up on 8000 with `NO_IDLE_EXIT=1` and no `OWNER_PID` in its env → second editor launch adopts the same PID.

## Summary by CodeRabbit

* **New Features**
  * Added a “keep server on exit” option to preserve the managed MCP server when the editor closes.
  * The next editor session can adopt and reconnect to the preserved server reliably.
  * Added behavior so the keep-alive mode is applied based on how the server was launched, not the current toggle state.
* **Documentation**
  * Updated teardown and session-adoption details, including explicit stop/restart behavior and when preservation occurs.
* **Tests**
  * Expanded end-to-end coverage for preservation, detach/teardown behavior, adoption, and keep-alive spawn environment handling.


dsarno (changes requested):
Solid feature work for a first contribution. What I verified and like:

- `detach_server()` mirrors `stop_server()`'s head exactly (async-generation bump, watch stop, settle STOPPED) while skipping kill and record cleanup — and the tests lock precisely those contracts, including the stub setup where the server is alive, branded, and listening (i.e. everything `stop_server` *would* kill).
- Env staging follows the established set-spawn-unset pattern, and `GODOT_AI_NO_IDLE_EXIT` is only unset when the plugin set it, so a user's own env value survives.
- Adoption of the survivor works: the WS auth token is recovered from the preserved record (`plugin.gd:260`), and a survivor adopted as *managed* means disabling the setting and exiting later kills it — the feature self-corrects rather than leaking servers.

Changes requested:

1. **Enable-mid-session breaks the feature's promise and manufactures a stale record.** The reaper env vars are staged at *spawn* time, but `_exit_tree` reads the setting *live* (`plugin.gd:560`). Sequence: server running (spawned with `GODOT_AI_OWNER_PID` set, no `NO_IDLE_EXIT`) → user enables `keep_server_on_exit` → quits. `detach_server()` preserves the process, record, and pid-file — but the owner-PID watchdog reaps the server moments after the editor dies. Net result: the user opted into keep-alive, the server dies anyway, and a record + pid-file pointing at a dead PID is left behind — exactly the stale-record scenario #774 exists to defuse (it heals via conservative external adoption, but only after the next session walks into it). Fix: record at spawn whether *this* server was launched keep-alive (a flag next to `_server_pid` on the lifecycle manager) and gate `_exit_tree`'s branch on that flag instead of the live setting. That also makes the docs sentence "the setting is read at spawn time" true for the whole feature, not just the env half. A dock hint ("takes effect after the next server start / Restart") would close the UX loop.

2. **Add a test for the teardown routing** once the spawn-time flag exists: flag set → `detach_server` path (nothing killed, record preserved); flag clear with setting enabled → `stop_server` path. Today the enable-mid-session case has no coverage, which is exactly where the bug in (1) lives.

3. **Make the test's setting restore exact** (`test_server_lifecycle.gd`, `test_keep_alive_env_staging_honors_setting`): `es.set_setting(SETTING_KEEP_SERVER_ON_EXIT, saved)` passes `null` when the setting didn't previously exist, which erases rather than restores. It happens to be papered over by `ensure_settings_registered()`, but the restore should be explicit: track whether the setting existed and either restore the value or erase it deliberately, so the test doesn't depend on re-registration to clean up.

---

yigitcankzl:
Thanks for the review — the mid-session-enable catch was right, and I went with the fix shape you suggested, plus one extension it turned out to need:

1. The lifecycle manager now carries a `_server_keep_alive` flag set at spawn (from whether the env opt-outs were actually staged), and `_exit_tree` routes through a new `teardown_for_editor_exit()` that only looks at that flag — never the live setting. Enabling mid-session now kills on exit as before, no stale record. I also persisted the flag into the managed-server record (`managed_server_keep_alive`) and recover it in `adopt_compatible_server`'s managed branch. Without that the feature was only one exit deep: session 2 adopting the survivor would have killed it on its own exit even with the setting still on. External adoption resets the flag — we can't know a foreign occupant's launch env, so kill-on-exit is the conservative answer. Also took the dock hint: when the live setting differs from the running server's flag, the status line appends "keep-server-on-exit applies after Restart" (flag exposed through `get_status_dict`, so the dock stays on the existing data path).

2. Both routing tests added: flag set + setting OFF → detach (nothing killed, record preserved); flag clear + setting ON → stop path kills and finalizes. Plus two adoption tests for the record recovery (managed recovers the flag, external resets it).

3. Restore is explicit now: a save/restore helper pair tracks whether the setting existed and erases it when it didn't, instead of leaning on re-registration.

Verified on Linux / Godot 4.7: full suite: 1944/1969 passed, 0 failed, 25 skipped (same skips as before). Re-ran the headless e2e, now three sessions deep: spawn keep-alive → quit → server survives (record shows `keep_alive = true`) → second editor adopts same PID → quit → survives again (flag recovered from the record) → third editor adopts the same PID.

dsarno (approved):
94ac9bb resolves all three review points, and goes beyond them in the right ways:

1. **Spawn-time flag routing** — `_server_keep_alive` is set at spawn (both `_start_server_impl` and `respawn_with_refresh`), persisted in the managed-server record (`managed_server_keep_alive`), recovered on managed adoption, conservatively cleared on external adoption and explicit stop, and `_exit_tree` now routes through `teardown_for_editor_exit()` on that flag — never the live setting. The enable-mid-session trap (detached server reaped by the owner-PID watchdog + stale record, the #774 scenario) is closed, and persisting the flag in the record means a keep-alive survivor adopted by a later session detaches again on *that* session's exit — a case my review didn't even ask for. The dock hint ("keep-server-on-exit applies after Restart") honestly covers both toggle directions.
2. **Routing tests** — both directions covered (flag set + setting OFF → detach, nothing killed, record preserved; flag clear + setting ON → kill + finalize), plus adoption recovery from the record and the external-adoption reset.
3. **Exact save/restore** — `_save_keep_setting`/`_restore_keep_setting` track existence and erase rather than restore-as-null.

Verified the stub's `managed_record` path backs the adoption test and that `get_status_dict`'s new `keep_alive` key flows to the dock via `get_server_status`. Docs updated to match the actual semantics. Approving — pending the CI run (workflows are awaiting first-time-contributor approval).

---

PR #813, opened 2026-07-22
fix(clients): configure Kimi Code via mcp.json instead of a nonexistent CLI verb

R0LM0:
## Summary
- Kimi Code has no `mcp` CLI subcommand. `kimi_code.gd` runs `kimi mcp add --transport http {name} {url}` on Configure, which always fails with `error: unknown option '--transport'` — reproduced live against Kimi Code CLI v0.28.1, where `kimi mcp --help` falls through to the root `--help` (no `mcp` command listed at all).
- Per the official docs (https://moonshotai.github.io/kimi-code/en/customization/mcp), Kimi Code manages MCP servers through `~/.kimi-code/mcp.json` (`mcpServers` map, HTTP entries via `url` + optional `transport`), not a CLI verb.
- Switches `kimi_code.gd` from `config_type = "cli"` to `config_type = "json"`, pointing at `~/.kimi-code/mcp.json` with `server_key_path = ["mcpServers"]` and `entry_extra_fields = {"transport": "http"}` — the same mechanism already used by `zed.gd` and the JSON fallback in `claude_code.gd`.
- `kimi_code` was the only descriptor in the registry with `config_type == "cli"` and no JSON fallback (a "pure CLI" client). Two dock/client tests hardcoded it as that example:
  - `test_config_path_facade_returns_empty_for_pure_cli_client` now builds a synthetic pure-CLI probe instead of depending on a specific real client's shape.
  - `test_client_rows_show_config_file_buttons_only_for_file_clients` now discovers a pure-CLI example dynamically and skips with an explicit message if the registry currently has none, instead of hardcoding `kimi_code`.
- Added `test_kimi_code_client_json_descriptor`, pinning the new descriptor shape as a regression guard (mirrors the existing `test_grok_client_toml_descriptor` / `test_windsurf_rebrand_to_devin_desktop` pattern).

## Test plan
- [x] `pytest -v` — 1463 passed, 13 skipped, 0 failed
- [x] `ruff check src/ tests/` — all checks passed
- [x] Godot-side `test_run` (clients + dock suites) — not run locally (would require a second Godot editor instance bound to port 8000, conflicting with an already-running dev session on this machine); relying on CI to exercise `test_clients.gd` / `test_dock.gd`
- [x] Manually verified end-to-end: added a `godot-ai` entry to `~/.kimi-code/mcp.json` by hand, confirmed Kimi Code CLI connects to it (`MCP server "godot-ai" connected · 43 tools (http)`), then applied this same descriptor change and confirmed `client_manage(op="status")` reports `kimi_code` as `configured` after Configure

## Summary by CodeRabbit

* **New Features**
  * Updated Kimi Code integration to rely on a local JSON config file (`~/.kimi-code/mcp.json`).
  * Supports configuring HTTP transport and locating MCP servers via the JSON `mcpServers` section.

* **Bug Fixes**
  * Improved behavior for command-line-only clients so configuration-file controls appear only when a JSON config is available.
  * Updated related validation to match the current client configuration semantics.

### DIFF
--- a/plugin/addons/godot_ai/clients/kimi_code.gd
+++ b/plugin/addons/godot_ai/clients/kimi_code.gd
@@ -5,10 +5,12 @@ extends McpClient
 func _init() -> void:
 	id = "kimi_code"
 	display_name = "Kimi Code"
-	config_type = "cli"
-	cli_names = PackedStringArray(["kimi", "kimi.exe"] if OS.get_name() == "Windows" else ["kimi"])
-	cli_register_template = PackedStringArray(
-		["mcp", "add", "--transport", "http", "{name}", "{url}"]
-	)
-	cli_unregister_template = PackedStringArray(["mcp", "remove", "{name}"])
-	cli_status_args = PackedStringArray(["mcp", "list"])
+	## Kimi Code has no `mcp` CLI subcommand (verified against v0.28.1 —
+	## `kimi mcp` falls through to the root --help, and the docs at
+	## moonshotai.github.io/kimi-code/en/customization/mcp confirm servers are
+	## managed via ~/.kimi-code/mcp.json, not a CLI verb). JSON is therefore
+	## the only working config method, not a fallback.
+	config_type = "json"
+	path_template = {"unix": "~/.kimi-code/mcp.json", "windows": "~/.kimi-code/mcp.json"}
+	server_key_path = PackedStringArray(["mcpServers"])
+	entry_extra_fields = {"transport": "http"}
--- a/test_project/tests/test_clients.gd
+++ b/test_project/tests/test_clients.gd
@@ -171,10 +171,47 @@ func test_config_path_facade_handles_unknown_and_file_clients() -> void:
 
 
 func test_config_path_facade_returns_empty_for_pure_cli_client() -> void:
+	## Decoupled from any specific registered client (#kimi-mcp-json): a "pure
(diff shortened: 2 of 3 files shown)


dsarno (approved):
CI approved and the full matrix is green (23/23). Verified the premise independently: on main, `kimi_code.gd` runs `kimi mcp add --transport http ...`, and Kimi Code v0.28.1 has no `mcp` subcommand — the official docs manage servers via `~/.kimi-code/mcp.json`. The descriptor move to `config_type = "json"` mirrors the zed.gd pattern, and the `"unix"` path_template key is the documented darwin+linux shorthand in `_base.gd`.

On the test changes: decoupling the two pure-CLI tests from a specific registry entry is the right call, and the explicit skip-with-reason when no pure-CLI client exists keeps the degradation loud rather than silent. The new `test_kimi_code_client_json_descriptor` regression guard pins the descriptor shape. Thanks for the manual end-to-end verification as well. LGTM.

dsarno:
Merged — thank you @R0LM0, and congrats on your first contribution to godot-ai! 🎉 This was a model first PR: you reproduced the failure against the real CLI, cited the upstream docs, reused the existing descriptor pattern instead of inventing a new one, and left the test suite more robust than you found it.

If you'd like a shoutout, come drop your handle in our Discord: https://discord.com/invite/FDZ5fr2QkP

Issue #816, opened 2026-07-22
Make MCP availability startup-order independent with a client-owned `godot-ai attach` bridge


Clubhouse1661:
## Problem

Godot AI currently configures most MCP clients with a Streamable HTTP URL such as:

```text
http://127.0.0.1:8000/mcp
```

The HTTP/WS backend is normally started and stopped by the Godot editor plugin. Many MCP clients initialize configured servers and discover their tools only during client or task startup.

This creates two failure modes:

1. **Client starts before Godot**

   The HTTP endpoint is unavailable when the client initializes. The client records startup failure and may not retry, so Godot AI tools remain unavailable after Godot starts.

2. **Backend restarts during a client session**

   Plugin reload, editor restart, dock restart, or self-update can replace the backend and invalidate Streamable HTTP session IDs. Clients that do not reconnect and reinitialize remain broken until restarted.

`notifications/tools/list_changed` cannot solve the cold-start case because no MCP session exists when initial connection fails.

## Desired behavior

For every supported local MCP client that can launch a stdio process:

- Godot AI tools are discoverable regardless of whether the MCP client or Godot starts first.
- Starting or reconnecting Godot makes existing tools usable without restarting the client.
- Multiple clients share one HTTP/WS backend.
- Same-version editor and plugin restarts recover transparently.
- Mutating tool calls are never duplicated during transport recovery.
- Unknown processes occupying configured ports are never terminated.
- Package or tool-schema upgrades produce an explicit reconfiguration/restart diagnostic rather than a respawn loop.

Clients that only support URL-based MCP servers should retain URL mode with its documented limitations.

## Root cause

The MCP endpoint and the MCP client have incompatible process ownership:

```text
Current:

MCP client ──HTTP──> backend owned by Godot
                            ↑
                         Godot plugin
```

With URL transport, the client assumes the endpoint already exists. The plugin cannot notify a client that never established a session.

The tool catalog itself is not dependent on an active Godot editor. Python registers the tools when the backend starts; editor-dependent calls already return structured “no active editor session” errors. The availability problem is therefore at the process and transport lifecycle boundary.

## Proposed architecture

Add a standard stdio MCP entry point:

```text
godot-ai attach
```

Each local client launches its own lightweight bridge, while all bridges and Godot editors share one externally owned backend:

```text
Codex ─────────┐
Claude Code ───┼── stdio bridge per client ── HTTP ── shared backend
Cursor ────────┘                                      │
                                                      │ WebSocket
[...]

PR #821, opened 2026-07-23
Add client-owned `godot-ai attach` bridge

Clubhouse1661:
## Summary

Implements PR 1 of #816: a client-owned `godot-ai attach` stdio bridge that starts or adopts one shared Godot AI HTTP/WebSocket backend.

This establishes the Python-side lifecycle needed to make MCP availability independent of whether the Godot editor or MCP client starts first. Client configuration rollout is intentionally deferred to a follow-up PR.

```text
MCP client ── stdio ── godot-ai attach
                            │
                            │ fresh HTTP MCP session per operation
                            ▼
                 shared backend on :8000
                            │
                            │ WebSocket on :9500
                            ▼
                     Godot editor plugin
```

### Client-owned attach entry point

- Adds `godot-ai attach`, dispatched before the existing root CLI parser.
- Preserves the existing `--transport`, `--port`, `--ws-port`, `--pid-file`, and related root CLI behavior.
- Accepts a minimal attach-specific surface:
  - `--port`
  - `--ws-port`
  - `--exclude-domains`

### Singleton backend startup and adoption

- Uses a per-port, cross-process advisory lock:
  - `msvcrt.locking` on Windows
  - `fcntl.flock` on POSIX
- Holds the lock until the backend reports healthy or startup reaches a bounded failure.
- Adopts an already-running compatible Godot AI backend.
- Starts a detached backend through the bridge’s current Python environment when the ports are free.
- Prevents the backend from inheriting the bridge’s stdin/stdout/stderr.
- Never kills or replaces a foreign or incompatible process.
- Reports explicit diagnostics for:
  - foreign port occupants
  - version/protocol/port/exclusion mismatches
  - startup failures and crash loops

### Conservative stdio proxy

- Exposes the backend’s MCP surface through a standard stdio server.
- Opens a fresh downstream Streamable HTTP MCP session for each operation.
- Uses bounded connect/write/pool timeouts and no post-dispatch read deadline.
- Preserves backend MCP errors and structured data unchanged, including:
  - `EDITOR_NOT_READY`
  - `data.sub_code = EDITOR_TEST_RUNNING`
- Retries `tools/call` only when connection failure proves that no request bytes were dispatched.
- Treats both `ConnectError` and `ConnectTimeout` as proven pre-dispatch failures.
- Never replays a mutating call after an ambiguous reset, timeout, replacement, or disconnect.
- Returns `TRANSPORT_OUTCOME_UNKNOWN` when a dispatched call may have completed.
- Names `test_manage(op="results_get")` as the recovery probe for interrupted `test_run` calls.

### In-flight backend monitoring

- Monitors backend identity without imposing a tool-call deadline.
- Uses direct status probes only:
  - no advisory lock
  - no backend spawning
  - no lifecycle side effects
- Uses a 5-second status-probe timeout.
- Requires three consecutive inconclusive probes before declaring the backend unavailable.
- Treats a positively observed instance-ID change as immediate replacement evidence.
- A transient slow or failed status probe therefore cannot cancel a healthy in-flight call.

### Instance-bound leases and attach-owned reaping

Adds loopback-guarded endpoints:

- `POST /godot-ai/lease/register`
- `POST /godot-ai/lease/heartbeat`
- `POST /godot-ai/lease/release`

Leases are:

- in-memory
- tied to the backend instance nonce
- scheduled with monotonic durations
- re-registered automatically after backend replacement
- expired automatically when a bridge crashes

Attach-owned backends exit only after:

```text
no connected Godot editors
AND no unexpired bridge leases
AND the configured grace period has elapsed
```

A boot grace prevents a newly spawned backend from exiting before its first bridge registers.

### Backend identity and compatibility status

Extends `/godot-ai/status` with:

- `instance_id`
- `owner_type`
- `attach_protocol_version`
- `tool_catalog_hash`

All existing status fields retain their names, types, and meanings.

Compatibility gates on package version, attach protocol, ports, and excluded domains. The catalog hash is diagnostic only and never authorizes process termination.

### Agent-facing recovery errors

Adds these codes to both Python and GDScript registries:

- `TRANSPORT_OUTCOME_UNKNOWN`
- `NEW_CLIENT_SESSION_REQUIRED`

Structured errors follow the existing convention:

- `sub_code`
- `retryable`
- `hint`

No-editor errors now tell the agent to open this project in Godot with the plugin enabled, wait for it to connect, and retry the same call. They explicitly state that restarting the MCP client is not required.

Crash-loop errors direct the user or agent to `backend-<HTTP-port>.log` in the Godot AI runtime directory rather than incorrectly suggesting that the editor is missing.

### FastMCP compatibility

The declared FastMCP 3.0 floor exposed two existing compatibility problems:

- module locations differ across the supported FastMCP range
- postponed annotations copied into telemetry wrappers could fail schema construction

This PR adds a narrow compatibility shim, resolves wrapper annotations in the original function namespace, and adds a dedicated FastMCP 3.0 CI job.

### Documentation

`AGENTS.md` now documents:

- external adoption does not transfer process ownership
- attach lock/log runtime-directory resolution
- `GODOT_AI_RUNTIME_DIR`
- Windows, XDG, and temporary-directory fallbacks
- the follow-up requirement that generated client configs pass the same `--exclude-domains` selection as the plugin

## Safety invariants

This PR preserves the following invariants:

1. The attach bridge never kills any process.
2. It never replays `tools/call` after possible dispatch.
3. It imposes no per-call deadline below the server-owned tool budget.
4. An attach-spawned backend never inherits bridge stdio.
5. An attach-spawned backend is never written to Godot’s managed-server record.
6. The startup lock is held until health or a bounded failure diagnostic.
7. Existing root CLI flags, status-field semantics, and error-code values remain unchanged.
8. Backend errors, including `EDITOR_TEST_RUNNING`, pass through unchanged.
9. Status identity fields are diagnostic and never constitute kill authorization.

## Validation

Python:

- `1501 passed`
- `13 skipped`
- Ruff passed
- `git diff --check` passed

Focused attach regression/integration suite:

- `15 passed`

Live Godot 4.7 validation:

- worktree backend started successfully
- editor plugin connected and adopted it
- `scene_open("res://main.tscn")` succeeded
- `test_run` completed through MCP:
  - `1975 passed`
  - `0 failed`
  - `32 skipped`
  - 64 suites
  - 2007 total tests

Additional checks:

- worktree verification passed
- GDScript import/parse validation passed
- backend stdout isolation tested
- real backend replacement recovery tested
- real post-dispatch disconnect tested with a backend-side call counter
- long-call passthrough tested without a bridge-side read deadline
- multi-bridge/version-skew behavior tested
- lease expiry and reap-after-release tested

## Coordination with #817

This branch is based on `main` after #817.

It uses the shared 300-second `test_run` budget, preserves `EDITOR_TEST_RUNNING` pass-through, and treats `test_run` as the canonical no-replay case for ambiguous transport failures.

The only GDScript implementation change is the addition of the two error-code registry constants.

## Deliberately out of scope

This PR does not change:

- client descriptors
- TOML/JSON/YAML config strategies
- client migration behavior
- dock UI
- the plugin’s managed-server record
- `server_lifecycle.gd`

A follow-up PR will render, verify, and migrate `godot-ai attach` command configurations across the supported client strategies. Until then, existing configured clients remain in their current URL or proxy mode.

## Summary by CodeRabbit

* **New Features**
  * Added `godot-ai attach` client-owned stdio bridge with shared-backend “start or adopt”, lease management, and attach-aware lifecycle coordination.
  * Extended `/godot-ai/status` with instance/compatibility metadata and added lease endpoints (register/heartbeat/release).
  * Included FastMCP 3.x compatibility for consistent tool error/result behavior.
* **Bug Fixes**
  * Improved attach proxy recovery to avoid unsafe tool-call replay and return structured errors for unknown outcomes.
  * Enhanced idle shutdown logic to account for attach/bridge leases.
* **Documentation**
  * Updated attach lifecycle guidance, runtime/diagnostic log locations, and attach protocol compatibility expectations.
* **Tests / CI**
  * Added unit and integration coverage for attach CLI/ensure/lease/proxy/status and a FastMCP 3.0 compatibility check in CI.


dsarno:
@Clubhouse1661 Great stuff.  Can you run the stormtest suite on this before merge?

Clubhouse1661:
> @Clubhouse1661 Great stuff. Can you run the stormtest suite on this before merge?

@dsarno Absolutely. I’ll add targeted unit tests to close the Codecov gaps, then run the full stormtest suite and share the results here.

Clubhouse1661:
Ran the stormtest suite on Windows with Godot 4.7 and an externally launched worktree server.

Results:

- Isolated reload mode: 10/10 reloads survived, averaging 2.1s recovery.
- Concurrent load without reloads: 1,519/1,539 calls succeeded at 33.1 calls/sec, with no reconnects and the editor remaining alive.
- Controlled default concurrent matrix, restarting Godot and restoring the project between every run:
  - PR `2a3ecd0`: 3/3 runs passed, 6/6 reloads survived, 3,897/4,008 calls succeeded.
  - Post-#817 main `4727ecf`: 3/3 runs passed, 6/6 reloads survived, 4,081/4,189 calls succeeded.

No Godot crashes occurred during the controlled comparison matrix. The PR’s results are consistent with post-#817 main.

PR #827, opened 2026-07-24
feat(clients): configure Codex through the attach bridge

Clubhouse1661:
## Summary

Implements the first client-config rollout slice from #816: generic command-shape support plus migration of Codex from Streamable HTTP URL mode to the client-owned `godot-ai attach` stdio bridge.

This gives Codex startup-order-independent tool discovery:

- Codex can start before Godot.
- The MCP tools are available immediately.
- Calls made without an editor return actionable guidance.
- Opening Godot later makes the same tools usable without restarting Codex.
- Same-version backend/editor restarts recover through the attach lifecycle added in #821.

This PR deliberately flips only Codex. Other client descriptors remain in their existing URL or bridge modes for later strategy-by-strategy rollout.

### Generic command-shape support

Adds descriptor vocabulary for clients that launch a local MCP command:

- `CommandShape`
- command/argument rendering
- initial generated fields
- user-preserved fields
- timeout fields
- legacy launch keys
- launch-shape verification

Descriptors cannot activate both the legacy `UvxBridge` path and the new command-shape path.

### Worker-safe attach launch resolution

Adds a three-tier resolver:

1. Development checkout venv
2. Exact-version uvx package
3. Matching system `godot-ai` installation

The resolver captures EditorSettings-derived values on the main thread before client actions move to workers. Generated arguments include the configured:

- HTTP port
- WebSocket port
- excluded domains
- plugin/package version

The uvx tier uses:

```text
uvx --link-mode copy --from godot-ai==VERSION godot-ai attach ...
```

System installations must report the exact bundled package version before they are accepted. If no compatible launcher exists, Configure returns an error without changing the client’s config file.

### Codex command migration

Codex now receives a standard stdio MCP entry instead of a URL entry:

```toml
[mcp_servers."godot-ai"]
command = "..."
args = [
  "...",
  "attach",
  "--port", "8000",
  "--ws-port", "9500",
]
enabled = true
startup_timeout_sec = 60
tool_timeout_sec = 360
```

The 360-second tool timeout leaves margin above `test_run`’s 300-second server budget from #817.

Existing URL entries and stale launch/version/port/exclusion shapes report `CONFIGURED_MISMATCH`, reusing the existing four-value client-status contract.

### Span-aware TOML migration

The TOML strategy now verifies launch values semantically and migrates the Codex entry without corrupting user-owned configuration.

Migration:

- removes the obsolete `url` key;
- writes `command` and `args`;
- preserves `enabled`;
- preserves user-modified startup and tool timeouts;
- preserves approval settings and unknown user fields;
- preserves inline and multiline arrays such as `enabled_tools`;
- preserves multiline strings and comments to the extent allowed by the existing merge model;
- renames legacy `[mcp_servers.godot_ai]` descendant subtables to the canonical `"godot-ai"` server ID;
- preserves per-tool subtables during that rename;
- leaves the file byte-identical when launch discovery fails.

Rendered samples are parsed by Python’s real `tomllib` in the cross-language fixture test.

### Windows consoleless launch behavior

Live acceptance found two independent Windows console-subsystem problems.

#### Client bridge

Launching `python.exe`, `uvx.exe`, or `godot-ai.exe` directly from Codex could open Windows Terminal.

Generated Windows entries now:

- use the selected dev venv’s sibling `pythonw.exe` directly; or
- use a `pythonw.exe` bootstrap for uvx/system commands;
- explicitly forward stdin, stdout, and stderr;
- start the underlying console command with `CREATE_NO_WINDOW`.

If no consoleless interpreter can be resolved at Configure time, the client configuration fails actionably instead of writing a command known to open a terminal.

Non-Windows launch shapes are unchanged.

#### Attach-owned backend

A full uvx cold-start test found that `spawn_backend()` still used the wheel environment’s `sys.executable` (`python.exe`). Despite the existing creation flags, the venv redirector opened Windows Terminal when it started the real backend interpreter.

The backend now selects `sys.executable`’s sibling `pythonw.exe` on Windows when present. This preserves environment identity: the bridge and backend still import the exact same `godot_ai` installation.

If `pythonw.exe` is unavailable, backend spawning falls back to the original executable rather than refusing to start.

The Windows flags are now:

```text
CREATE_NO_WINDOW | CREATE_NEW_PROCESS_GROUP
```

`DETACHED_PROCESS` was removed because Win32 documents that it causes `CREATE_NO_WINDOW` to be ignored. The unit test explicitly pins both desired flags and the absence of `DETACHED_PROCESS`.

### Dock and documentation

Updates:

- Codex launch-mode/status messaging
- manual command output
- README manual configuration
- port-change guidance
- contributor documentation for the rollout behavior

URL mode remains documented as an advanced fallback with its existing reconnect limitation.

## Migration and release notes

Existing Codex users must:

1. Run **Configure** for Codex once.
2. Restart Codex once so it loads the new stdio entry.

The Configure-all-mismatched flow can perform the first step.

The uvx entry pins the bundled plugin version. The matching package version must therefore be available on PyPI before releasing the plugin update to non-development users.

Intentional URL-mode users will continue to see `CONFIGURED_MISMATCH`; an opt-out setting is outside this initial rollout slice.

### Python

```text
1599 passed, 19 skipped
```

This includes:

- real TOML parsing of rendered GDScript fixtures;
- attach argument parser validation;
- Windows `pythonw` bootstrap stdio forwarding;
- backend sibling-`pythonw` selection and fallback;
- explicit Windows creation-flag assertions.

### GDScript

Attach client configuration suite:

```text
17 passed, 0 failed
```

General client suite:

```text
137 passed, 7 skipped, 0 failed
```

The broader GDScript suite was also green during implementation before the final Python-only backend hardening changes.

### Dev-venv tier

Tested on Windows with Godot initially closed:

1. Fully stopped Codex, Godot, related bridge/backend processes, and listeners.
2. Started Codex first.
3. No terminal window appeared.
4. Codex discovered all 42 Godot AI tools.
5. `editor_state` returned the expected `PLUGIN_DISCONNECTED` guidance:
   - open the project in Godot with the plugin enabled;
   - retry the same call;
   - restarting the MCP client is not required.
6. Opened the worktree project in Godot.
7. Retried `editor_state` successfully from the same Codex session.

### Local-wheel uvx tier

Built the branch as a local `godot_ai-3.0.6-py3-none-any.whl` and configured the real Codex entry through:

```text
pythonw bootstrap → uvx --from <local-wheel> → godot-ai attach
```

After a complete shutdown ensuring there was no backend to adopt:

1. Relaunched Codex with Godot closed.
2. No terminal window appeared.
3. Codex discovered all 42 tools.
4. `editor_state` returned the expected no-editor recovery guidance.
5. Process-tree inspection confirmed the attach-owned backend itself was:

```text
...\uv\cache\archive-v0\...\Scripts\pythonw.exe
    -m godot_ai
    --transport streamable-http
    ...
```

The complete observed Windows path was therefore:

```text
Codex
  → pythonw stdio bootstrap
  → uvx
  → wheel-installed attach bridge
  → wheel-environment pythonw backend
```

This validates both tool discovery and absence of visible terminal windows across the full non-development launch chain.

## Invariants

- [x] Codex receives a standard stdio MCP server entry.
- [x] Godot does not need to be running when Codex starts.
- [x] Missing-editor errors tell the agent to open Godot and retry without restarting Codex.
- [x] Generated ports and excluded domains match the plugin’s backend settings.
- [x] The uvx package version matches the bundled plugin version.
- [x] Existing user-controlled Codex fields and descendant subtables survive migration.
- [x] Legacy URL and stale command entries report `CONFIGURED_MISMATCH`.
- [x] Failed launcher discovery never writes a partial configuration.
- [x] `test_run` has sufficient client-side timeout margin.
- [x] Windows bridge and backend processes do not open terminal windows.
- [x] The attach-owned backend remains in the same Python environment as its bridge.
- [x] Non-Windows launch behavior is unchanged.
- [x] Other client descriptors are not flipped in this PR.

## Follow-up

Later #816 rollout slices can reuse the generic command-shape resolver for:

- JSON clients
- CLI-managed clients
- Grok Build via the TOML strategy
- Hermes via YAML
- remaining client descriptors in smoke-tested batches

## Summary by CodeRabbit

* **New Features**
  * Codex client setup now supports attach-based, command-driven TOML so tools remain available across same-version editor restarts.
  * Launcher discovery is expanded across dev venv, exact-version uvx, and system installs, with stricter “fail closed” mismatch handling and clear **Configure all** repair guidance.
  * Windows attach uses a consoleless `pythonw.exe` bootstrap for hidden stdio.
* **Documentation**
  * Updated Quick Start, manual configuration examples, port-conflict guidance, and attach configuration notes, including an advanced URL fallback.
* **Tests**
  * Added/expanded end-to-end and unit coverage for attach config parsing/rendering, semantic mismatch detection, and Windows stdio/launch behavior.


Clubhouse1661:
@dsarno — requesting your review on this one, both as an implementation review and as sign-off on the rollout precedent. This is the first #816 client-config slice (steps 5–6): generic command-shape support and the Codex migration from URL mode to the attach bridge. Because generated `uvx` configs pin the shipped package version, this also determines the first release in which Codex users get startup-order independence.

The implementation has been through several review rounds and live Windows acceptance across both discovery tiers. The three decisions I especially need from you are:

1. **Intentional URL mode.** Existing `url =` entries report `CONFIGURED_MISMATCH`, so the dock offers migration. A user who intentionally remains on URL mode will therefore keep an amber status. Is that acceptable for this rollout, or should we require an explicit “keep URL mode” opt-out? The current implementation adds neither an opt-out nor a fifth status enum.

2. **The migration contract.** This establishes the pattern subsequent clients will copy: span-aware TOML merging; preservation of user fields, timeouts, subtables, and comments; removal of the legacy `url` key; and canonical renaming of legacy alias sections and their descendant subtables. Launch-affecting drift—version pin, ports, exclusions, or command path/shape—reports `CONFIGURED_MISMATCH`, with **Configure all** as the repair path. If you want that contract shaped differently, now is the inexpensive time to change it.

3. **Windows launch behavior.** Live acceptance exposed console-subsystem issues at two levels. The client launch now uses `pythonw.exe` directly for a dev venv, or a stdio-forwarding `pythonw` bootstrap with `CREATE_NO_WINDOW` for `uvx` and system installs. The attach-owned backend likewise selects the environment’s sibling `pythonw.exe`. The full cold-start chain—Codex → bootstrap → `uvx` → wheel-installed bridge → attach-owned backend—was verified with no visible windows, all 42 tools discovered, and the expected `PLUGIN_DISCONNECTED` recovery guidance. Is this acceptable as the reference behavior for the later JSON and CLI migrations?

Two additional invariants: generated configs include the plugin’s current `--exclude-domains` selection, with incompatible running backends returning `NEW_CLIENT_SESSION_REQUIRED`; and if no verified launcher tier is available, configuration fails without modifying the file.

If these decisions look right, your approval would let me carry this pattern into the Claude Desktop replacement and the remaining client migrations.

dsarno (approved):
Approving all three decisions. Caveat up front: I'm away for the weekend and haven't gone through this line by line. It's been through several review rounds and live acceptance on both discovery tiers, and I trust you to carry it from here.

**1. Intentional URL mode — ship as-is.** No opt-out, no fifth status value. URL mode is strictly worse given the reconnect limitation, and the endpoint of #816 is to retire it as anything but a fallback. An opt-out added now would ossify the mode we're trying to remove, at the cost of a settings key and a branch in all four strategies, for a population that's probably close to zero.

The effort is better spent elsewhere: the amber dot isn't the real harm — **Configure all** silently changing an intentional URL user's transport is. `check_status_details` returns an empty `error_msg` on every mismatch branch, so the dock can't say why a row went amber. Populating just the legacy-key branch ("URL entry found; Configure replaces `url` with a launch command") addresses the actual objection without touching the four-value contract. If someone genuinely wants to stay on URL mode later, a marker comment in their own config file is cheaper than a setting — the merge model already preserves comments. Not needed now.

**2. Migration contract — right shape, one thing to close before it's copied.** Two parts worth locking in explicitly: preservation is driven by the pinned/initial/removed key sets computed in `render_body` rather than by the descriptor lists, which is what keeps `command_user_fields` documentation-plus-test-pin with no control flow on the descriptor (clear of #229); and discovery failure fails closed with no partial write. Both generalize cleanly to JSON, YAML, and CLI.

The gap is that a mismatch carries no reason. Ten branches, all returning `CONFIGURED_MISMATCH` with an empty message. That's fine for one client and one file; across twelve it turns every "my dot went amber" report into a blind debugging session, and the causes differ in kind — legacy transport, version pin, ports, exclusions, launcher path moved, shape changed. Add a `reason` string to the details dictionary before the pattern goes into the next slice. It doesn't change the status enum or the dock's colors, and retrofitting it into four strategies after twelve descriptors have shipped costs considerably more than adding it to one. This matters more than it looks because status is exact equality against a freshly resolved absolute path, so a Python upgrade or a rebuilt venv is drift indistinguishable from a stale port.

**3. Windows — accept as the reference behavior.** The `DETACHED_PROCESS` removal is correct; Win32 documents `CREATE_NO_WINDOW` being ignored when it's set alongside. Isolation survives regardless, since `CREATE_NO_WINDOW` gives the child its own hidden console rather than inheriting the parent's, and `CREATE_NEW_PROCESS_GROUP` keeps the Ctrl+C boundary. Keep the test pinning both flags and the absence of the third.

Two notes for the later migrations. The inline `-c` bootstrap is the right call — I went looking for a module entry point to replace it and there isn't one available, because the uvx and system tiers resolve a system interpreter that has no `godot_ai` installed — but it does mean that string is now a wire format sitting in user config files, and any whitespace edit to it flips every Windows user to amber. Worth a comment at the constant so nobody tidies it later. Separately, make the fail-open/fail-closed split explicit in AGENTS.md: client-side resolution fails closed (no `pythonw`, Configure errors, file untouched) while backend spawn falls back to `python.exe` and fails open. That's the right split — a visible console beats a dead backend at runtime, a missing config beats a known-bad one at write time — but it's non-obvious enough that the JSON and CLI slices could easily pick the opposite.

None of this blocks the merge. Reason codes are the only item I'd want landed before the next client goes in, and that can be here or in the following PR — your call. The failing Docstring Coverage check is CodeRabbit not understanding GDScript; ignore it.

Thanks for the thoroughness on this one, particularly the live cold-start acceptance across both tiers. The Windows console work is the kind of thing that only surfaces in real use and would have been genuinely painful to find after a release. Go ahead and carry the pattern into the Claude Desktop replacement and the remaining clients.

---

PR #835, opened 2026-07-26
Report non-imported paths distinctly from reimport (#778)

### LINKED ISSUE #778: filesystem_manage.reimport should distinguish non-imported files such as .gd scripts

dsarno:
Closes #778.

## Problem

`filesystem_manage(op="reimport")` called `EditorFileSystem.update_file()` on every existing path and appended all of them to `reimported`. Scripts, scenes and hand-written resources are not imported assets, so a `.gd` entry in that list read as proof that a parse or import had run when nothing had. #765 documented the limitation in the tool description but left the response shape saying otherwise.

## Change

The `.import` sidecar decides the split — it is the editor's own record that a path goes through the import pipeline:

- `reimported` — imported assets only
- `skipped_non_imported` + `skipped_non_imported_count` — everything else
- `skipped_non_imported_hint` — attached **only** when that bucket is non-empty, so the all-assets path this op exists for pays nothing. It names `script_patch`/`script_create` for GDScript diagnostics and `scan` for an asset awaiting its first import.

An extension allow-list was rejected: importers come and go with plugins, so the list would drift out of agreement with the editor it claims to describe.

**Behaviour is unchanged for every path** — `update_file()` still runs on all of them, because refreshing an externally-edited `.tscn`/`.tres` is a real use of this op. This splits the report, not the work.

One documented edge: an asset the editor has never imported (just written, no scan yet) has no sidecar and reports as non-imported. That is accurate at the moment of the call — `update_file()` did not import it either — and the hint names `scan` as the way through.

## Verification

- `pytest` — 1643 passed, 5 skipped
- `script/ci-godot-tests` — 2009/2036 passed, 0 failed; confirmed by name that the four new `test_reimport_*` cases actually executed rather than a stale script passing
- **Live smoke against the running editor**, including a real Godot-imported asset rather than only the fixture — wrote an actual PNG into `test_project/`, let the editor import it, and confirmed Godot writes the `.import` sidecar the classification depends on:

```
paths: ["res://_mcp_smoke_778.png", "res://main.tscn"]
  reimported:            ["res://_mcp_smoke_778.png"]
  skipped_non_imported:  ["res://main.tscn"]
```

Smoke artifacts removed and the project rescanned; `git status` clean apart from the four changed files.

## Tests

GDScript coverage for the mixed imported/non-imported batch the issue asks for, plus the sidecar-path edge (`foo.png.import` is the record, not the resource). The pre-existing `test_reimport_existing_file` asserted a `.txt` counted as reimported — updated, since that assertion encoded the bug. One Python integration test covers the pass-through of the new fields.

---

## Summary by CodeRabbit

- **New Features**
  - Improved filesystem reimport reporting by distinguishing resources that were actually reimported from refreshed entries that lacked import evidence.
  - Added dedicated result lists and counts for **reimported**, **skipped non-imported**, and **not found** outcomes.
  - Added clearer messaging to help interpret cases where a refresh was performed but reimport couldn’t be confirmed.

- **Documentation**
  - Updated reimport operation documentation to reflect the new result categories and the meaning of skipped items and hints.

### DIFF
--- a/plugin/addons/godot_ai/handlers/filesystem_handler.gd
+++ b/plugin/addons/godot_ai/handlers/filesystem_handler.gd
@@ -15,6 +15,11 @@ const ScriptHandler := preload("res://addons/godot_ai/handlers/script_handler.gd
 const _SCAN_START_GRACE_MSEC := 750
 const _SCAN_SETTLE_MAX_MSEC := 28000
 
+## Sidecar the editor writes next to every imported resource. `reimport` reads
+## it to tell imported assets from files that merely have a filesystem entry
+## (see `_is_imported_resource`).
+const IMPORT_SIDECAR_SUFFIX := ".import"
+
 ## Shared single-flight latch for scan_filesystem. `is_scanning()` alone can't
 ## enforce single-flight: `EditorFileSystem.scan()` doesn't flip `is_scanning()`
 ## for a frame or two (hence _SCAN_START_GRACE_MSEC), so a second request landing
@@ -135,6 +140,7 @@ func reimport(params: Dictionary) -> Dictionary:
 			"EditorFileSystem not available", false)
 
 	var reimported: Array[String] = []
+	var skipped_non_imported: Array[String] = []
 	var not_found: Array[String] = []
 
 	for path_variant in paths:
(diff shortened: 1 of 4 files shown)


dsarno:
Thanks — took the first, declining the second.

**Sidecar path assertion — fixed** in `1ba0e1f`. Correct catch: the test asserted `skipped_non_imported_count == 1`, which any path in that bucket would satisfy, so it wasn't actually pinning the sidecar path. Now asserts the array, matching the mixed-batch test. Filesystem suite still 26/26.

**Importer-backed fixture — not adding one**, for two reasons:

1. **It would test Godot, not this handler.** The handler doesn't branch on whether an import actually runs — `update_file()` is called unconditionally on every existing path, and classification reads one thing: does `<path>.import` exist. A real texture fixture exercises Godot's importer, which isn't the code under test.

2. **It would be flaky.** A committed `.png` has no sidecar in a fresh checkout; the editor writes it during its filesystem scan. The suite would then depend on scan timing in CI to decide which bucket the fixture lands in — intermittent failures for no added signal.

The concern behind the nitpick is real, though — the whole change rests on Godot actually writing `<path>.import`, and a synthetic fixture can't confirm that. So I verified it against the live editor before opening the PR rather than asserting it in the suite: wrote a real PNG into `test_project/`, let the editor import it, confirmed Godot created `_mcp_smoke_778.png.import`, then reimported it alongside `main.tscn`:

```
reimported:            ["res://_mcp_smoke_778.png"]
skipped_non_imported:  ["res://main.tscn"]
```

That's in the PR description with the rest of the verification. `Godot tests / Linux (Godot 4.5)` passing also covers the sidecar convention holding at the engine floor, not just on 4.7.

---

Issue #782, opened 2026-07-21
Investigate: name the affected file(s) in the error doorbell hint (weak-agent follow-through)


dsarno:
## Evidence (from #761's two Stage-0 behavioral evals)

Both instruments found that a count-only doorbell fails weaker agents: in the Claude harness, Haiku received the genuine-error doorbell in 12/12 trials and reported the seeded broken file in **0/12** — under both old and new hint text, and even in surviving sessions that read the logs afterward (Sonnet: 10/10). The independent Codex replication converged on the same recommendation: treat survivor-file naming as a separable hypothesis worth a clean targeted test. (Caveat carried over from the eval: the Haiku number is partly confounded by the since-fixed #764.)

## Hypothesis

A hint that names the failing file(s) — *"2 new GDScript errors since your last call, including res://src/legacy_stats.gd"* — converts doorbell deliveries into follow-through for agents that ignore an unnamed count.

## Mechanism constraint (established in #761)

Paths cannot ride the count-only `error_watermark`: `_sync_error_watermark_for_session` sums every unrecognized integer key (`src/godot_ai/transport/websocket.py`), so flat additions double-count on old servers. A **separate optional envelope field** (e.g. `error_paths`, bounded to the N most recent editor-origin error paths, populated plugin-side) is the compat-safe carrier; old servers ignore it, old plugins simply don't send it.

## Scope discipline

This is an investigation issue. Ship nothing until a targeted A/B (the #761 V3 harness is reusable: `script/eval-v3-stage0/`) shows the named hint moves weak-agent follow-through; the implementation itself is small once justified.

Explicitly **not** in scope: #761's Stage-1 classification and Stage-2 settlement machinery — both parked on eval evidence.

---
_Generated by [Claude Code](https://claude.ai/code)_

dsarno:
Status note for whoever picks this up: **the harness this issue tells you to reuse is not in the repo.**

The scope section says "the #761 V3 harness is reusable: `script/eval-v3-stage0/`". That path doesn't exist in `script/` today, and it appears in no commit reachable from `main` — so it was never merged, not deleted later. `script/` currently has no eval tooling at all (the only `eval`-named files in history are `test_game_eval*`, which are `game_eval` tests, unrelated).

That doesn't invalidate the issue — the hypothesis and the mechanism constraint still hold, and both remain the interesting part:

[...]

PR #836, opened 2026-07-26
Document the watermark's fold-unknown-keys contract (#782)

dsarno:
Comment only, no behavior change.

## Why

`_sync_error_watermark_for_session` ends in:

```python
new_total = overlap + sum(deltas.values())
```

Every integer key that isn't explicitly popped (`_WARN_WATERMARK_KEYS`, the two overlapping debugger/game components) lands in that sum — **including keys the receiving server has never heard of**. So a plugin that stamps a new watermark counter against an older server doesn't get it ignored; it inflates that server's "N new errors since your last call" doorbell. There's no allow-list gating this: the existing key sets only route what they recognize, and everything else falls through to the sum.

Nothing at the call site said so. The docstring covered the reset/overlap semantics carefully but not the one rule that bites a future contributor from another process's version.

This constraint was worked out while investigating #782 (naming the failing file in the doorbell hint). That investigation is being closed unstarted — its A/B gate now requires rebuilding an eval harness that was never committed to this repo — so the landmine is recorded here rather than left in a closed issue. It applies to anyone extending the watermark, not just to that proposal.

The comment also names the safe alternative, which the #782 sketch had right: carry genuinely new signal in a separate optional envelope field. Non-int values already fail `_normalized_watermark_int` and get skipped, so that shape degrades safely in both directions.

## Verification

- `pytest tests/unit -k "watermark or websocket"` — 11 passed

---

## Summary by CodeRabbit

* **Documentation**
  * Clarified cross-version behavior for error and warning counter synchronization, including how unrecognized watermark keys are handled over time.
  * Documented a compatibility approach for transmitting genuinely new signal data via an optional envelope that older servers will ignore.
  * Added guidance on counter value normalization, including how coercible inputs are treated and what gets dropped.

### DIFF
--- a/src/godot_ai/transport/websocket.py
+++ b/src/godot_ai/transport/websocket.py
@@ -505,6 +505,19 @@ def _sync_error_watermark_for_session(session: Session, value: dict[str, int]) -
     current component value is counted when above zero. The debugger and game
     error components overlap (both observe the running game's script errors),
     so their deltas are combined with max(), not summed.
+
+    **Adding a component is a cross-version contract change.** Every integer
+    key that isn't explicitly popped below lands in ``sum(deltas.values())``,
+    so a new counter is folded into the error total by servers that have never
+    heard of it — a plugin stamping one against an older server inflates the
+    "N new errors" doorbell rather than being ignored. There is no allow-list
+    to gate that: ``_PER_RUN_WATERMARK_KEYS`` and ``_WARN_WATERMARK_KEYS`` only
+    route keys they recognize, and everything else falls through to the sum.
+    Carry genuinely new signal (paths, classifications, anything that isn't a
+    count of new errors) in a separate optional envelope field instead, where a
+    server that doesn't know it drops it: non-int values already fail
+    ``_normalized_watermark_int`` and are skipped, so that shape degrades
+    safely on both sides. See #782 for the case that established this.
     """
 
     updates: dict[str, int] = {}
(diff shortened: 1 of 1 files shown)


dsarno:
Both correct, fixed in `87b2df0`. On a comment whose only value is precision, "broader than the implementation" is the whole failure mode — good catches.

Verified each against the code before rewriting:

**Baseline.** For an unrecognized key, `previous is None` and it's not in `_PER_RUN_WATERMARK_KEYS`, so the first stamp falls through both delta branches — it only lands in `updates`. Inflation starts from the **second** stamp, once a previous value exists to diff against. My wording implied the first stamp counted in full. `run_seq` is likewise skipped by `continue` before the delta math, not popped afterward, so "isn't explicitly popped below" mis-described it.

**Normalization.** Confirmed directly:

```
'5'      -> 5
1.5      -> 1
True     -> 1
['a']    -> None
{'x': 1} -> None
'abc'    -> None
```

`int()` is far more permissive than "non-int values fail". That mattered beyond pedantry: the sentence was recommending a separate envelope field *because* non-ints get skipped, so overstating the skip oversold the safety of the very alternative the comment points people to. A list drops out as intended, but a stringified count would silently coerce and be counted.

The rewrite states both precisely, keeps the hazard front and centre, and changes no behavior. `ruff` clean, 11 watermark/websocket unit tests pass.

---

PR #837, opened 2026-07-26
Treat a dead Windows spawn PID with no pid-file as a handoff (#797)

### LINKED ISSUE #797: Windows uv-venv trampoline PID: healthy server spawn mislabeled as "server exited"
## Observed (2026-07-21, #794 Windows smoke, anomaly 2)

On Windows 11 with a **uv-created venv**, one of four editor boots logged `MCP | server exited after 5146ms` while the real server process kept running and was subsequently adopted; the session came up healthy.

## Suspected mechanism

uv venvs on Windows ship `python.exe` as a **trampoline** that spawns the real interpreter as a child process. The plugin's spawn path watches the PID of the process it launched — the trampoline — which exits once the child is up, while the pid-file PID (the real interpreter) lives on. The plugin then mislabels a healthy launch as "server exited", and only the crash-survivor adoption path (#758/#774) rescues the session.

## Impact

- Cosmetic-to-confusing on healthy boots: a false "server exited after Nms" log line, then adoption.
- Potential real cost: anything keyed on the watched spawn PID (teardown decisions, restart heuristics, the dock's server-state label) may act on the trampoline's lifetime instead of the server's. Editor-quit teardown of the adopted server verified correct in the smoke (#669/#774 behavior held), so the blast radius looks contained — but the mislabel itself is a footgun for future logic.

## Repro sketch

Windows + uv venv (root `.venv` whose `python.exe` is a trampoline), plugin-managed server spawn; watch for `MCP | server exited after <n>ms` on editor boot with the server actually alive (intermittent — 1/4 boots in the smoke run).

## Possible directions (needs a maintainer look, not committing to one)

- Resolve the real interpreter PID at spawn time (pid-file handshake already exists) and watch that instead of the trampoline.
- Or treat "watched PID exited but pid-file PID alive within grace window" as a handoff, not an exit — suppress the exit log and skip the exit-path bookkeeping.
- Windows-only; macOS/Linux uv venvs exec rather than trampoline (verify).

Surfaced during the #794 integration smoke; spawn/adoption code was untouched by that wave (`git diff` empty over the relevant files), so this is pre-existing behavior.

---

dsarno:
Closes #797.

## Problem

A uv-created venv on Windows ships `python.exe` as a **trampoline**: it spawns the real interpreter as a child and exits as soon as that child is up. The PID `OS.create_process` returns is the trampoline's, so on a perfectly healthy boot the watched PID dies within a second or two while the real server is still starting.

`check_server_health` already heals onto the pid-file PID as soon as one appears — but until then, a dead spawn PID past `SPAWN_GRACE_MS` fell into the fast-exit branch:

```
MCP | server exited after 5146ms
```

...on a session that was about to come up fine, rescued only by the crash-survivor adoption path (#758/#774). The reported log even lands right at the 5s grace boundary, which matches this exactly.

## Fix

An **absent pid-file plus a dead spawn PID inside a window** is the handoff signature. That inference is safe because `start_server` wipes the pid-file immediately before every spawn (`server_lifecycle.gd:841`), so `read_pid_file() == 0` means "this server has not published yet", never a stale leftover from a previous run.

While that holds, keep watching instead of reporting an exit. Then either:

- the pid-file lands → the **existing** heal branch adopts the real PID (no false log, and the record is corrected), or
- the window closes → the **unchanged** fast-exit diagnosis runs.

**Gated to Windows.** POSIX uv venvs exec rather than trampoline, so a dead spawn PID there really is a dead server; delaying that diagnosis would only slow honest crash reporting on the platforms where this can't happen. The decision is a pure static function taking `os_name` as a parameter rather than calling `OS.get_name()` internally, so the Windows path is exercisable from any host.

`SPAWN_HANDOFF_MS` is 15s — above `SPAWN_GRACE_MS` (or it would change nothing) and below `SERVER_WATCH_MS` (or a genuinely dead Windows spawn would fall off the end of the watch undiagnosed). Both bounds are asserted in tests rather than left to the comment.

One detail worth calling out: the first tick that observes the death is stamped, and the diagnosis reports **that** time rather than the time we gave up waiting. The whole point of #797 is an honest log line, and swapping one misleading number for another would have missed it.

## Verification

- `pytest` — 1643 passed, 5 skipped
- `script/ci-godot-tests` — **2014/2041 passed, 0 failed** (up 5, the new cases)
- Live on Linux: restarted the headless editor so it spawned a fresh server through the modified watch path — `started server (PID 12370, v3.0.7)` → `connected to server`, no false exit line.

### What I could not verify

**The actual Windows trampoline scenario.** I have no Windows host here, so the handoff branch itself has not been run against a real uv trampoline. What is verified: the decision logic under a simulated `"Windows"` `os_name`, that the gate is off on every other platform, and that a real Linux spawn still comes up clean through the modified code.

CI's `Godot tests / Windows` job runs the new tests on a real Windows host and exercises `check_server_health` there for real — but a CI venv is unlikely to be uv-trampolined, so a green result shows no regression rather than confirming the fix. Closing the loop properly wants one boot on Windows 11 with a root `.venv` created by uv, checking that `MCP | server exited after Nms` no longer appears on a healthy start. Happy to leave this open for that if you'd rather not take it on test evidence alone.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows server startup handling to prevent valid process handoffs from being misidentified as crashes.
  * Extended the startup monitoring/grace window to keep watching while the real server process stabilizes.
  * Preserved accurate timing by recording the first observed PID death to avoid premature failure diagnosis.

* **Tests**
  * Added Windows-specific coverage for handoff pending, PID-file arrival, timeout expiry, and non-Windows behavior.
  * Added coverage for the first-death timing stamp logic to ensure correct earliest detection.


dsarno:
Good catch, fixed in `8feb56e`. That test re-implemented the first-write-wins assignment inside itself, so it was asserting against its own copy of the logic — it would have passed even if `check_server_health` stopped stamping entirely. A test of the test.

Rather than adding clock/PID seams, I took the extraction option, matching what `is_spawn_handoff_pending` already does in this change:

```gdscript
static func first_death_stamp(current_stamp_ms: int, elapsed_ms: int) -> int:
	return current_stamp_ms if current_stamp_ms > 0 else elapsed_ms
```

The watch loop now calls it, and the suite drives that same function across successive ticks. Same behavior, but a regression in the production path fails a test now. Seams would have worked too; extraction keeps `check_server_health` readable and avoids widening the host stub for one field.

Also split out the fresh-spawn entry case (`first_death_stamp(0, 5146) == 5146`), which the old test never covered.

Suite: `server_lifecycle` 55/55, full GDScript run 2015/2042, 0 failures.

---

dsarno:
## Probe result: the hop is real, the suspected mechanism is not confirmed

Ran the one-shot probe on a `windows-latest` runner with a uv-created venv, then deleted it (`ecaf65d`). What it found:

```json
{
  "spawn_pid": 7084,
  "pid_file_pid": 6088,
  "pids_differ": true,
  "spawn_pid_still_alive": true,
  "spawn_pid_exited_after_ms": -1,
  "pid_file_pid_alive": true,
  "trampoline_confirmed": false
}
```

`.venv\Scripts\python.exe` is 255,200 bytes — a uv shim, not a real CPython binary.

**Confirmed:** the process hop is real. The server runs under a different PID (6088) than the one `OS.create_process` hands back (7084). Anything keyed on the spawned PID is reasoning about the wrong process.

**Not confirmed:** #797's suspected mechanism, that the shim exits once its child is up. The spawned PID stayed alive for the full 90s, behaving as a live parent. On that runner the fast-exit branch could never have fired.

### Why this doesn't invalidate the fix

The gate keys on the **observable condition** — watched PID dead, no pid-file yet — not on any theory of why it died. Whatever kills the watched process, waiting for the pid-file is the right response; and where the watched PID stays alive, as on this runner, the branch simply never fires. The reported symptom is real and reproducible enough to have been logged with a timestamp (`server exited after 5146ms` with the server alive and subsequently adopted); the fix responds to that condition directly.

What has changed is the **comments**, which asserted the trampoline-exits mechanism as established fact. They now describe the observed symptom and mark the mechanism as unconfirmed, citing the probe's numbers.

### Caveats on the probe, stated plainly

- **One field was mis-instrumented.** `pid_file_published_after_ms` read `90139`, which is just the loop's end time — the loop only breaks when both conditions are met, and one never was. It does *not* mean the pid-file took 90s. Anyone re-running this should fix that before reading it.
- A GitHub runner's uv version and process behavior may differ from the reporter's Windows 11 box. This is one data point, not a survey.

### What this means for merging

The honest position: this makes the change **safe but not proven to fix the reported boot**. It cannot regress non-Windows (gated) and cannot fire when the watched PID stays alive. Whether it resolves the 1-in-4 boot on your actual machine still wants the manual check — one boot on Windows 11 with a uv `.venv`, confirming `MCP | server exited after Nms` no longer appears on a healthy start.

Happy to hold this open for that, or land it as a correctly-targeted guard with the mechanism documented as unconfirmed. Your call.

---

Issue #822, opened 2026-07-23
Post-merge review of #821 (attach bridge): validate findings and fix before next version bump


dsarno:
#821 shipped the attach bridge (merged as `7c2204d2`). I ran a deep three-track review against the exact merged head (`2a3ecd0` — the same commit the stormtest matrix ran on), covering (1) the lock/adopt/lease/reap lifecycle, (2) the stdio proxy and transport-recovery semantics, and (3) the FastMCP compat / CI / test-quality surface.

Headline first: **the core never-duplicate-a-mutation property genuinely holds.** Every ambiguous path abandons or conservatively labels, the single ConnectError replay is provably pre-dispatch, and the property is pinned by a real-subprocess integration test that counts backend-side executions. The stormtest results are consistent with that. The findings below are about the machinery *around* that core — several of them reachable in production configurations — plus a hardening list.

@Clubhouse1661 — please review these and weigh their validity; the analysis is code-level and I've kept file:line references throughout, but you know this design best. Where you agree, let's get a fix PR up **before we bump the version**, so the first release carrying the attach bridge doesn't ship with the known sharp edges.

---

### 1. Reaper ignores bridge leases for plugin-spawned backends

Bridge leases are registered against **any** adopted backend, but only the attach-owned branch of `watch_idle` honors them. `watch_owner` and the plugin-spawned branch of `watch_idle` (`src/godot_ai/server.py:396-404, 419-427`) never consult `leases.active_count`, so on the most common startup order the lease provides no protection:

1. Editor starts first → plugin spawns the backend (`GODOT_AI_PLUGIN_SPAWNED=1`, `--owner-pid`).
2. An attach bridge adopts that backend and registers a live, heartbeating lease.
3. The editor crashes (or is killed).
4. `watch_owner` sees owner dead + zero WS sessions — leases aren't sessions — and SIGTERMs the backend within ~2 polls (~10s), killing the bridge's in-flight tool call with a non-retryable `TRANSPORT_OUTCOME_UNKNOWN` and forcing the instance churn the lease was designed to prevent. The 120s idle backstop is equally lease-blind.

This undermines the #816 editor-independence story exactly where users will hit it: "editor first, then client" is the default real-world sequence. The invariants in #821's description are upheld *as written* ("attach-owned backends exit only after…") — the gap is that plugin-owned backends accept lease registrations they don't honor. Suggested direction: lease-gate those two branches the way the attach branch does, or define an explicit ownership handoff.

[...]

Issue #686, opened 2026-07-12
[S3] Kill-path PID-reuse safety: tracked/adopted PIDs killed without a kill-time brand re-check


dsarno:
Surfaced by the codebase audit. Several server-kill paths trust a PID that was branded (or merely recorded) earlier, and SIGKILL/taskkill it later with no re-verification that the PID is still alive **and** still a godot-ai process. Kernel PID recycling turns each into a "kill an innocent process" hazard.

### a. `stop_server` never re-verifies the tracked PID (`server_lifecycle.gd:~961`)
`stop_server` seeds `candidates = [int(_server_pid)]` unconditionally (guarded only by `<= 0`) and kills it. Every *other* candidate in the same function is gated by `_pid_cmdline_is_godot_ai_for_proof`; the tracked pid gets no `pid_alive` and no brand check. Nothing clears `_server_pid` when the server dies mid-session (`check_server_health` stops watching after `SERVER_WATCH_MS`).

**Failure:** server crashes hours into a session, OS recycles the PID to an unrelated process, user later quits Godot → `_exit_tree` → `stop_server` kills the innocent process with its child tree.

**Fix:** seed `candidates` conditionally — only append `_server_pid` when `_pid_alive_for_proof(pid) and _pid_cmdline_is_godot_ai_for_proof(pid)` (a false negative is fail-safe: the port stays held and the record is preserved).

### b. `force_restart_server` kills raw listener PIDs with no per-PID brand check (`server_lifecycle.gd:~1125`)
Gated only by `can_restart_managed_server()` (proof we once managed *a* server, not that the *current* occupants are ours), it then kills `_find_all_pids_on_port(port)` — raw lsof/netstat PIDs, no brand check, no live re-probe.

**Failure:** adopted godot-ai server exits on its own (crash / idle backstop) before the user clicks the drift Restart button; another dev tool binds 8000; click → the unrelated tool is killed.

**Fix:** brand-gate the kill list the way `stop_server` does (keep only PIDs where `_pid_cmdline_is_godot_ai_for_proof(pid)`).

### c. `status_matches_record` strong-proof tier authorizes an **automatic** unbranded kill (`plugin.gd:~1173`)
[...]

PR #839, opened 2026-07-26
Honor active attach leases when the editor exits normally (#824)

### LINKED ISSUE #824: Make plugin-managed backend teardown honor active attach leases on clean editor exit
## Problem

PR A for #822 makes both Python reaper paths honor active attach-bridge leases:

- the owner-PID watchdog;
- the plugin-spawned idle backstop.

This allows a plugin-spawned backend to survive an editor crash while one or more MCP clients still hold active leases.

A normal editor exit still behaves differently. The GDScript `_exit_tree` teardown path stops a plugin-managed backend without consulting its active leases. The following startup order therefore still causes backend churn:

1. Open Godot; the plugin spawns the backend.
2. Start an MCP client; its `godot-ai attach` bridge adopts the backend and registers a lease.
3. Close Godot normally.
4. The plugin kills the backend despite the live bridge lease.
5. An in-flight call may become `TRANSPORT_OUTCOME_UNKNOWN`, and the bridge must establish a new backend instance before the editor can reconnect.

The desired behavior is for a live client lease to transfer the backend into externally managed/detached lifetime when the editor exits cleanly.

## Proposed behavior

Before stopping a plugin-managed backend during normal editor teardown:

1. Verify the backend using the existing managed-process and instance identity rules.
2. Query whether the current backend instance has any unexpired attach leases.
3. If no leases are active, preserve the existing clean-stop behavior.
4. If leases are active:
   - do not kill the backend;
   - clear the plugin’s managed PID/record state;
   - detach using semantics equivalent to the existing `keep_server_on_exit` path;
   - allow the Python lease-aware reaper to stop the backend after there are no editor sessions, no active leases, and the configured grace period has elapsed.

Possible backend surfaces include an instance-bound lease-status endpoint or an additive status field. Any lease information is advisory for deciding whether to detach; it must never become authorization to kill an otherwise unverified process.

## Design constraints

- Preserve the plugin’s existing no-proof-no-kill rule.
- Do not transfer kill authority merely because a process self-reports `owner_type = "plugin"`.
- Keep teardown bounded; editor exit must not hang on a slow or unavailable HTTP request.
- Preserve existing `keep_server_on_exit` behavior.
- Preserve the external-adoption behavior where `_server_pid = -1` and the plugin has no authority to kill the adopted process.
- Account for multiple simultaneous bridges. The backend must remain alive until the last lease is released or expires.
- Document the lease-bounded tokenless-loopback exposure window.
- Treat this as teardown-ordering-sensitive work; review against the lifecycle safeguards established around #46.

## Required tests

- **No active lease:** editor-spawned backend stops on normal editor exit exactly as it does today.
- **One active lease:** normal editor exit leaves the same backend PID and `instance_id` alive.
- **Multiple active leases:** releasing one bridge does not stop the backend; releasing or expiring the final lease allows it to reap.
- **Editor restart:** a new editor can adopt/reconnect to the detached backend without restarting the MCP client.
- **Lease query is instance-bound:** stale lease information from a replaced backend cannot affect teardown.
- **Foreign or incompatible port occupant:** the plugin never kills it.
- **`keep_server_on_exit`:** existing behavior remains unchanged.
- Teardown completes within a bounded time when the lease-status request fails or times out.

## Live acceptance flow

1. Start the editor first and confirm it spawned the backend.
2. Start `godot-ai attach` and confirm a lease is active.
3. Close the editor normally.
4. Confirm the backend PID and `instance_id` did not change.
5. Reopen the editor and confirm tools recover without restarting the MCP client.
6. Close the editor again, release the final bridge lease, and confirm the backend reaps after its grace period.

## Acceptance criteria

- A plugin-managed backend with an active attach lease survives normal editor exit.
- The same backend instance can accept a later editor reconnection.
- A backend with no editor sessions and no unexpired leases still terminates automatically.
- Existing no-lease, external-adoption, and `keep_server_on_exit` behavior does not regress.
- No new path can kill a foreign or insufficiently verified process.

dsarno:
Closes #824.

## Problem

Both Python reaper paths already honor attach-bridge leases (#822 PR A). The GDScript teardown did not, so this order still caused churn:

1. Open Godot → plugin spawns the backend
2. Start an MCP client → its `godot-ai attach` bridge adopts that backend and registers a lease
3. Close Godot normally → **the plugin kills the backend despite the live lease**

An in-flight call can become `TRANSPORT_OUTCOME_UNKNOWN`, and every bridge must establish a new backend before the next editor can reconnect.

## Change

**Backend** — `/godot-ai/status` gains `active_lease_count`. Additive, so an older plugin ignores it and an older backend simply omits it. Advisory only, consistent with the existing fields: it may justify *declining* to kill, never permission to kill. Instance-bound by construction, since the count ships in the same response as the `instance_id` it belongs to and so cannot be stale relative to that instance.

**Plugin** — `teardown_for_editor_exit` consults it before stopping. With leases held it drops the managed record and detaches instead of killing, giving up kill authority so the next editor adopts the survivor through the external branch rather than as a managed server it may kill.

Two deliberate choices worth flagging for review:

- **The server's own pid-file is left in place.** It's the backend's publication, not the plugin's claim on it, and adoption reads it. Only the managed record — the thing that confers kill authority — is dropped.
- **Nothing leaks.** A plugin-spawned backend keeps its idle backstop armed (only `keep_server_on_exit` disarms it) and that backstop is lease-aware, so this *defers* the stop to "no editors AND no leases AND grace elapsed" rather than stranding a process.

## Fail direction

Every uncertain case reads as zero leases and keeps today's kill-on-exit behavior: no managed PID, a probe that fails or times out, a process that doesn't identify as `godot-ai`, or a backend too old to publish the field. A false zero costs exactly what today already costs; a false positive would leave a process running on a guess.

Bounded by `SERVER_STATUS_PROBE_TIMEOUT_MS`, so editor exit can't hang on a wedged server, and the probe is skipped entirely when there's no managed PID.

## Verification

- `pytest` — 1644 passed, 5 skipped
- `script/ci-godot-tests` — **2021/2048, 0 failed** (+6, the new cases)
- **Live against the running plugin-spawned backend** — this is the real one, since the whole feature hinges on a plugin-owned server reporting leases truthfully:

```
status: name=godot-ai owner_type=plugin active_lease_count=0
POST /godot-ai/lease/register  -> active_lease_count: 1
POST /godot-ai/lease/release   -> active_lease_count: 0
```

## Tests

GDScript covers the issue's acceptance list: detach with a lease held (record cleared, PID untracked, nothing killed), kill with no leases, a backend too old to publish the field, a non-`godot-ai` process that can't talk the editor out of a clean stop, a clamped nonsense value, and no probe at all when there's no managed PID. The pre-existing `keep_server_on_exit` teardown tests are untouched and still pass, which is the regression guard for that path.

Python covers the additive field and a register/release round trip through the real lease routes. The exact-payload contract test in `test_server_status.py` was updated to include the new key — it exists to catch exactly this kind of surface change.

---

## Summary by CodeRabbit

* **New Features**
  * Server status (`/godot-ai/status`) now includes `active_lease_count`, indicating how many editor attachment leases are currently active.
* **Bug Fixes**
  * Editor shutdown handling for keep-alive servers is now lease-aware: if active leases are present, the server is detached for later adoption rather than terminated; otherwise it is stopped as before.
* **Tests**
  * Updated and expanded lifecycle and status-route coverage for attach leases, including missing/invalid counts, zero vs. positive behavior, and multi-lease tracking.

### DIFF
--- a/plugin/addons/godot_ai/plugin.gd
+++ b/plugin/addons/godot_ai/plugin.gd
@@ -889,15 +889,42 @@ static func _probe_live_server_status(port: int, timeout_ms: int = SERVER_STATUS
 	if not (parsed is Dictionary):
 		result["error"] = "invalid_json"
 		return result
-	result["reachable"] = true
-	result["name"] = str(parsed.get("name", ""))
-	result["version"] = _extract_server_version(parsed)
-	result["ws_port"] = int(parsed.get("ws_port", 0))
-	## `package_path` was added in v2.4.4 (#416) so the dock's
-	## "Incompatible server" banner can name the source of a version
-	## skew. Older servers omit it; treat the missing field as "".
-	result["package_path"] = str(parsed.get("package_path", ""))
+	result.merge(_project_status_payload(parsed), true)
 	return result
+
+
+## Project a parsed `/godot-ai/status` body into the probe's result shape.
+##
+## Extracted from the probe so it can be tested against a real payload. The
+## probe is a whitelist — a field the server publishes does not reach callers
+## unless it is copied here — and that is silent: the consumer just sees a
+## missing key. #824's lease check shipped reading `active_lease_count` while
+## this projection dropped it, so the branch was dead on every platform, and
+## the tests could not see it because they hand-built the result dict this
+## function is supposed to produce. Add new fields here, and cover them with a
+## projection test rather than a fabricated `live_status`.
+static func _project_status_payload(parsed: Dictionary) -> Dictionary:
+	var projected := {
+		"reachable": true,
+		"name": str(parsed.get("name", "")),
+		"version": _extract_server_version(parsed),
+		"ws_port": int(parsed.get("ws_port", 0)),
+		## `package_path` was added in v2.4.4 (#416) so the dock's
+		## "Incompatible server" banner can name the source of a version
+		## skew. Older servers omit it; treat the missing field as "".
+		"package_path": str(parsed.get("package_path", "")),
+	}
+	## #824: advisory attach-lease count, consumed by teardown to decide
+	## detach-vs-kill. Absent stays absent rather than defaulting to 0, so
+	## `ServerLifecycleManager.active_lease_count` keeps distinguishing "backend
+	## too old to publish this" from "backend reports zero leases" — both stop
+	## the server, but only one of them is a compatibility statement.
+	## Non-numeric JSON is dropped for the same reason it is clamped
+	## downstream: a malformed value must not read as occupancy.
+	var raw: Variant = parsed.get("active_lease_count")
+	if raw is float or raw is int:
+		projected["active_lease_count"] = int(raw)
+	return projected
 
 
 func _probe_live_server_status_for_port(port: int) -> Dictionary:
--- a/plugin/addons/godot_ai/plugin.gd
+++ b/plugin/addons/godot_ai/plugin.gd
@@ -919,11 +919,16 @@ static func _project_status_payload(parsed: Dictionary) -> Dictionary:
 	## `ServerLifecycleManager.active_lease_count` keeps distinguishing "backend
 	## too old to publish this" from "backend reports zero leases" — both stop
 	## the server, but only one of them is a compatibility statement.
-	## Non-numeric JSON is dropped for the same reason it is clamped
-	## downstream: a malformed value must not read as occupancy.
+	## Anything that is not a finite whole number is dropped, for the same
+	## reason the value is clamped downstream: a malformed count must not read
+	## as occupancy and keep a server alive. Godot parses every JSON number as
+	## a float, so the whole-number test is what distinguishes a real count
+	## from junk — truncating 1.5 to 1 would manufacture a held lease.
 	var raw: Variant = parsed.get("active_lease_count")
-	if raw is float or raw is int:
-		projected["active_lease_count"] = int(raw)
+	if raw is int or raw is float:
+		var numeric := float(raw)
+		if is_finite(numeric) and numeric == floor(numeric):
+			projected["active_lease_count"] = int(numeric)
 	return projected
 
 
--- a/plugin/addons/godot_ai/utils/server_lifecycle.gd
+++ b/plugin/addons/godot_ai/utils/server_lifecycle.gd
@@ -984,19 +984,23 @@ func _start_server_impl(async_gen: int) -> void:
 ## the watch crossed SPAWN_GRACE_MS and reported an exit — rescued only by the
 ## crash-survivor adoption path.
 ##
-## A uv venv's `python.exe` is a shim rather than the interpreter, and the
-## real server does run under a *different* PID than the one
-## `OS.create_process` hands back — confirmed on a Windows runner (spawned
-## 7084, pid-file 6088). What is NOT confirmed is the original report's
-## suspected mechanism, that the shim exits once its child is up: on that
-## runner the spawned PID stayed alive for 90s, behaving as a live parent.
-## So the shim's exit is one possible cause of the reported death, not an
-## established one.
+## A uv venv's `python.exe` is a shim rather than the interpreter, and the real
+## server does run under a *different* PID than the one `OS.create_process`
+## hands back. But the original report's suspected mechanism — that the shim
+## exits once its child is up — is **disproven**, not merely unconfirmed. A
+## 12-boot run on Windows 11 with a uv venv found the spawned trampoline alive
+## on every boot, with the child owning both the pid-file and the listener; a
+## CI runner showed the same. The shim is a live parent for the process's whole
+## life, so it is not what kills the watched PID.
 ##
-## This gate is therefore keyed to the observable condition — watched PID
-## dead, no pid-file yet — and not to any theory of why it died. Whatever
-## kills it, waiting for the pid-file is the correct response, and when the
-## watched PID stays alive (as on that runner) this branch simply never fires.
+## Two consequences worth keeping straight. First, this gate is keyed to the
+## observable condition — watched PID dead, no pid-file yet — not to any theory
+## of why it died, so it stays correct whatever the cause. Second, and less
+## comfortable: in that same 12-boot run the false "server exited" line never
+## appeared AND the watched PID never died, so the guard never fired. Those
+## clean boots are evidence the symptom did not reproduce, NOT evidence this
+## guard fixes it. The true cause of the original 1-in-4 report is still
+## unknown; if it resurfaces, start from that rather than from the trampoline.
 ##
 ## `real_pid <= 0` means no pid-file exists yet, and that reliably means "this
 ## server has not published one" rather than "stale leftover": `start_server`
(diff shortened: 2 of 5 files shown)


dsarno:
Two findings: took the coverage gaps, declining the main-thread one with evidence.

## Taken — instance binding and multi-bridge (`ada2d65`)

The Linked Issues check was right that instance binding was thinner than my comment claimed. The count comes from whoever answers on the port, which is **not** proof it's the process about to be stopped — another editor's backend, or an attach-owned one, can hold the port after ours dies. A stranger's leases could have talked this editor out of stopping its own server.

Now the same alive+branded proof `stop_server` uses gates the check, and it short-circuits before the probe. Failing it is harmless either way: a dead PID has nothing to kill, and a recycled-but-unbranded PID is refused by `stop_server`'s own #686 gate — both land on the historical path.

Multi-bridge was genuinely untested. Added: the gate is a threshold, not an equality (releasing one of several keeps the backend up; only the last release returns it to kill-on-exit), driven through the real lease routes 2 → 1 → 0.

Editor-restart adoption remains a live-smoke item — it's a two-process, two-lifetime flow that a unit test can only fake. It's in the Windows smoke running now, and I'd rather report that result than assert it from a stub.

## Declined — moving the probe off the main thread

The guideline is real but scoped to a different phase. `AGENTS.md:34`: *"All GDScript executes in `_process()` with a 4ms frame budget."* This runs in `_exit_tree`, where there is no next frame to protect.

More concretely, the very next line on this path already blocks far harder than the probe does. `stop_server` calls `_wait_for_port_free(port, 2.0)`, plus `_find_managed_pid` and several `_pid_cmdline_is_godot_ai_for_proof` calls — each spawning a subprocess (`netstat`/`lsof`/`tasklist`). An 800ms bounded HTTP probe is strictly smaller than the 2s port wait it precedes, so deferring it would leave the actual blocking untouched.

And deferring during teardown would be actively unsafe here. `teardown_for_editor_exit` → `stop_server`/`detach_server` both begin with `_invalidate_async_startup()`, which exists to **join** in-flight worker threads before the plugin frees — the macOS reload-churn wedge (use-after-free on the worker) is why. Starting new async work at that exact moment reintroduces the bug class that hardening was written to prevent.

There's also a correctness constraint: the teardown decision needs the answer before it can choose kill vs. detach. Deferring means either joining (blocking anyway) or deciding without it, which defeats the feature.

#824 asked for teardown to be **bounded**, not asynchronous, and `SERVER_STATUS_PROBE_TIMEOUT_MS` is what bounds it.

## Verification

`ruff` clean · `pytest` 1645 passed, 5 skipped · `ci-check-gdscript` OK · `ci-godot-tests` **2024/2051, 0 failed** (+9 over base).

---

dsarno:
Fair, and fixed in `ca42ae9`. Both gaps were real:

- The routing test held **exactly one** lease, so narrowing the gate to `== 1` would have passed it.
- The proof-gate case only asserted the helper's return value, never that teardown then actually reached `stop_server`.

Now `teardown_for_editor_exit()` is driven for both. Three leases must still detach; an unprovable PID must reach the stop path *despite* a backend reporting four leases, and must not fire the detach path's record clear — `stop_server` deliberately preserves the record when the port is still held so the next `start_server`'s drift branch can retry the kill.

`finalize_calls` is the discriminator, since `stop_server` ends in `_finalize_stop_if_port_free` and `detach_server` never touches it. That makes the routing assertion unambiguous rather than inferred from side effects.

`server_lifecycle` 66/66; full run **2025/2053, 0 failed**.

Not addressing the Docstring Coverage warning: it is counting the two new *test* functions plus a nested helper, and this file documents with comment blocks rather than docstrings — `test_status_route_reports_live_server_version` directly above them has none. Adding docstrings there to clear a threshold would fight the file's own convention.

---

dsarno:
## Fixed in `74b245b` — the Windows smoke found a real defect, and it wasn't Windows-specific

`_probe_live_server_status` builds its result as a **whitelist** (`reachable`, `name`, `version`, `ws_port`, `package_path`) and never copied `active_lease_count` out of the parsed JSON. So `live.get("active_lease_count")` was always null, the count was always 0, and `teardown_for_editor_exit` always fell through to `stop_server()`. **The lease branch was unreachable on every platform.** The backend published the field correctly; the plugin dropped it on the floor.

### Why the tests missed it, which is the more important part

They set `host.live_status = {"name": "godot-ai", "active_lease_count": 1}` — asserting the consumer against a dict shape **the real probe never produces**. And my "verified live" claim in the original commit checked the *backend endpoint*, which was never the broken half. Both halves were tested; the seam between them was not.

That's the same failure mode flagged [two commits earlier](https://github.com/hi-godot/godot-ai/pull/839#issuecomment-5085633409) about the death-stamp test, in a worse form: there the test copied production logic, here it fabricated production data.

### The fix

Extracted the projection into `_project_status_payload`, so it is testable against a real `/godot-ai/status` body, and added a test that drives that body **through the projection into the consumer** — the two halves can no longer drift.

I verified the new test has teeth rather than assuming: reverted the projection line, ran the suite, watched `test_status_projection_carries_the_lease_count_to_the_consumer` fail, then restored it and watched it pass. That property is exactly what the original tests lacked.

Two behaviors preserved deliberately: absent stays absent (so "backend too old to publish this" stays distinguishable from "reports zero"), and a non-numeric value is dropped, so malformed JSON can't read as occupancy.

The probe's docstring now says plainly that it is a whitelist and that new fields need a projection test, since the silent-drop failure is invisible from the consumer side.

### Verification

Reproduced the smoke's B1 on Linux through the real path — lease registered against the live plugin-spawned backend, clean editor quit through `teardown_for_editor_exit`:

```
MCP | detaching server: 1 attach lease(s) still held, leaving it to the server's own idle reaper
backend after exit: ALIVE, same instance_id
```

`ruff` clean · `pytest` 1645 passed · `ci-godot-tests` **2029/2056, 0 failed**.

Thanks for the report — the `instance_id` comparison in particular, since the port answering afterward makes this look green to a naive check.

---

dsarno:
Taken, in `bcce097`. The finding is correct on its own terms: Godot parses every JSON number as a float, so `raw is float` accepted `1.5` and `int()` truncated it to `1` — manufacturing a held lease out of a malformed payload and keeping a server alive on it. That directly contradicts the rule the surrounding comment states.

Now requires a **finite whole number**. Whole floats (`2.0`) are the normal wire shape and still project; `1.5`, `±INF` and `NAN` are dropped alongside the non-numerics.

Added both halves to the test, since the guard has a failure mode in each direction — rejecting floats outright would have dropped every real count:

- junk row extended with `1.5, 0.5, -2.5, INF, -INF, NAN` → must omit the field
- new case asserting `0.0, 1.0, 2.0, 9.0` **do** project, because that is what the wire actually carries

Worth stating the reachability honestly: this needs a process that identifies as `godot-ai` and sends a fractional count, so it's contract hygiene rather than an observed failure. I took it because the guard is two lines and the alternative is code that doesn't do what its own comment claims — which is the same class of problem as the projection bug this PR just fixed.

`server_lifecycle` 70/70 · full run **2030/2057, 0 failed**.

---

PR #841, opened 2026-07-26
Capture forensics when a spawn is judged to have fast-exited (#797)

dsarno:
Follow-up to #797, which is closed but **not solved**.

## Why

A 12-boot run on the reported configuration (Windows 11, uv venv) reproduced neither the symptom nor its suspected mechanism. The trampoline was alive on every boot, with the child owning both the pid-file and the listener — so "the shim exits once its child is up" is ruled out as the cause. Something killed that watched PID once in four boots, and we still don't know what.

The reason it stayed unexplained is that the log line at the time carried no evidence:

```
MCP | server exited after 5146ms
```

That's the whole record. Asking someone to reproduce a 1-in-4 bug under observation is a poor plan; capturing the state at the moment of judgement is a better one.

## What it emits

One line at the top of `_diagnose_spawn_fast_exit`, with the facts that would settle it:

```
#797 spawn-exit forensics: shape=handoff_child_alive os=Windows launch=dev_venv
elapsed=5146ms first_dead=5146ms spawn_pid=30188(alive=false)
pid_file_pid=9360(alive=true) listeners=[9360]
```

It **names the shape** rather than leaving it to be inferred:

| shape | meaning |
|---|---|
| `handoff_child_alive` | watched PID gone, a different live PID owns the pid-file — the launcher-handoff hypothesis |
| `watched_pid_still_alive` | the death that triggered this was transient |
| `no_pid_file_published` | the server never got far enough to publish |
| `all_dead` | an ordinary crash |

`watched_pid_still_alive` is the one worth having. Given the trampoline demonstrably never dies, a fast-exit diagnosed while the watched PID is *alive on re-read* would mean a transient liveness read — a different bug from a process that really exited, and one nobody would guess from the old line.

`elapsed` and `first_dead` are reported separately because they diverge after a #837 handoff wait: one is when we gave up, the other is when the process actually died. Conflating them would misdate the event in exactly the reports this exists to serve.

## Cost and safety

Nothing in the healthy case — the path only runs when a spawn is already being declared dead, and every seam is one the surrounding diagnosis already calls. No behavior change; logging only.

I verified every seam exists on the **real** host rather than only the test stub (`_read_pid_file_for_proof`, `_find_all_pids_on_port`, `_pid_alive_for_proof`, `get_server_launch_mode`, `_log_buffer.log`). That check is a direct response to the #824 projection bug, which shipped green precisely because a stub supplied something production didn't.

## Verification

- `script/ci-check-gdscript` — OK
- `script/ci-godot-tests` — **2035/2062, 0 failed** (+5)
- `ruff` clean (no Python touched)

Formatting is a pure function, so the contract is tested without a live editor. **What is not verified: the line firing in production** — the condition is a spawn fast-exit, which I can't force here. The call site is on the diagnosis path and its seams are real, but the first genuine sighting will be the next occurrence, which is the point.

The line is prefixed with the issue number so a future reporter can search for it.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced Windows “fast-exit” diagnostics by emitting a one-line “spawn-exit forensics” snapshot when a server terminates immediately after launch.
  - The snapshot reports timing details (elapsed vs. first-dead), PID/identity evidence, and a deterministic `shape=` classification to clearly distinguish handoff/child-alive, watched-process-alive, missing PID publication, and fully dead scenarios.

- **Tests**
  - Added coverage to verify correct `shape` classification, that `elapsed` differs appropriately from `first_dead`, and that the output is a strict single-line string with the expected prefix.

### DIFF
--- a/plugin/addons/godot_ai/utils/server_lifecycle.gd
+++ b/plugin/addons/godot_ai/utils/server_lifecycle.gd
@@ -1050,11 +1050,24 @@ static func first_death_stamp(current_stamp_ms: int, elapsed_ms: int) -> int:
 ## through seams the surrounding diagnosis already uses, on a path that only
 ## runs when a spawn is being declared dead, so it costs nothing in the
 ## healthy case.
+## Deliberately does NOT scrape the port for listener PIDs. This runs from the
+## 1 Hz watch loop, on a live frame, so a `_find_all_pids_on_port` subprocess
+## here would stall the editor for a diagnostic. Deferring it via
+## `_run_blocking` was the alternative and is worse: that helper is
+## `await`-based, so it would turn this, `_diagnose_spawn_fast_exit` and
+## `check_server_health` into coroutines — making the watch callback resume
+## across arbitrary frames while its branches set terminal state and trigger
+## re-adoption walks. That is the teardown-ordering hazard
+## `_invalidate_async_startup` exists to contain, and it is not worth taking
+## on for a log line.
+##
+## Little is lost: the probe on the very next line already establishes whether
+## a godot-ai server answers on the port, and `_diagnose_spawn_port_conflict`
+## names a foreign occupant when there is one. If you are tempted to add the
+## PID list back, put it behind that existing conflict path rather than here.
 func _log_spawn_exit_forensics(elapsed: int) -> void:
-	var port := ClientConfigurator.http_port()
 	var spawn_pid := int(_server_pid)
 	var pid_file_pid := int(_host._read_pid_file_for_proof())
-	var listeners: Array[int] = _host._find_all_pids_on_port(port)
 	_host._log_buffer.log(format_spawn_exit_forensics({
 		"os": OS.get_name(),
 		"launch_mode": ClientConfigurator.get_server_launch_mode(),
@@ -1069,7 +1082,6 @@ func _log_spawn_exit_forensics(elapsed: int) -> void:
 		"spawn_alive": spawn_pid > 0 and bool(_host._pid_alive_for_proof(spawn_pid)),
 		"pid_file_pid": pid_file_pid,
 		"pid_file_alive": pid_file_pid > 0 and bool(_host._pid_alive_for_proof(pid_file_pid)),
-		"listeners": listeners,
 	}))
 
 
@@ -1078,7 +1090,6 @@ func _log_spawn_exit_forensics(elapsed: int) -> void:
 static func format_spawn_exit_forensics(facts: Dictionary) -> String:
 	var spawn_pid := int(facts.get("spawn_pid", 0))
 	var pid_file_pid := int(facts.get("pid_file_pid", 0))
-	var listeners: Array = facts.get("listeners", [])
 	## The single most diagnostic bit, stated rather than left to be inferred:
 	## a live pid-file process while the watched one is gone is the launcher
 	## handoff shape; both gone is a real crash.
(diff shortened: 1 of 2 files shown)


dsarno:
Concern taken, prescribed fix declined — removed the cost instead. `364045e`.

**Why this one lands where the #824 teardown probe didn't.** That objection was about `_exit_tree`, where there is no next frame to protect. This runs from the **1 Hz watch loop, on a live frame**, so the 4ms budget genuinely applies. Different context, valid finding.

**Why not `_run_blocking`.** It's `await`-based (`server_lifecycle.gd:157`), so routing through it would make `_log_spawn_exit_forensics` a coroutine, then `_diagnose_spawn_fast_exit`, then `check_server_health` itself. The watch callback would resume across arbitrary frames while its branches set terminal diagnoses, stop the watch, and trigger re-adoption walks. That's precisely the teardown-ordering hazard `_invalidate_async_startup` exists to contain — and `_run_blocking` returns `null` for a cancelled walk, so the forensics could silently not log during exactly the teardown races it's meant to illuminate. Not a trade worth making for a log line.

**So the scrape is gone.** Little is lost: the probe on the very next line already establishes whether a `godot-ai` server answers on the port, and `_diagnose_spawn_port_conflict` names a foreign occupant when there is one — the PID list was largely redundant with both. The comment now records why the field is absent and where to put it if it's ever genuinely needed, so it doesn't get helpfully re-added into a frame stall.

Everything remaining is read from state already in hand plus a single pid-file read, so **the forensics now adds no measurable main-thread cost at all** — a better outcome than deferring the scrape would have been.

`server_lifecycle` 75/75 · full run **2035/2062, 0 failed** · Codecov green.

---

dsarno:
Correct, and a good catch — fixed in `958955f`. You traced the call chain; I trusted a test that couldn't see it.

`check_server_health` calls `_diagnose_spawn_fast_exit(_spawn_dead_since_ms)` (#837 passes the true death time so the user-facing line stays honest), and the forensics forwarded that same value as `elapsed_ms` while separately reading `_spawn_dead_since_ms` for `first_dead_ms`. **The two fields always printed the same number** — the one distinction they exist to record, collapsed. The PR description claimed they diverge. They did not.

Rather than compute at the call site, `_log_spawn_exit_forensics` now derives the diagnosis timestamp from `_server_spawn_ms` itself and takes no `elapsed` parameter at all, so no caller can make the two converge again.

**The test failure is the part worth noting.** `test_forensics_reports_true_death_time_separately_from_detection` fed the pure formatter two different values and asserted they stayed different — which production never does. That's the third instance in this PR series of the same mistake: a test supplying data the real call path doesn't. Replaced with one that drives `_log_spawn_exit_forensics` against a real `McpLogBuffer` and asserts on the **actual logged line**.

Verified it has teeth by reintroducing the bug and watching it fail:

```
CATCHES BUG: test_forensics_wiring_reports_diagnosis_time_not_the_death_time
  -> elapsed must be the diagnosis time, not a copy of the death time
```

One process note others may hit: my first attempt at that negative check reported `failed: 0` while running **zero tests** — the suite wouldn't instantiate because the same edit had a parse error (`Cannot infer the type of "lines"`), and `script/ci-check-gdscript` does not catch that class. A green-looking `failed: 0` with `total: 0` is indistinguishable from success unless you read `total`. Worth reading both.

`server_lifecycle` 75/75 · full run **2035/2062, 0 failed**.

---

Issue #838, opened 2026-07-26
Migrate remaining MCP clients to the `godot-ai attach` bridge


dsarno:
Rollout successor to #816, which is closed: its architecture shipped (bridge + singleton startup + status validation + stdio proxying in #821, lease/transport hardening in #825 and #826, first client migrated in #827). What remains is per-client migration, which is mechanical but gated on strategy work that does not exist yet.

## The prerequisite, first

`clients/_base.gd` defines the shapes:

```gdscript
enum CommandShape { NONE, FLAT, TYPED_FLAT, COMMAND_ARRAY, NESTED_COMMAND }
```

and says outright: *"Only COMMAND_ARRAY is consumed by the TOML strategy in the first #816."*

So today exactly one strategy understands command shapes. `_json_strategy.gd`, `_cli_strategy.gd` and `_yaml_strategy.gd` have no shape support at all — **21 of 22 client descriptors are blocked behind three strategy implementations**, not behind 21 small edits. Anyone who picks this up expecting "migrate a few clients" hits that wall on the first one.

## Sequence

**1. Shape support in the remaining strategies** (blocks everything else)

- [ ] `_json_strategy.gd` — FLAT, TYPED_FLAT, NESTED_COMMAND as required by its clients
- [ ] `_cli_strategy.gd`
- [ ] `_yaml_strategy.gd`

Each needs the same data-only descriptor treatment TOML already got: command/argument representation, transport discriminator, legacy URL keys to remove, user-owned fields to preserve, verification rules.

**2. One verified representative per shape, with a real startup-order smoke**

Before batching anything, prove each shape against a client whose command form you have actually verified. #816's warning stands and is the reason this step exists: *do not assume all JSON clients share Claude Desktop's flat format.* The smoke is the one from #816's test list — client started before Godot, Godot started before client, tools usable after Godot appears without restarting the client.

**3. Claude Desktop first among the clients**

It is the only client currently routed through a third-party `mcp-proxy`, so it has the worst status quo and the most to gain. This was step 7 of #816's plan for the same reason.

**4. Batch the remainder**, per shape, once that shape has a verified representative.

**5. Docs and dock labels** (#816 steps 10–11)

- [ ] README manual-configuration examples match generated attach config
- [ ] `docs/port-conflicts.md` — drop the obsolete "edit the URL field" guidance for attach-mode users
- [ ] Dock labels distinguish URL / attach bridge / plugin-managed / externally adopted
[...]

PR #845, opened 2026-08-01
Migrate Claude Desktop to the attach transport

Clubhouse1661:
## Summary

This PR implements the `FLAT` JSON command-shape slice of #838 and migrates Claude Desktop from the third-party `mcp-proxy` bridge to the client-owned `godot-ai attach` transport.

Claude Desktop now receives a local stdio launch entry that can start or adopt the shared HTTP backend independently of Godot’s startup order. This lets Claude discover tools before the editor opens and remain connected across normal editor restarts.

The PR also fixes Store/MSIX config delivery on Windows. Auto-configure now writes the config file that Store-installed Claude actually reads instead of reporting success after writing the conventional roaming path.

## Why

Claude Desktop was the final client using `mcp-proxy`. Its published releases declare an unbounded `mcp>=1.17.0` dependency, so cold resolution now installs the incompatible `mcp 2.0.0` release:

- `mcp-proxy 0.11.0` fails first on the removed `streamablehttp_client` import.
- `mcp-proxy 0.12.0` fails first on `request_ctx`, but carries both incompatible imports.

Migrating Claude Desktop to `godot-ai attach` removes that dependency and aligns it with the startup-order-independent transport already used by Codex.

During live verification, a second delivery problem was found. Store Claude reads a virtualized private config under:

`%LOCALAPPDATA%\Packages\Claude_*\LocalCache\Roaming\Claude\claude_desktop_config.json`

Godot runs outside the MSIX container, so writing `%APPDATA%\Claude\claude_desktop_config.json` could succeed and verify against itself while Claude never saw the entry.

### JSON command-shape support

- Threads resolved attach-launch data through JSON configure, status, verification, and manual-instruction paths.
- Implements `CommandShape.FLAT` rendering and verification in `_json_strategy.gd`.
- Rejects unsupported future JSON command shapes with an actionable error instead of rendering an empty `{}` entry.
- Adds the `command_supports_url_fallback` capability flag rather than deriving URL support from `command_shape`.
  - Codex explicitly enables it.
  - Claude Desktop leaves it disabled because local `claude_desktop_config.json` entries are stdio-only.

### Lossless Claude Desktop migration

Claude Desktop now declares:

- `CommandShape.FLAT`
- `command_legacy_keys = ["url"]`
- `command_env_legacy_keys = ["UV_LINK_MODE"]`

Reconfiguring an old entry:

- replaces the legacy `mcp-proxy` command and arguments;
- removes the conflicting legacy `url`;
- removes only the obsolete `UV_LINK_MODE` environment value;
- preserves unrelated environment variables;
- preserves user-controlled fields such as `disabled`;
- preserves unknown fields added by future Claude versions.

A failed launch lookup leaves the existing file byte-identical.

### Store/MSIX config resolution

Descriptors can now provide ordered, platform-specific `config_path_candidates` containing one wildcard directory segment.

For Claude Desktop on Windows:

- a unique `Claude_*` Store package root is authoritative;
- its private config is used or created directly, avoiding reliance on MSIX copy-on-write read-through;
- conventional roaming is used when no Store package exists;
- multiple matching Store packages fail closed;
- ambiguity is returned as a structured diagnostic and surfaces as `Status.ERROR`;
- Configure, status, Remove, manual instructions, and post-write verification use the same resolution path;
- the package-family publisher suffix is not hardcoded;
- Configure never writes both config copies.

The ambiguity check runs before launcher discovery, so users see the actual package-path problem rather than `NOT_CONFIGURED`, a launcher error, or “could not resolve this OS.”

### Cleanup and status responsiveness

- Removes the obsolete `UvxBridge` descriptor surface and its helpers:
  - `MCP_PROXY_VERSION`
  - `mcp_proxy_bridge_args`
  - `resolve_uvx_path`
  - `bridge_env_for_uvx`
- Retains `uv_cache_cleanup.gd`; it still protects normal uv/uvx startup paths.
- Resolves the shared attach launch once per aggregate status refresh instead of separately for Claude Desktop and Codex.
- Gives the MCP `client_status` operation the dock’s 30-second aggregate budget. The former 5-second budget was shorter than the allowed 6-second timeout for a single CLI status probe.
- Updates README, configuration, lifecycle, architecture, and troubleshooting documentation.

### Automated

- Ruff: clean
- Python: **1,657 passed / 19 skipped**
- Focused GDScript:
  - `clients`: **144 passed / 7 skipped**
  - `client_attach_config`: **21 passed**
- Full GDScript: **2,068 passed / 21 skipped / 1 failed**

The remaining failure is the pre-existing, unrelated:

`filesystem.test_reimport_mixed_batch_splits_the_report`

It occurs when `EditorFileSystem.update_file()` invalidates `filesystem_handler.gd` during its own active call.

The aggregate `client_status` tool now returns all 22 clients successfully instead of timing out. On the verification machine it completed in approximately 5.27 seconds.

## Live Windows smoke

Tested with Store Claude Desktop **1.24012.9.0**.

The Store-private config was first restored to a verified entry-free backup. Configure was then run through the plugin rather than by hand.

Confirmed:

- the dock wrote the Store-private config;
- existing `coworkUserFilesPath` and `preferences` content survived;
- the conventional roaming file remained byte-identical;
- Claude initialized the generated `godot-ai` entry and completed tool, prompt, and resource discovery;
- Claude cold-started the backend while no editor was attached;
- opening Godot afterward adopted the existing backend;
- closing and reopening Godot left the backend PID unchanged;
- closing and reopening Claude with Godot open also left the backend PID unchanged;
- no client restart was required for the editor restart ordering.

## Summary by CodeRabbit

* **New Features**
  * Added Claude Desktop support through the local `godot-ai attach` connection.
  * Added Windows Store/MSIX configuration discovery and improved platform-specific path handling.
  * Added URL fallback support where available.
  * Preserved existing client settings during configuration updates.

* **Bug Fixes**
  * Improved configuration migration, validation, and diagnostics for missing or ambiguous paths.
  * Reduced repeated client status lookups and added a dedicated status timeout.
  * Improved port-conflict guidance for attach-mode connections.

* **Documentation**
  * Updated setup, troubleshooting, lifecycle, and port-configuration guidance for Claude Desktop and Codex.


dsarno:
👏👏👏🙏🙏

dsarno:
One more Fable review coming.

plugin/addons/godot_ai/clients/_base.gd
```diff
@@ -168,10 +184,50 @@ var toml_body_template: PackedStringArray = PackedStringArray()
...
+		if group.is_empty():
+			continue
+		var path := String(group[0])
+		if FileAccess.file_exists(path):
+			return {"path": path, "error": ""}
+		# A wildcard only resolves when its package directory exists. Treat that
+		# installation evidence as authoritative and create its private config
+		# directly instead of relying on copy-on-write read-through.
+		if template.contains("*"):
+			return {"path": path, "error": ""}
```
dsarno: **Question (edge case): creating the Store-private config from scratch can shadow existing roaming entries.**

When the wildcard package root exists but the private leaf doesn't, this branch picks the private path as the create target, and `_read_or_init` then starts from `{}`. Consider a user who had a populated `%APPDATA%\Claude\claude_desktop_config.json` (say, with other `mcpServers`) and then installed Store Claude but the app hasn't yet materialized its private copy: before Configure, Claude sees the roaming entries via MSIX read-through; after Configure creates the private file with only the `godot-ai` entry, the private copy wins and those other servers silently vanish from Claude's view. `test_ordered_config_candidates_choose_effective_existing_file` pins this outcome, so I assume it's deliberate — but would it be safer to seed a freshly created wildcard-authoritative config from the first later candidate whose file exists (a one-time copy of the roaming content)? That still honors "Configure never writes both" since roaming stays untouched, and there's no in-container content to conflict with when the leaf doesn't exist yet. If the window is considered too narrow to matter (Claude appears to create the private file on first run), a sentence in `docs/client-configuration.md` documenting the trade-off would help the next reader.

---

plugin/addons/godot_ai/client_configurator.gd
```diff
@@ -415,13 +444,69 @@ static func client_status_probe_snapshot(id: String) -> Dictionary:
...
+			str(probe.get("cli_path", "")),
+			launch_context,
+			resolved_launch,
+		)
+		var status = details.get("status", Client.Status.NOT_CONFIGURED)
+		clients.append({
+			"id": client_id,
+			"display_name": client_display_name(client_id),
+			"status": Client.status_label(status),
+			"installed": bool(probe.get("installed", false)),
```
dsarno: **Suggestion: surface `error_msg` in the MCP `client_status` payload.**

A key point of this PR is that path ambiguity (and launch-discovery failure, corrupt config files, etc.) now reaches callers as `Status.ERROR` with an actionable message — but the sweep keeps only `status`, so an MCP consumer sees `"status": "error"` with no way to learn why. The dock gets the message through its own worker; the `client_status` tool does not. Since `details` already carries it, including it would make the tool self-diagnosing:

```gdscript
		var entry := {
			"id": client_id,
			"display_name": client_display_name(client_id),
			"status": Client.status_label(status),
			"installed": bool(probe.get("installed", false)),
		}
		var error_msg := str(details.get("error_msg", ""))
		if not error_msg.is_empty():
			entry["error"] = error_msg
		clients.append(entry)
```

---

plugin/addons/godot_ai/clients/_base.gd
```diff
@@ -168,10 +184,50 @@ var toml_body_template: PackedStringArray = PackedStringArray()
...
+	var fallback_create_path := ""
+	for template_variant in templates:
+		var template := str(template_variant)
+		var group := McpPathTemplate.expand_path_candidates(template)
+		if group.size() > 1:
+			var message := (
+				"%s has multiple matching config package paths for %s: %s. "
+				+ "Remove the stale package installation or edit the intended config manually."
+			) % [display_name, template, ", ".join(group)]
+			push_warning(message)
```
dsarno: **Minor: this warning repeats on every resolution while the ambiguous state persists.**

`resolved_config_path_details()` runs several times per status refresh (the configurator's pre-check, the strategy's own check, plus `is_installed()`'s `resolved_config_path()` fallback), and refreshes recur on dock activity — so a user with two `Claude_*` package roots will accumulate identical warnings in the editor's output for as long as the condition exists. Since the same message already reaches the user as `Status.ERROR` on the dock row (and, per my other comment, ideally the MCP payload), consider de-duplicating — e.g. a `static var` remembering the last warned message, or warning only from the configure/remove paths rather than every resolution.

---

plugin/addons/godot_ai/handlers/client_handler.gd
```diff
@@ -30,14 +42,64 @@ func remove_client(params: Dictionary) -> Dictionary:
...
+	var worker := Thread.new()
+	var start_error := worker.start(worker_callable)
+	if start_error != OK:
+		connection.send_deferred_response(request_id, ErrorCodes.make(
+			ErrorCodes.INTERNAL_ERROR,
+			"Could not start client status worker (error %d)." % start_error,
+		))
+		return
+	while worker.is_alive():
+		await tree.process_frame
```
dsarno, line 106: **Minor: the sweep thread is never joined if the tree tears down mid-run.**

If the editor exits (or the plugin reloads for self-update) while the worker is alive, `await tree.process_frame` stops resuming, the suspended coroutine is dropped, and this `Thread` is destroyed without `wait_to_finish()` — Godot logs the "Thread destroyed without wait_to_finish" error at shutdown. The probes are wall-clock bounded so the window is small, but this is exactly the class of shutdown noise `mcp_dock.gd` engineers against with `_drain_client_status_refresh_workers()` / `_orphaned_client_status_refresh_threads`. Worth either registering these workers somewhere an exit path can drain them, or a short comment stating the leak-at-exit is accepted and why.

---

plugin/addons/godot_ai/clients/hermes.gd
```diff
@@ -32,7 +32,6 @@ func _init() -> void:
 	entry_initial_fields = {}
 
 	# No UVX bridge needed - Hermes is HTTP-native.
```
dsarno: **Nit:** with the assignment gone, this comment now refers to a concept (`UvxBridge`) that no longer exists anywhere in the codebase — it can be dropped along with the line below it, or reworded to "Hermes is HTTP-native; no command shape needed."

---

dsarno (commented):
Reviewed both commits (`6955df5` + `781dee6`) against the diff, the surrounding code on `main`, and the repo's client-configuration docs.

## Overview

This migrates Claude Desktop off `mcp-proxy` onto the client-owned `godot-ai attach` transport (the FLAT JSON slice of #838), adds ordered wildcard `config_path_candidates` so Store/MSIX Claude gets its virtualized private config written directly, and — in the second commit — moves the aggregate `client_status` MCP command onto a deferred background worker with an aligned 30-second budget.

## What looks solid

- **Fail-closed design is consistent**: ambiguous `Claude_*` package roots, unsupported command shapes, and missing launchers all error without touching the config file, and the byte-identical-on-failure guarantee is pinned by tests.
- **Lossless migration semantics**: deep-copy of existing entries, targeted `command_legacy_keys` / `command_env_legacy_keys` removal, preservation of unknown future fields — all tested, including the emptied-`env` omission case.
- **The wildcard segment matcher** correctly handles the prefix/suffix overlap trap (`Claude_` vs `Claude_`+`Claude_`), Windows case-insensitivity, and single-`*` enforcement, with tests for each.
- **Cross-language fixture coverage** (`attach_json_sample.json` verified by both the GDScript renderer and the Python attach argument parser, with a pinned `plugin_version` so releases don't churn it) is a nice guard for the quoted-path bootstrap shape.
- **The timeout story is coherent**: dock worker 30s, dispatcher deferred budget 30000 ms, Python client 30.0 s, single 6 s CLI probe — each pinned by a test so they can't drift apart silently.
- The second commit's deferred sweep correctly reuses the established `DEFERRED_RESPONSE` machinery (late responses are dropped by `connection.gd`), keeps probes off the handler thread, and `McpCliFinder`/`env_lookup`/the launch cache are all already mutex- or snapshot-guarded for the new worker-thread call sites. It also fixes two things I'd flagged while reading the first commit (empty `cli_path` for CLI clients in the aggregate, and `_manual_command` reaching into a private strategy helper).

## Findings

Nothing I'd block on — one design question and a few small improvements, inline:

1. **Question (edge case)** — creating the Store-private config from `{}` can shadow other `mcpServers` a user had in the conventional roaming file that Claude was, until then, seeing via MSIX read-through (`_base.gd`).
2. **Suggestion** — `run_client_status_sweep` drops `details.error_msg`, so the MCP `client_status` tool reports `"error"` with no reason while the dock gets the actionable message (`client_configurator.gd`).
3. **Minor** — the ambiguity `push_warning` fires on every resolution, several times per refresh, for as long as the state persists (`_base.gd`).
4. **Minor** — the deferred sweep `Thread` is never joined if the editor exits or the plugin reloads mid-run (`client_handler.gd`).
5. **Nit** — stale `UvxBridge` comment left behind in `hermes.gd`.

---

plugin/addons/godot_ai/clients/_base.gd
Clubhouse1661: Fixed in `cd40fca`. When a unique Store package exists but its private config does not, resolution now returns the first later existing candidate as a one-time `seed_path`. Configure reads and validates that complete config, merges the `godot-ai` entry, and writes only the authoritative Store-private target; the roaming source remains byte-identical.

I added regression coverage for preserving another MCP server and unrelated preferences, plus the failure case where an invalid roaming seed must not create the private file or modify the source. The behavior is also documented in `client-configuration.md`.

plugin/addons/godot_ai/client_configurator.gd
Clubhouse1661: Fixed in `cd40fca`. The aggregate sweep now includes an optional `error` field whenever `details.error_msg` is non-empty, while healthy rows retain the existing payload shape without an empty error field.

I extracted row construction into `_client_status_sweep_entry` and added coverage for both actionable-error and healthy responses. The documented `client_status` schema now includes `error?` as well.

plugin/addons/godot_ai/clients/_base.gd
Clubhouse1661: Fixed in `cd40fca`. Path-resolution warnings are now deduplicated per client descriptor, with mutex protection because resolution can run from status workers. The same actionable error is still returned on every call, but the editor warning is emitted only once while that ambiguity persists.

The warning key is cleared when resolution succeeds or the condition disappears, so a later or changed ambiguity can still produce a new warning.

plugin/addons/godot_ai/handlers/client_handler.gd
Clubhouse1661: Fixed in `cd40fca`. The handler now tracks every started status worker and exposes `prepare_for_teardown()`, which stops new status work and joins all tracked threads before the handler is released.

`McpDispatcher.clear()` invokes that teardown hook before clearing lazy handlers, and the self-update path now clears the dispatcher before plugin scripts are replaced. This covers ordinary editor/plugin teardown and self-update reload. I added handler and dispatcher lifecycle tests and ran a live concurrent status/reload smoke; reload waited for the worker and completed without an orphan-thread error.

plugin/addons/godot_ai/clients/hermes.gd
Clubhouse1661: Fixed in `cd40fca`. Reworded it to:

> Hermes is HTTP-native; no command shape is needed.

dsarno:
👍🏼👍🏼☑️

Issue #507, opened 2026-06-01
Surface the --allow-host LAN opt-in in the dock UI and manual command (follow-up to #421)


dsarno:
## Background

#421 (PR #505 / integrated in #506) landed the **server-side core** of the LAN opt-in: a `--allow-host CIDR` flag that binds both transports off loopback and widens the DNS-rebinding guard's Host allowlist to named CIDRs (Origin / Sec-Fetch-Site checks stay on). That's the functional, security-critical gate, and it's fully tested.

Two **additive UI/UX** pieces from the original #421 proposal were intentionally deferred — they make the opt-in *discoverable and usable* from the editor but are not security boundaries. Capturing them here so they aren't lost when #421 auto-closes.

### 1. "Allow remote hosts (CIDR)" field — in the Settings tab
The dock's **"Clients & Tools" secondary window** is a `TabContainer` (currently **Clients** + **Tools** tabs; `mcp_dock.gd:610-647` notes a third tab is just one more `_build_*_tab` + `set_tab_title`). This control is a setting, so it belongs there with the other settings-style controls — **not floating in the main dock**.

- [ ] Add a **Settings tab** to that window (new `_build_settings_tab`, mirroring `_build_tools_tab`), and house the allow-host field there alongside the other developer/config controls (developer-mode, port pickers, etc.). If it's cleaner to fold this into the existing **Tools** tab renamed/reframed as **Settings**, that's fine too — the point is it lives under the Clients/Settings tabbed window, grouped with settings.
- [ ] **Developer-mode-gated** text field accepting one or more CIDRs (matching `--allow-host`: CIDR/bare IP, comma-separated/repeatable).
- [ ] A **warning banner** next to it explaining the DNS-rebinding risk and recommending an SSH tunnel / Tailscale on untrusted networks; make the user name the network they trust rather than offer a blanket "expose everything" toggle.
- [ ] Persist via `EditorSetting`, and thread it into the server launch — the plugin spawns the Python server, so pass `--allow-host <value>` on spawn when the setting is non-empty (`server_lifecycle.gd`).

### 2. Surface the LAN URL in `manual_command`
- [ ] When a non-loopback allowlist is active, the configurator's `manual_command` (the dock's "Run this manually" panel, also shown per-client in the Clients tab) should surface the **LAN URL** (e.g. `http://192.168.1.50:8000/mcp`) instead of / in addition to the loopback URL, so the user can copy-paste the right address into a remote agent.

## References
[...]

PR #853, opened 2026-08-04
feat: Vision Routing - route screenshot tool through vision APIs for text-only models

michaltomczykowski:
## Summary

Adds **Vision Routing**: an opt-in way for text-only AI models (e.g. DeepSeek) to "see" the editor and the running game. When enabled, every single-image `editor_screenshot` capture is sent to a vision model on a worker thread, and the resulting **text description** is returned to the AI instead of the raw image (`vision_description` + `note`; the server drops the image block for routed captures, and `include_image=false` captures carry the description in metadata). On any failure (missing key/model, network error, API error) the original image passes through unchanged, with a short "Vision routing unavailable (provider): reason" note - the screenshot tool never breaks.

## What's included

- **Vision Routing section** inside **Clients & Tools -> Settings** (no dock row): enable checkbox, provider dropdown, per-provider API key field (masked, with show/hide), per-provider model-id field, **Test connection** button and status line.
- **Providers are curated; the model id is required user input.** The `PROVIDERS` table fixes what is repo-owned - label, API dialect (OpenAI chat-completions vs Gemini `generateContent`), endpoint shape, environment variable and encrypted key slot. The model id is entered per provider (stored in Editor Settings next to the key), because third-party model ids retire and only the account holder gets notified; date-stamped suggestions live in the UI placeholder and `docs/vision-routing.md`:
  - Groq - `qwen/qwen3.6-27b` (free tier)
  - Google Gemini - `gemini-flash-latest` (free tier, AI Studio key)
  - xAI Grok - `grok-4.5` (paid)
  - Switching providers switches to that provider's entered model - the software never guesses a model name.
- **Keys never stored in plain text**: AES-256-CBC encrypted in Editor Settings with a machine-derived key (domain-tagged AES/MAC keys; the IV is inside the MAC; storage is refused when the machine reports no unique id). Local obfuscation, documented as such. `GROQ_API_KEY` / `GOOGLE_API_KEY` / `XAI_API_KEY` env vars take priority over stored keys.
- **Test connection validates vision, not just auth**: it sends the same image-bearing request as real routing (placeholder image), so one click checks key + model existence + image capability; it reports the key source (`stored key` / `via ENV_VAR`) and disables the button while testing.
- **Bounded worker exchange**: one absolute 10s budget for connect + request + body, a 1 MiB response-body cap, and a 13s deferred ledger inside the server's 15s `editor_screenshot` window, so a hung provider surfaces as a clean plugin timeout, not a server-side abort.
- **Agent context preserved**: `user_prompt` from the requesting agent is included in the vision prompt for editor captures and game captures (stashed by request id across the deferred game hop).
- `routed_via` reports `provider:<entered model>` verbatim (e.g. `groq:qwen/qwen3.6-27b`).
- **Settings tab rework**: the Settings tab is now usable without Developer mode. The LAN `--allow-host` block moved behind a collapsed "Remote access (advanced)" disclosure (warning text unchanged) that auto-expands when a non-empty CIDR allowlist is configured; Developer mode still gates the log viewer and dev sections.
- Docs: `docs/vision-routing.md` (providers table with date-stamped suggestions, setup, behavior, failure modes).
- Tests: `test_project/tests/test_vision_routing.gd` - 23/23 pass in-editor (encrypted key roundtrip + tamper rejection incl. IV flip, prompt, deferred-reply conversion, per-provider model round-trip, empty-model decline, game-capture stash forwarding, OpenAI/Gemini body builders incl. `generationConfig.maxOutputTokens`, Gemini parser, ping body shape); full project suite green.

## Notes

- Feature is **off by default**; the enabled flag, provider, model ids and keys live in Editor Settings, never committed with a project.
- No API keys are included in this PR.
- Live-verified in the editor: editor + game screenshots routed through real providers (Groq and Gemini) return accurate descriptions; failure paths (missing key/model, provider HTTP errors) pass the original image through with the failure note; Test connection shows both OK and provider-error states.

### DIFF
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -30,7 +30,7 @@ not the MCP tool names.
 | `project_run` | Play the project, then wait briefly for game liveness (autosave persists in-memory MCP edits unless `autosave=False`) |
 | `test_run` | Run GDScript test suites in the editor — see [testing.md](testing.md) for writing suites and the `McpTestSuite` API |
 | `logs_read` | Read plugin / game / editor / combined log buffers. `source="editor"` surfaces parse errors, GDScript reload warnings, @tool/EditorPlugin runtime errors, push_error/push_warning, and visible Debugger dock Errors-tab rows — use this when the editor's Output or Debugger Errors panel shows red/yellow rows |
-| `editor_screenshot` | Capture the editor 3D viewport (`viewport`), editor 2D viewport (`viewport_2d`), cinematic Camera3D render (`cinematic`), or running game framebuffer (`game`) |
+| `editor_screenshot` | Capture the editor 3D viewport (`viewport`), editor 2D viewport (`viewport_2d`), cinematic Camera3D render (`cinematic`), or running game framebuffer (`game`). With Vision Routing enabled, images are replaced by a text description (`vision_description` / `routed_via`) - see [vision-routing.md](vision-routing.md) |
 | `editor_reload_plugin` | Reload the plugin and wait for reconnect (works with external and plugin-managed servers) |
 | `animation_create` | Create an Animation clip (auto-creates AnimationPlayer + library if missing) |
 
--- a/docs/vision-routing.md
+++ b/docs/vision-routing.md
@@ -1,0 +1,83 @@
+# Vision Routing
+
+Vision Routing lets models without image support (e.g. DeepSeek) still "see"
+the editor and the running game. When enabled, every capture made through the
+`editor_screenshot` tool is sent to Groq's free vision API on a worker thread,
+and the resulting **text description** is returned to the AI instead of the raw
+image.
+
+The feature is implemented natively in the plugin (`vision_routing.gd`) and is
+off by default. It is opt-in per editor installation; the enabled flag and the
+API key live in **Editor Settings**, so they are not committed with the
+project.
+
+## Why
+
+Most MCP clients used with godot-ai can consume image blocks. Some cannot
+(DeepSeek and other text-only models reject or ignore them). With routing on,
+the screenshot tool stays useful for those models: the image is replaced by a
+tiny 2x2 placeholder, and the description is delivered as text metadata
+(`vision_description`, plus a `note` that the server forwards into the tool
+result's text block).
+
+## How it works
+
+1. The AI calls `editor_screenshot` (any `source`).
+2. The plugin captures the frame exactly as before.
+3. If routing is enabled, the image is sent to
+   `qwen/qwen3.6-27b` on Groq (free tier) on a background thread - the editor
+   main thread never blocks.
+4. The response is deferred until the description is ready; the payload the
+   model receives contains:
+   - `vision_description`: the Groq description.
+   - `note`: the description (plus any pre-existing note, e.g. stale-frame).
+   - `routed_via`: the model id, so the client knows the capture was routed.
+   - `image_base64`: a 2x2 transparent placeholder PNG (kept so the payload
+     stays well-formed; the server omits the image block for routed captures).
+5. Any failure (missing key, network error, API error) passes the **original
+   image through unchanged** - the screenshot tool never breaks.
+
+The same flow applies to editor viewport / cinematic / 2D captures and to
+running-game captures (`source="game"`, which are intercepted in the debugger
+plugin).
+
+Coverage sweeps (`coverage=true`, multiple images) are passed through
+untouched - routing applies to single-image captures.
+
+## Setup
+
+1. Open the Godot AI dock -> **Clients & Tools** -> **Vision Routing** tab.
+2. Toggle **Enable routing** on.
+3. Paste a Groq API key (free tier: <https://console.groq.com>). The key is
+   stored encrypted (AES-256-CBC, key derived from this machine) in Editor
+   Settings - not plain text, but local obfuscation only. Alternatively set
+   the `GROQ_API_KEY` environment variable; it takes priority over the stored
+   key.
+4. Press **Test connection** to verify the key works.
+
+A quick **Vision Routing** toggle also sits in the dock right under
+**Developer mode**, so you can flip routing on/off without opening the settings
+window (e.g. when you switch from a text-only model to one that analyzes
+images itself).
+
+## Notes
+
+- The Groq call adds ~1-2s to each screenshot while routing is enabled.
+- Screenshots are downscaled to at most 1024px on the longest edge before
+  being sent to Groq.
+- If you stop using a key that was typed into the settings window, consider
+  rotating it in the Groq console - an encrypted-at-rest key is still local
+  obfuscation, not a secret vault.
+- Routing only applies when the connected model is expected to be text-only;
+  keep it off for image-capable models to avoid the extra hop.
+
+## Developer notes
+
+- `plugin/addons/godot_ai/vision_routing.gd` - routing core, Groq worker,
+  encrypted key storage, UI construction.
+- Hook points: `editor_handler.gd::take_screenshot` (editor captures) and
+  `mcp_debugger_plugin.gd::_on_screenshot_response` (game captures).
+- Server-side metadata forwarding: `src/godot_ai/handlers/editor.py`.
+- Tests: `tests/unit/test_runtime_handlers.py` (routed metadata + image-block
+  omission) and `test_project/tests/test_vision_routing.gd` (key storage,
+  prompt, deferred-reply conversion).
--- a/docs/vision-routing.md
+++ b/docs/vision-routing.md
@@ -2,7 +2,7 @@
 
 Vision Routing lets models without image support (e.g. DeepSeek) still "see"
 the editor and the running game. When enabled, every capture made through the
-`editor_screenshot` tool is sent to Groq's free vision API on a worker thread,
+`editor_screenshot` tool is sent to a curated vision model on a worker thread,
 and the resulting **text description** is returned to the AI instead of the raw
 image.
 
@@ -10,6 +10,23 @@ The feature is implemented natively in the plugin (`vision_routing.gd`) and is
 off by default. It is opt-in per editor installation; the enabled flag and the
 API key live in **Editor Settings**, so they are not committed with the
 project.
+
+## Providers
+
+The provider and its model are **curated** - the model id is fixed per provider
+and shown in brackets in the Provider dropdown, so switching providers can
+never silently call the wrong model:
+
+| Provider | Model | Dialect | Free tier |
+| --- | --- | --- | --- |
+| Groq | `qwen/qwen3.6-27b` | OpenAI chat-completions | Yes |
+| Google Gemini | `gemini-2.5-flash` | generateContent (REST) | Yes (AI Studio key) |
+| xAI Grok | `grok-2-vision-latest` | OpenAI chat-completions | No (paid) |
+
+Each provider has its own encrypted key slot and its own environment-variable
+override: `GROQ_API_KEY`, `GOOGLE_API_KEY` and `XAI_API_KEY`. The environment
+variable always takes priority over the stored key. Changing the provider in
+the dropdown switches the key field to that provider's stored key.
 
 ## Why
 
@@ -24,14 +41,15 @@ result's text block).
 
 1. The AI calls `editor_screenshot` (any `source`).
 2. The plugin captures the frame exactly as before.
-3. If routing is enabled, the image is sent to
-   `qwen/qwen3.6-27b` on Groq (free tier) on a background thread - the editor
-   main thread never blocks.
+3. If routing is enabled, the image is sent to the selected provider's model
+   (see [Providers](#providers)) on a background thread - the editor main
+   thread never blocks.
 4. The response is deferred until the description is ready; the payload the
    model receives contains:
-   - `vision_description`: the Groq description.
+   - `vision_description`: the text description.
    - `note`: the description (plus any pre-existing note, e.g. stale-frame).
-   - `routed_via`: the model id, so the client knows the capture was routed.
+   - `routed_via`: `provider:model` (e.g. `groq:qwen/qwen3.6-27b`), so the
+     client knows the capture was routed and by which model.
    - `image_base64`: a 2x2 transparent placeholder PNG (kept so the payload
      stays well-formed; the server omits the image block for routed captures).
 5. Any failure (missing key, network error, API error) passes the **original
@@ -48,12 +66,15 @@ untouched - routing applies to single-image captures.
 
 1. Open the Godot AI dock -> **Clients & Tools** -> **Vision Routing** tab.
 2. Toggle **Enable routing** on.
-3. Paste a Groq API key (free tier: <https://console.groq.com>). The key is
-   stored encrypted (AES-256-CBC, key derived from this machine) in Editor
-   Settings - not plain text, but local obfuscation only. Alternatively set
-   the `GROQ_API_KEY` environment variable; it takes priority over the stored
-   key.
-4. Press **Test connection** to verify the key works.
+3. Pick a provider in the **Provider** dropdown (the model it will use is
+   shown in brackets, e.g. `Google Gemini (gemini-2.5-flash)`).
+4. Paste the key for that provider (Groq: <https://console.groq.com>, Google:
+   <https://aistudio.google.com>, xAI: <https://console.x.ai>). Each provider
+   keeps its own key slot. Keys are stored encrypted (AES-256-CBC, key derived
+   from this machine) in Editor Settings - not plain text, but local
+   obfuscation only. Alternatively set `GROQ_API_KEY`, `GOOGLE_API_KEY` or
+   `XAI_API_KEY`; the environment variable takes priority over the stored key.
+5. Press **Test connection** to verify the key works.
 
 A quick **Vision Routing** toggle also sits in the dock right under
 **Developer mode**, so you can flip routing on/off without opening the settings
@@ -62,19 +83,21 @@ images itself).
 
 ## Notes
 
-- The Groq call adds ~1-2s to each screenshot while routing is enabled.
+- The vision call adds ~1-2s to each screenshot while routing is enabled.
 - Screenshots are downscaled to at most 1024px on the longest edge before
-  being sent to Groq.
+  being sent to the provider.
 - If you stop using a key that was typed into the settings window, consider
-  rotating it in the Groq console - an encrypted-at-rest key is still local
-  obfuscation, not a secret vault.
+  rotating it in the provider console - an encrypted-at-rest key is still
+  local obfuscation, not a secret vault.
 - Routing only applies when the connected model is expected to be text-only;
   keep it off for image-capable models to avoid the extra hop.
 
 ## Developer notes
 
-- `plugin/addons/godot_ai/vision_routing.gd` - routing core, Groq worker,
-  encrypted key storage, UI construction.
+- `plugin/addons/godot_ai/vision_routing.gd` - routing core, per-provider
+  workers (OpenAI chat-completions + Gemini generateContent dialects),
+  encrypted key storage, UI construction. The `PROVIDERS` constant is the
+  single place where provider/model pairs are curated.
 - Hook points: `editor_handler.gd::take_screenshot` (editor captures) and
   `mcp_debugger_plugin.gd::_on_screenshot_response` (game captures).
 - Server-side metadata forwarding: `src/godot_ai/handlers/editor.py`.
--- a/docs/vision-routing.md
+++ b/docs/vision-routing.md
@@ -1,7 +1,7 @@
 # Vision Routing
 
 Vision Routing lets models without image support (e.g. DeepSeek) still "see"
-the editor and the running game. When enabled, every capture made through the
+the editor and the running game. When enabled, every single-image capture made through the
 `editor_screenshot` tool is sent to a curated vision model on a worker thread,
 and the resulting **text description** is returned to the AI instead of the raw
 image.
@@ -21,7 +21,7 @@ never silently call the wrong model:
 | --- | --- | --- | --- |
 | Groq | `qwen/qwen3.6-27b` | OpenAI chat-completions | Yes |
 | Google Gemini | `gemini-2.5-flash` | generateContent (REST) | Yes (AI Studio key) |
-| xAI Grok | `grok-2-vision-latest` | OpenAI chat-completions | No (paid) |
+| xAI Grok | `grok-4.5` | OpenAI chat-completions | No (paid) |
 
 Each provider has its own encrypted key slot and its own environment-variable
 override: `GROQ_API_KEY`, `GOOGLE_API_KEY` and `XAI_API_KEY`. The environment
--- a/docs/vision-routing.md
+++ b/docs/vision-routing.md
@@ -13,20 +13,28 @@ project.
 
 ## Providers
 
-The provider and its model are **curated** - the model id is fixed per provider
-and shown in brackets in the Provider dropdown, so switching providers can
-never silently call the wrong model:
+The provider itself is **curated** - label, API dialect, endpoint shape,
+environment variable and encrypted key slot are fixed in the `PROVIDERS`
+table. The **model id is not**: third-party model ids retire, and only the
+account holder gets notified when one does, so the model is a required
+per-provider field you enter next to the key. Switching providers switches to
+that provider's entered model - the software never guesses a model name.
 
-| Provider | Model | Dialect | Free tier |
+Suggestions that were verified working **as of 2026-08** (always check your
+provider console for current ids; a provider's model *listing* API still
+advertises retired ids - only a real call proves serviceability):
+
+| Provider | Dialect | Suggested model | Free tier |
 | --- | --- | --- | --- |
-| Groq | `qwen/qwen3.6-27b` | OpenAI chat-completions | Yes |
-| Google Gemini | `gemini-2.5-flash` | generateContent (REST) | Yes (AI Studio key) |
-| xAI Grok | `grok-4.5` | OpenAI chat-completions | No (paid) |
+| Groq | OpenAI chat-completions | `qwen/qwen3.6-27b` | Yes |
+| Google Gemini | generateContent (REST) | `gemini-flash-latest` (or `gemini-3-flash-preview`) | Yes (AI Studio key) |
+| xAI Grok | OpenAI chat-completions | `grok-4.5` | No (paid) |
 
-Each provider has its own encrypted key slot and its own environment-variable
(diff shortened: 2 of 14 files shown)


dsarno (commented):
Review of head `50eec29`, with the CodeRabbit/Copilot feedback cross-checked against the code rather than repeated. I read the full diff, ran the Python suite, ruff, and the previously failing guard tests against this head locally, and pulled the CI logs for the red jobs.

## Overview

Opt-in Vision Routing: routed `editor_screenshot` captures are described by a curated vision provider on a worker thread and returned as text (`vision_description` / `routed_via`), with the server dropping the image block for routed captures. The architecture is right for this codebase: worker threads keep the main thread unblocked (per AGENTS.md), replies ride the existing deferred-response ledger, the provider/model table is curated, the feature is off by default, and every failure path falls back to the original image. Docs and tests ship with it. Nice work overall — the items below are what I'd want resolved before merge.

## CI: was red at `f04b984`, fixed by `50eec29` — but the fix duplicates the dispatch

All four Python jobs failed at `f04b984`: splitting `take_screenshot` into a wrapper + `_take_screenshot_impl` broke the 4 guard tests in `tests/unit/test_docs_screenshot_sources.py`, which parse the `take_screenshot` block for the `match source:` dispatch and the invalid-source message. `50eec29` restores the contract — verified locally on this head: the 4 guard tests pass, the 24 screenshot handler tests pass, `ruff check` is clean, and the full `tests/unit` suite is green (1328 passed). CI hasn't reported on the new head yet.

However, the fix works by duplicating the dispatch: two `match source:` blocks and two identical invalid-source strings now live in the same file (`editor_handler.gd:429`/`439` in the wrapper, `:459`/`:527` in the impl), and the guard tests pin only the wrapper's copy. The two vocabularies can drift silently — a source added to the impl but not the wrapper gets rejected before the impl ever sees it, and no test notices the impl's copy. Prefer a single source of truth: validate only in the wrapper (the impl's `match` keeps its real dispatch arms; its `_:` arm becomes an internal-error guard), or point the guard test's `get_func_block` at `_take_screenshot_impl`.

## Existing bot feedback — what I verified as real (none addressed as of `50eec29`)

Must-fix, in rough priority order:

1. **`_error_notes` is mutated from worker threads and main thread without synchronization** (`vision_routing.gd:106`) — confirmed. `_parse_description` writes it on the worker; `_on_route_complete`/`shutdown()` read/erase/clear on the main thread, and two routes can be in flight at once (editor + game). Godot Dictionaries give no guarantee under concurrent structural mutation. CodeRabbit's ownership fix (worker returns `{description, error}` through `call_deferred`) is the right shape and deletes the shared state instead of locking it.
2. **`shutdown()` only joins alive threads** — confirmed. A finished-but-unjoined worker (its deferred `_on_route_complete` dropped during teardown — exactly the plugin-reload window) is released by `_threads.clear()` without `wait_to_finish()`, producing the "Thread object destroyed without its completion having been realized" error. Drop the `is_alive()` gate; joining a finished thread is free.
3. **Second "Test connection" click clobbers a live ping thread** (`_threads["_test"] = thread`) — confirmed; a ping can run ~12s and the button stays enabled the whole time. Guard on the existing thread, and ideally disable the button while testing (doubles as UX feedback).
4. **Body-read loop ignores `_active`** — confirmed; teardown can block up to `REQUEST_TIMEOUT_MS` in `wait_to_finish()`, contradicting the ~one-poll-interval comment above `shutdown()`.
5. **Unguarded `EditorSettings.get_setting` reads** — confirmed, and it's also against this repo's own convention: everything else registers-on-first-read or guards (`mcp_dock.gd:1453-1456` dev mode, `plugin.gd:1623-1636` managed-server settings). `is_routing_enabled()` runs on every `take_screenshot` and every game screenshot response, so a fresh install logs a warning per capture into the editor log that agents read via `logs_read(source="editor")`. Follow the `DEV_MODE_SETTING` pattern.
6. **`vision_routing = vision_routing` self-assignment in `McpDebuggerPlugin._init`** — confirmed, with one nuance worth recording: production wiring works today because `plugin.gd` assigns the property directly (`_debugger_plugin.vision_routing = _vision_routing`), so this is a latent contract bug, not a live one. The constructor parameter silently drops the dependency for any caller that uses it. Either `self.vision_routing = vision_routing` or remove the parameter.
7. **`grok-2-vision-latest`** — I can't independently re-verify retirement from this environment, but CodeRabbit's web-sourced claim is consistent with xAI's grok-2 deprecation wave. Since the curated table is the whole point of the provider design, please verify against current xAI docs before merge — or ship Groq + Gemini now and add xAI once confirmed. A dead model id isn't fatal (fallback returns the image) but it ships a provider option that can never work, plus a misleading "paid" table row.

The nitpicks are worth taking too, especially: check `Thread.start()`'s return; persist the key on `text_submitted`/`focus_exited` rather than every keystroke (also stops writing partially-typed keys to disk); use the documented `provider:model` format in the Python test stubs; add the per-provider `routed_via` test; fix the two "described by Groq" comments; scope the docs intro to single-image captures.

## New findings

### Security

- **Key derivation is weaker than intended off-Windows** (`vision_routing.gd:609-611`). `USERNAME`/`USERPROFILE` are Windows-only, so on Linux/macOS the key reduces to SHA-256(machine-id | "" | "" | OS name | fixed salt) — identical for every user on a shared machine. Worse, `OS.get_unique_id()` can return an empty string on some platforms, at which point the key is a constant derivable from this public source file. Cheap fixes: also mix in `USER`/`HOME`, and refuse to store (env-var-only, with a status hint) when `get_unique_id()` is empty. The docs honestly frame this as local obfuscation, so severity is low — but these are one-line improvements.
- **Extending Copilot's HMAC finding** (`vision_routing.gd:636`): besides including the IV in the MAC, derive separate AES and MAC keys (one extra hash each with a domain tag) instead of reusing the same 32 bytes for both. And add an IV-flip case to `test_decrypt_rejects_tampered_blob` — the current test only flips ciphertext, so it keeps passing while the IV gap exists.
- **Label the injected description's provenance.** The description is third-party text that rides `note` into the driving agent's context verbatim; a prompt-injected or compromised provider response reads as if the plugin said it. One line in `_on_route_complete` — e.g. `"Vision description (%s): %s" % [routed_via, description]` — makes the note self-attributing, including for clients that ignore `routed_via`.

Things that check out: TLS verified via `TLSOptions.client()`; keys travel in headers (`x-goog-api-key` / `Authorization`), never URLs; env vars take priority; the key field is masked; keys never appear in logs; no new network listener, so the loopback WS trust boundary is untouched; feature and keys are per-editor-install, never committed with a project.

### UX / robustness

- **Timeout budgets don't nest on the editor path — the "screenshot tool never breaks" promise doesn't hold under a slow provider.** The server aborts editor captures at `SCREENSHOT_TIMEOUT_SEC = 15s` (`src/godot_ai/handlers/editor.py:26`), but a routed capture can legitimately take ~20s (`CONNECT_TIMEOUT_MS` 4s + `REQUEST_TIMEOUT_MS` 8s request + 8s body, `vision_routing.gd:92-93`), and the plugin claims a 30s deferred window (`vision_routing.gd:165`). A rate-limited free-tier provider that answers in 16s — or hangs — yields a server-side timeout error with **no image at all**; the fallback pass-through fires after the server already gave up. The game path is fine (35s server / 30s ledger). Fix: give the whole worker exchange one total budget of ~10-12s and set `_deferred_timeout_ms` below 15s so the fallback still lands inside the server window.
- **`include_image=False` still pays for routing.** `include_image` is consumed server-side, so a metadata-only capture with routing enabled still burns a vision API call and its latency. For a text-only model that's arguably the feature working (it gets `vision_description` in the metadata); for an image-capable client using metadata-only calls it's pure cost. Consider forwarding `include_image` in the `take_screenshot` params as a routing hint — or at least note the behavior in `docs/vision-routing.md`.
- **Test connection tests the resolved key, not necessarily the field.** With an env var set, "Test connection" validates the env key while the user is editing a (possibly wrong) stored key. The hint text explains the priority, but the status line is where the eyes are — say which source was tested, e.g. "OK — Groq responded (via GROQ_API_KEY)".
- Micro: `_sync_ui_states` assigns `button_pressed`, which re-emits `toggled` and re-runs `_on_enable_toggled` (a harmless but pointless second settings write) — `set_pressed_no_signal()` is the idiom for programmatic sync.

### Code quality (minor)

- Dead hook: `_build_prompt` reads `params["user_prompt"]` / `"_user_prompt"` (`vision_routing.gd:252-256`), but nothing server-side ever passes one — the tool schema has no such param and stray kwargs are stripped by middleware. Wire a real param through the tool or drop the branch.
- `MODEL_ID` is described as a "backwards-compatible alias" in a brand-new file — it's just the Groq model id plus a test convenience; say that.
- `_on_route_complete` does `data.erase("image_base64")` and then unconditionally sets `data["image_base64"] = _PLACEHOLDER_PNG` a few lines later — drop the erase.
- `vision_routing.gd` and `docs/vision-routing.md` are missing trailing newlines.

## Verdict

Solid, well-documented opt-in feature with genuinely good fallback instincts and the right threading architecture. Before merge I'd want: the thread-lifecycle batch (items 1–4), the settings guards (5), the constructor fix (6), the editor-path timeout budget, and a decision on the xAI model id — plus de-duplicating the `take_screenshot` dispatch that the CI fix introduced. The rest is polish.

---

michaltomczykowski:
Thanks for the review! Commit \9fd4a4b\ addresses the CodeRabbit findings:

- **Thread lifecycle**: \shutdown()\ now joins every *started* thread (not only alive ones); \_start_route\ checks the \Thread.start()\ result and passes the screenshot through unchanged if the worker can't start; the Test-connection button rejects a second click while a ping is in flight.
- **Worker/main-thread race (critical)**: workers now return \{desc: ..., error: ...}\ as the deferred result instead of mutating a shared \_error_notes\ dictionary; \_error_notes\ is gone entirely.
- **Teardown**: the HTTP body-read loop now honors \_active\, so shutdown is not blocked by a request timeout.
- **Settings**: all \EditorSettings\ reads are guarded with \has_setting()\ (fresh installs no longer emit errors).
- **Key storage**: API keys are persisted on commit (Enter / focus loss) instead of per keystroke; the blob HMAC now authenticates IV + ciphertext.
- **Providers**: xAI Grok model updated to \grok-4.5\; Groq \qwen/qwen3.6-27b\ re-verified as the current vision model (live-tested against the API).
- **Bug fix**: \McpDebuggerPlugin._init\ now assigns \self.vision_routing\ (was a no-op self-assignment).
- **Tests/docs**: \
outed_via\ tests assert the documented \provider:model\ format on both response paths; added \	est_route_complete_reports_active_provider_in_routed_via\; docs scope routing to single-image captures.

Local verification: \pytest tests/unit\ 1315 passed / 16 skipped; Godot test suite 2114 passed (vision_routing 16/16); live routed screenshot through the editor plugin returns the Groq description. The new CI run is awaiting approval.

michaltomczykowski:
Follow-up commit \646035c\ addresses the second review round (cross-checked against the code, not repeated):

**Architecture**
- De-duplicated the \	ake_screenshot\ dispatch: the wrapper is thin again; the single \match source:\ block + invalid-source message live in \_take_screenshot_impl\, and the guard tests in \	est_docs_screenshot_sources.py\ now parse the impl, so the vocabularies can't drift.

**Security**
- Key derivation now uses domain-tagged AES/MAC keys (no shared 32 bytes), mixes in \USER\/\HOME\, and refuses to store a key when the machine reports no unique id (falls back to env-var-only with a UI hint).
- The injected description is now self-attributing in \
ote\: \Vision description (provider:model): ...\.
- Added an IV-flip tamper test (the IV is inside the MAC, so decryption must fail).

**Robustness / UX**
- Timeout budgets now nest: the whole worker exchange is capped at \TOTAL_ROUTE_BUDGET_MS\ (10s) and the editor deferred ledger is 13s, so the fallback pass-through lands inside the server's 15s \editor_screenshot\ window instead of the server aborting first.
- \user_prompt\ is now a real \editor_screenshot\ tool parameter, forwarded to the plugin and sent to the vision model as context (the old dead \_user_prompt\ read is gone).
- Test connection reports which key source was tested (\OK - Groq responded (via GROQ_API_KEY).\ vs \(stored key)\) and disables the button while a ping is in flight.
- \_sync_ui_states\ uses \set_pressed_no_signal()\ so programmatic sync doesn't re-emit \	oggled\.
- Docs note that \include_image=false\ captures still route (that is the point for text-only models); trailing newlines added.

**Verification**: \pytest tests/unit\ 1317 passed / 16 skipped, ruff clean; Godot suite 2116 passed (vision_routing 18/18); live routed screenshot returns the labeled description. Both new CI runs are awaiting approval.

dsarno (commented):
Follow-up on `646035c`: I re-verified both fix commits against the code (not just the commit messages), and the second round checks out.

## Round 2 verification

Every item from my review is genuinely addressed, each with a regression test: the total route budget is a single absolute 10s deadline threaded through all three HTTP loops with the deferred ledger at 13s (inside the server's 15s window); the dispatch de-dup went the clean way (thin wrapper, single `match source:` in `_take_screenshot_impl`, guard tests repointed); key derivation now uses domain-tagged AES/MAC keys, mixes in `USER`/`HOME`, and refuses to store when the machine reports no unique id; the IV is inside the MAC with an IV-flip tamper test; the note is self-attributing; `user_prompt` is a real tool parameter; test-connection reports the key source and disables the button; `set_pressed_no_signal` in `_sync_ui_states`. The round-1 items (thread lifecycle, `_error_notes` removal, `has_setting` guards, `self.vision_routing`, `grok-4.5`) all check out too.

Independent verification on this head: `ruff check` clean, `pytest tests/unit` 1330 passed / 3 skipped, and CI is green across all 23 checks.

## Triage of CodeRabbit's two newest rounds

Three findings are valid and worth one small follow-up commit:

1. **Ping-path `Thread.start()` return is still unchecked** (`_on_test_connection`). `_start_route` now checks it, but if the ping thread fails to start, `_tab_status` sticks at "Testing..." and — since `646035c` — the Test connection button stays disabled forever. Capture the result, clear `_threads["_test"]`, restore the status, and re-enable the button on failure.
2. **Cap the response body size** in `_http_post_json` (e.g. 1 MiB). The deadline bounds time but not bytes; a faulty or hostile provider can stream a large body inside the window. Cheap guard, worth having on a socket that talks to third-party hosts.
3. **`user_prompt` doesn't reach game captures** — `route_game_payload` passes `{}` to `_start_route`, so `_build_prompt` loses the caller context for `source="game"` (I'd flagged this as a residual too). Stash the params by request id when the editor handler defers the game capture, and consume them in `route_game_payload`. Alternatively, document that `user_prompt` applies to editor captures only and tighten the docs sentence that currently implies any source.

Two findings I'd push back on, with the receipts:

- **"Preserve the envelope and stamp `readiness`"** — not applicable. `McpConnection.send_deferred_response` already stamps `status`, `readiness`, and `error_watermark` centrally on every deferred reply that lacks them (`connection.gd:672-682`), which is exactly how the pre-existing game screenshot path (`_on_screenshot_response` sending bare `{"data": payload}`) has always worked. Both routing call sites construct `{"data": ...}` payloads with no other top-level keys, so there is nothing to lose and nothing extra to stamp.
- **"Move routing to a `_process()` state machine"** — this inverts the guideline it cites. "All GDScript executes in `_process()` with a 4ms frame budget. Never block" is about the *main thread*; the worker `Thread` + `call_deferred` design exists precisely to keep blocking HTTP (TLS handshake, polling, body reads) off it. Rewriting the exchange as an in-frame state machine would move that work *into* the frame budget. The current architecture is the right one for this codebase and should stay.

With the three small items above, this is in merge shape from my side.

---

dsarno (commented):
Live end-to-end smoke of Vision Routing at `646035c`, run in a Linux container (editor under `xvfb-run` + `--rendering-driver opengl3`, real MCP session, real Gemini API). One new bug found, and everything else held up.

## New finding: the curated Gemini model id is also retired

`gemini-2.5-flash` is no longer served to new API keys — `generateContent` returns:

```
HTTP 404: "This model models/gemini-2.5-flash is no longer available to new users.
Please update your code to use a newer model for the latest features and improvements."
```

Verified with **two** freshly-created AI Studio keys (`gemini-2.0-flash` 404s the same way), so the Gemini provider is dead-on-arrival for any new user — the same class of issue as the original `grok-2-vision-latest`. Note the model is still listed by `GET /v1beta/models`, so a listing check wouldn't catch it; only a real `generateContent` call does. Two ids verified working with a new key:

- `gemini-flash-latest` — Google's maintained rolling alias; best fit for the "screenshot tool never breaks" goal (`routed_via` then truthfully reports the alias). 
- `gemini-3-flash-preview` — closer to the fixed-id philosophy, but previews get retired too.

Fix is the `PROVIDERS` entry (`model` + `path`), the file-header comment, and the docs table. Two silver linings, both live-confirmed: the **Test connection** button pings the exact model endpoint, so it correctly surfaces this as `FAILED - Google Gemini HTTP 404: ...` for affected users; and the 404 fell back cleanly (original image returned, no tool error).

## What the live smoke confirmed working

With the model id patched locally to `gemini-flash-latest`, the full pipeline works:

- Routed a running game's frame (the CI quadrant scene): the returned `vision_description` matched the pixel-validated ground truth exactly — "Top-Left quadrant: Solid pure red … Top-Right: green … Bottom-Left: blue … Bottom-Right: white".
- `routed_via: google:gemini-flash-latest`; `note` self-attributes ("Vision description (google:gemini-flash-latest): …"); `include_image=true` returns a single text block with the image dropped server-side; `include_image=false` carries the description in metadata.
- Fallback held in both failure modes exercised live: enabled-with-no-key (clean pass-through + the log hint naming the env var) and provider API error (404 → original image, no tool error).
- Also reproduced live: `user_prompt` doesn't reach `source="game"` captures (the quadrant question I passed never entered the prompt; the base prompt happened to cover it) — matches the round-2 follow-up already on file.

With the Gemini id updated plus the three small round-2 follow-ups, this remains in merge shape from my side.

---

dsarno (commented):
Design direction for the two open questions, discussed with and endorsed by the maintainer. This supersedes the "switch Gemini to `gemini-flash-latest`" suggestion in my previous comment.

## 1. Make the model id required user input — don't ship curated model defaults

The retired-id problem shouldn't be patched one id at a time; it should be removed as a class. Measured on this PR, shipped pins were 2-for-3 dead on arrival (`grok-2-vision-latest`, `gemini-2.5-flash`) past an author, two review bots, and a reviewer — staleness is the steady state of baked-in third-party model ids, and the fallback makes a dead default *silently* dead. Meanwhile the person configuring this feature has already created a provider account and pasted an API key out of a developer console; the model name is on the same page, and when a model retires, the provider notifies the account holder — not this repo.

So: **curate the protocol, require the model.**

- `PROVIDERS` keeps encoding what's stable and repo-owned: label, dialect (`openai` vs `gemini` `generateContent`), host/path shape, env var, encrypted key slot.
- The model id becomes a required per-provider field, stored alongside the per-provider key (empty by default). Switching providers switches to that provider's entered model, so the original "never silently calls the wrong model" property survives — the fear was the *software* guessing; a typed id is the opposite of a guess.
- `routed_via` keeps reporting `provider:<entered model>` verbatim.

Three accompaniments so the required field is a feature and not a trap:

1. **Test connection must validate vision, not just auth.** The current ping is text-only ("Reply with exactly: OK"), so a text-only model id would pass the test and then fail — or hallucinate — at screenshot time. Attach the existing `_PLACEHOLDER_PNG` to the ping body; one click then validates key + model existence + image capability. (Live-verified this week that the ping correctly surfaces provider 404s with their own error text, so a wrong or retired id fails loudly right where it's entered.)
2. **Empty model behaves exactly like empty key** — routing declines, image passes through, log line + tab status say "set a model id". That path already exists and is live-proven for keys; this is symmetry, not new machinery.
3. **Suggestions, not defaults**: placeholder text and a date-stamped docs table of known-good ids ("as of 2026-08 — check your provider console"). A stale suggestion rots visibly (the user sees the 404 and goes looking); a stale default rots invisibly behind the fallback. As of today, verified working: Groq `qwen/qwen3.6-27b`, Gemini `gemini-flash-latest` or `gemini-3-flash-preview` (note `ListModels` still advertises retired ids — only a real `generateContent` call proves serviceability).

This also deletes the follow-up burden the alternative designs carried (fallback chains, rolling-alias debates, a scheduled CI canary with maintainer-held keys): zero repo-side maintenance for third-party model lifecycles. Also still worth keeping from the earlier rounds: append a short templated reason to `note` on routing failure, so text-only agents — who otherwise just receive a useless image block — learn the feature is down and why.

## 2. UI placement: no dock row; Vision Routing is a Settings section

- **Drop the dock quick-toggle** (`build_toggle_row`). The dock is status surface; Vision Routing is configuration, and it's currently the only feature config living there (it's not even dev-gated, so it's in every user's face). With the model field required, the bare toggle can't make the feature work on its own anyway.
- **Fold the Vision Routing UI into the existing Settings tab as a section** rather than a fourth top-level tab — `build_tab(tabs)` becomes `build_section(parent)`; the controls themselves are unchanged.
- That requires the Settings tab to be usable **without developer mode**, which surfaces a pre-existing wart: the tab's only current content is the dev-gated "Allow remote hosts (CIDR)" block (`_build_settings_tab`, #507), whose description plus security warning is a wall of text. Suggested restructure, per the maintainer: put the allow-hosts block behind its own collapsed disclosure ("Remote access (advanced)" — `FoldableContainer` is available on the supported Godot versions, or a plain toggle + hidden VBox), so the warning renders exactly at the point of configuration instead of on every Settings visit. Keep the warning text itself verbatim — it's load-bearing for the WS trust boundary — and **auto-expand the section whenever a non-empty CIDR is already configured**, so an active off-loopback bind is never hidden behind a collapsed header. With that, the disclosure replaces the dev-mode gate for this block; dev mode keeps gating the log viewer and dev sections as today.

---

dsarno:
Live-run finding on the round-3 suite: `test_empty_model_declines_route_like_empty_key` (`test_project/tests/test_vision_routing.gd:309`) is environment-dependent.

The test forces the **groq** model slot empty and sets a **groq** key, but `_start_route` resolves the provider via `_active_provider_id()` from Editor Settings. On a machine where `vision_routing/provider` is any other configured provider — exactly the dogfooding setup this feature encourages — the route proceeds instead of declining, the assertion fails, and the failing path fires a real network request carrying the test's garbage `"AAAA"` payload at the configured provider.

Repro (Godot 4.7-stable, clean checkout of `4230e3e`, live editor):

- `vision_routing/provider = "google"`, `vision_routing/google_model = "gemini-flash-latest"`, `GOOGLE_API_KEY` set → `vision_routing` suite 22/23: `test_empty_model_declines_route_like_empty_key: Expected true`.
- Same editor, provider flipped to `"groq"` → 23/23 green.

Suggested fix: save/restore and pin `vision_routing/provider = "groq"` for the test's duration, the same way it already pins the key and model slots.

For what it's worth, the rest of round 3 verified nicely against a live editor here: 2127/2142 project-wide with only this failure, and the new failure-note pass-through worked as documented when the provider returned a real HTTP 400 (image came through with the `Vision routing unavailable (google): …` note attached, for both editor and deferred game captures).

---

dsarno:
Follow-up to my earlier note: I've now also live-verified the **happy path** end-to-end with a valid Gemini key (Godot 4.7-stable under Xvfb, `4230e3e`, provider `google`, model `gemini-flash-latest`, key via `GOOGLE_API_KEY`). Test image was a programmatically drawn scene — a red spider on a web under a full moon. Everything worked as designed:

- **Editor capture (`viewport_2d`)**: image block dropped, text-only reply with `routed_via: "google:gemini-flash-latest"` and an accurate `vision_description` — it correctly identified the starfield, the web "fanning outward from a central anchor point," and "a stylized bright red spider… round white eyes with dark pupils, eight bent red legs, hanging from a single vertical white line/thread."

- **Game capture with the deferred `user_prompt` stash** (the round-3 addition): sent `user_prompt: "Why is the spider red? And what mood does this scene evoke?"` with `source="game"`. The description directly answered both questions ("the sprite artwork itself is styled entirely in bright red with a dark red 'X' mark on its body… evoking a nocturnal, spooky, or Halloween-themed mood") — live proof the stashed prompt survives the deferred hop through `route_game_payload` and reaches the provider.

- **Test connection**: the image-bearing ping showed both states cleanly — `FAILED - Google Gemini HTTP 400: … API_KEY_INVALID` with the key source (`via GOOGLE_API_KEY`) when the key was bad, and `OK - Google Gemini responded (via GOOGLE_API_KEY).` once valid. Validating key + model + vision capability in one click works as intended.

- One incidental observation, arguably a feature: when a capture itself is degenerate (the 2D screen wasn't the active main screen yet, so the tool grabbed a 2×2 stub), the routed description faithfully reported "entirely solid black pixels" rather than masking the bad capture — routing doesn't hide upstream capture problems.

Combined with the failure-path check from before (HTTP 400 → image pass-through with the `Vision routing unavailable (…)` note, on both editor and game paths), the full feature matrix behaves as documented. The only outstanding item from my testing remains the provider-dependent test in the comment above.

---
ui_test_connection_ok
ui_editor_full

dsarno:
Confirmed: `d321750` fixes the environment-dependent test. Re-ran the `vision_routing` suite on the same machine that previously failed it (configured `provider = "google"`, google model set, `GOOGLE_API_KEY` in env) — now **23/23** with the provider pinned for the test's duration, and the no-machine-id skip guard is a nice touch.

The reworked `test_game_capture_stashes_and_forwards_user_prompt` now drives the real `route_editor_screenshot` → `route_game_payload` path with the `_route_worker` spy and passes; since the enabled-gate lives in the callers rather than the service, the test is deterministic regardless of ambient settings. The Gemini `generationConfig.maxOutputTokens` addition matches the documented v1beta schema and is covered by the new body-builder assertion (a live re-validation of that field was blocked on my side by an expired key — auth is rejected before body validation, so no schema signal there; the unit coverage plus yesterday's live-verified surrounding body shape make the risk negligible).

One more incidental fallback datapoint while re-testing: a routed capture with the expired key surfaced as `Vision routing unavailable (google): Google Gemini HTTP 401: … UNAUTHENTICATED` with the image passing through — the failure contract holds for the 401 variant too.

Nothing further from me on this round.

---

michaltomczykowski:
Hey, thanks a lot! If there is nothing more to adress i think it finished then

dsarno:
Confirmed — nothing further from my review. Everything raised across the four rounds has been addressed and verified against a live editor (suite green in the environment that caught the round-3 test issue, happy path and failure paths both exercised end-to-end against real providers). Final call on merging rests with the maintainers as usual. Thanks for the quick, thorough turnarounds on each round — this landed in really solid shape.

---

dsarno:
Merged — thank you @chivago19300-source for this contribution! Vision Routing is a genuinely useful capability (text-only models can finally "see" the editor and game), and the way you handled four rounds of review — fast turnarounds, every finding addressed properly rather than patched around, live-verifying as you went — made this a pleasure to review. Welcome aboard, and we'd be glad to see more from you. 🎉

---

PR #855, opened 2026-08-06
feat: GridMap and CSG authoring tools for 3D map building

michaltomczykowski:
Adds two new MCP tool domains for 3D geometry authoring in the editor.

## GridMap (`gridmap_manage`)
- `gridmap_set_item` — set a cell item at (x, y, z), `item=-1` erases; orientation validated 0..24
- `gridmap_fill` — fill a 3D box region in one undo action (4096-cell cap, batched into a single undo-history entry)
- `gridmap_clear` — clear all cells, undoable via snapshot (refuses oversized maps above 4096 cells)
- `gridmap_get_used_cells` — read placed cells
- `gridmap_list_library_items` — discover MeshLibrary item ids/names/meshes before placing (the 3D analogue of `tileset_get_atlas_image`)

## CSG (`csg_manage`)
- `csg_create` — box / sphere / cylinder / torus / polygon under a Node3D parent, with `union` / `intersection` / `subtraction` (holes, caves, tunnels); an omitted name falls back to the shape class
- `csg_set_operation` — change the boolean operation with undo

Both follow the existing `tilemap_handler.gd` contract: `McpNodeValidator` resolution, `scene_file` guards, `EditorUndoRedoManager` undo, lazy handler registration in `plugin.gd`, `register_manage_tool` rollups, readiness gating, `tool_catalog.gd` + `domains.py` sync, docs in `TOOLS.md`.

## Tests
- Python unit: 20 new tests in `tests/unit/test_gridmap_csg_handlers.py` (forwarding + `require_writable` gating; read ops verify no writable gate)
- Integration: 5 rollup dispatch tests in `tests/integration/test_mcp_tools.py` (incl. envelope-readiness preservation for the new commands)
- Godot: `test_gridmap.gd` (12 tests) + `test_csg.gd` (12 tests) — undo round-trips, region-fill undo preserving outside cells, erase via negative item, library listing, exact `MAX_FILL_CELLS` boundary, oversized-clear guard, default-name / torus / polygon / intersection coverage, wrong-type / scene-mismatch / out-of-range guards, stored-cell read-backs
- `test_server_instructions.py` frozen surface updated

## Verification
- `pytest tests/unit tests/integration` — 1686 passed / 0 failed
- Godot suite (editor) — 2144 passed / 0 failed (gridmap 12/12, csg 12/12)
- `ruff check` — clean
- Command-name parity (Python `send_command` vs `plugin.gd` `register_lazy`) — verified
- CI — 23/23 checks green

Issue #781, opened 2026-07-21
Design: custom tool extension point — project-registered MCP tools without core changes


dsarno:
## Motivation

#776 proposed running a project's gdUnit4 suites through the MCP. We're declining the in-core coupling (see the closing comment there), but the underlying need is real and recurring: projects want to expose project-specific operations — third-party test runners, build steps, custom editors — to agents through the already-connected MCP channel, without forking core handlers.

## Sketch (for discussion, not committed)

- **Plugin side**: a registration API on the dispatcher — an addon or project script (e.g. `res://mcp_tools/*.gd`, or an explicit `McpToolRegistry.register(...)` call from any `@tool` script) declares `{name, description, params_schema, handler}`, with opt-in support for the deferred-response pattern (`DEFERRED_RESPONSE` + `send_deferred_response`) and a per-command timeout entry.
- **Server side**: custom commands are discovered via the existing capabilities handshake and exposed either under a single namespaced rollup (`custom_manage(tool=..., params=...)`) or as dynamically registered MCP tools. The rollup keeps the advertised surface stable and is the conservative default.
- **Capability surfaces** (instructions / catalog / status) must reflect registered custom tools — the same honesty bar #772 sets for exclusions.

## Constraints

- No behavior change for built-in tools.
- Custom tools run with the same editor privileges any addon already has — not a new trust boundary, but it must be documented plainly.
- Server-side schema validation for custom tool params.
- Domain-exclusion and telemetry interplay defined up front: custom-tool errors must not pollute fleet metrics (tag them as custom).
- A docs page with a worked example.

## Validation case

A gdUnit4 runner addon built outside core against this interface (see #776 for the mechanism research: trigger via `GdUnitCommandHandler`, subscribe to `GdUnitSignals`, deferred response on `SESSION_CLOSE`) — by its proposer, if willing, and we'd gladly link it from the README once it exists.

## Open questions

1. Rollup (`custom_manage`) vs dynamic per-tool MCP registration?
2. Discovery timing: register at plugin scan vs lazily?
3. How do custom-tool failures surface in the error-code taxonomy?
4. Should the dock list registered custom tools?

---
_Generated by [Claude Code](https://claude.ai/code)_

21ess:
@dsarno hi. This looks quite funny. I usually use the gdUnit4 plugin and hope to try to solve this problem. Is there anyone handling this issue at present? If not, I am willing to contribute.

dsarno:
[...]

Issue #776, opened 2026-07-20
Feature proposal: run a project's gdUnit4 suites via MCP (in-editor, with warnings surfaced)


eth0net:
## Motivation

`test_run` today drives the addon's own `McpTestSuite`/`McpTestRunner`. Projects that use **gdUnit4** (a very common Godot test framework) have no MCP path to run *their* tests — an agent has to shell out to gdUnit4's headless CLI in a separate process, which is slow and, run against the live project, races the editor's `.godot` import cache. Being able to run the project's gdUnit4 suites through the connected editor — the same way `project_run` drives game scenes — would let an agent iterate on game logic and see failures **and warnings** without leaving the MCP.

Flagging up front: I couldn't find any existing issue/discussion on third-party test-runner support, so this may be out of scope — very happy to be told so before anyone invests.

## How gdUnit4 6.1.x actually works (so "in-editor" is well-defined)

I dug through gdUnit4 `v6.1.3` source to confirm the mechanism, because it changes the design:

- gdUnit4 **never runs tests in-process** in the editor. Its own GUI "Run" **spawns a second Godot process** (`GdUnitTestRunner.tscn`) that executes suites and streams results back over **TCP**.
- The editor process hosts a `GdUnitServer` and **re-emits** every received event on a global bus: `GdUnitSignals.instance().gdunit_event` (`GdUnitEvent`) / `gdunit_message`.
- Runs are triggerable programmatically without the GUI via `GdUnitCommandHandler.instance()` (e.g. `test_session_command.execute(tests: Array[GdUnitTestCase], is_debug=false)`), with test cases discovered through `GdUnitTestDiscoverer` / `GdUnitTestSuiteScanner`. `GdUnitEvent`/`GdUnitReport` carry per-test status + message + line; `SESSION_CLOSE` marks the end.

So the addon would **not** run tests itself — it would drive gdUnit4's machinery from inside the editor and subscribe to gdUnit4's own signal bus (gdUnit4 still owns the isolated subprocess, exactly as when a user clicks Run). That means:
- no new headless/cache-concurrency hazard (it's gdUnit4's normal editor flow), and
- it maps cleanly onto the existing **deferred-command** pattern: trigger → return the deferred sentinel → accumulate events → `send_deferred_response` on `SESSION_CLOSE`.

Requires the gdUnit4 plugin to be enabled (guarded via `Engine.has_meta("gdunit_server_port")`); if absent, return an actionable error.

[...]

PR #856, opened 2026-08-06
feat: terrain domain - deterministic heightmap meshes with collision

michaltomczykowski:
Adds a new MCP tool domain for heightmap terrain authoring in the editor.

## Terrain (`terrain_manage`)
- `terrain_create` — build a deterministic FastNoiseLite heightmap under a Node3D parent: an indexed ArrayMesh ("TerrainMesh") with height-tinted vertex colors and smooth per-vertex normals, plus an optional "TerrainCollision" StaticBody3D with a `ConcavePolygonShape3D` from the same triangles. Params: `size` (4..128), `cell_size`, `seed`, `noise_type` (simplex / simplex_smooth / perlin / ridged / value), `frequency`, `octaves`, `height_scale`, `base_height`, `generate_collision`
- `terrain_regenerate` — rebuild an existing terrain container in place with new params; one undo action restores the previous mesh/collision state

**Deterministic by design**: the same seed + params always produce the same mesh, so agents can iterate on seeds and compare screenshots honestly. Holes and carved detail are CSG territory (`csg_manage`) — composition, not duplication.

Follows the same conventions as the gridmap/csg PR: `McpNodeValidator` resolution, `scene_file` guards, `EditorUndoRedoManager` undo (node-first history routing), lazy handler registration in `plugin.gd`, `register_manage_tool` rollup, readiness gating, `tool_catalog.gd` + `domains.py` sync, docs in `TOOLS.md`.

## Tests
- Godot: `test_terrain.gd` (10 tests) — stored-Variant readbacks from the mesh (exact vertex count, positions), full-array seed determinism (interior cell, since grid (0,0) maps to seed-independent noise(0,0)), regenerate round-trip + undo restore, collision on/off, degenerate-param guards, wrong-type / scene-mismatch errors
- Python unit: `tests/unit/test_terrain_handlers.py` — forwarding + `require_writable` gating for both ops
- Integration: 3 rollup dispatch tests (create, regenerate, envelope-readiness preservation)
- `test_server_instructions.py` frozen surface + `test_dispatcher.gd` lazy-handler count updated

## Verification
- `pytest tests/unit tests/integration` — 1673 passed / 0 failed
- Godot suite (editor) — 2131 passed / 0 failed (terrain 10/10)
- `ruff check` — clean
- Command-name parity (Python `send_command` vs `plugin.gd` `register_lazy`) — verified

## Summary by CodeRabbit

- **New Features**
  - Added terrain authoring tools for creating and regenerating deterministic heightmap terrain.
  - Supports configurable size, noise, seed, elevation, and optional collision generation.
  - Terrain creation and regeneration can be undone in the editor.
  - Added terrain tools to the available domain catalog and management interface.

- **Bug Fixes**
  - Added validation for invalid terrain parameters, parent nodes, scene paths, and regeneration targets.

- **Documentation**
  - Documented terrain creation and regeneration operations.

- **Tests**
  - Added coverage for terrain generation, determinism, collision, validation, regeneration, and undo behavior.


michaltomczykowski:
Addressed all five review comments in 50fa1bd:

**Fixed (with regression tests):**
- **Frequency was applied twice** — get_noise_2d(x * freq, y * freq) on top of FastNoiseLite.frequency squared the effective frequency. Now raw grid coordinates; new 	est_terrain_frequency_changes_shape distinguishes 0.02 vs 0.2.
- **Owners assigned before attachment were rejected** — _build_children set scene_root as owner while the container was not yet in the tree, so the generated mesh/collision would not persist on scene save. Owners are now assigned via a do-method in the undo chain (_assign_owners, walks descendants) after dd_child; tests assert owner == scene_root on every generated node.
- **Routing metadata persisted into the scene** — _apply_rebuild now drops the _mcp_terrain_rebuild marker after it has served its history-routing purpose.
- **Collision restoration on regenerate-undo was uncovered** — the undo test now regenerates with generate_collision: false, asserts the body is removed, then asserts undo restores it.

**Addressed with measurement (frame-budget comment):**
A max-size (128×128) build measures **~27ms** (default 48×48 ~4ms) — a one-shot hitch on an explicit authoring command, the same order as the existing 4096-cell tilemap/gridmap fills accepted in this codebase. The 4ms guideline governs the plugin's continuous per-frame processing; generation is deliberately synchronous because it touches no scene state until the final attach, and the repo reserves worker threads for blocking I/O (vision routing). Documented in the handler header.

Verification: terrain 11/11, full Godot suite 2132/0, pytest 1673/0, ruff clean, 23/23 CI.

dsarno:
Thank you for this one too — the craftsmanship is the same high standard as #855 (which just merged), and the deterministic-seed design is genuinely the right call for agent iteration. A security pass over the diff also came back clean.

We're holding this PR rather than merging it into core, and the reasoning is policy, not quality — the same core-vs-custom line drawn in #776/#781:

- **"Terrain" is not a Godot concept.** GridMap and CSG (#855) are thin authoring wrappers over built-in engine node types. This PR instead *invents* a node shape — an opinionated generator with baked-in aesthetics (height-tinted vertex colors, a fixed material, `TerrainMesh`/`TerrainCollision` conventions). That's a content generator, not an editor-capability wrapper, and it sits uneasily with our rule that the surface should look like a Godot product.
- **A terrain domain in core invites scope creep.** Texture painting, erosion, chunking/LOD, holes — each will be a small step from what shipped, and every accepted one grows a compatibility surface core owns forever, in territory that dedicated terrain addons already serve well.

To be fair to the other side of the ledger: it depends only on engine built-ins (FastNoiseLite, SurfaceTool), so the third-party-coupling disqualifier from #776 doesn't apply. This was a close call.

The plan: #781 is designing a custom tool extension point so addons can register their own MCP tools without core changes. This terrain domain would be an excellent second validation case for that design alongside the gdUnit4 runner — a first-party-quality consumer that exercises the registration surface with real geometry authoring. We'd love to see it live as an addon built against that interface once it lands, and we'd link it from the README.

Leaving this open and marked as held pending #781. Your input on the registration surface over there would be valuable — you've now built exactly the kind of tool it needs to serve.

---

dsarno (commented):
## Security review

No malicious or suspicious code, and the guard parity with sibling handlers checks out: both Python ops call `require_writable_async` before `send_command` (matching `tilemap.py`, pinned by the unit tests); the GDScript side routes create through `McpScenePath.require_edited_scene` + parent resolve + Node3D type check and regenerate through `McpNodeValidator.resolve_or_error` with `scene_file` support. The mesh-generation math is exactly what it claims — FastNoiseLite sampling, normals, indexed SurfaceTool emission — with all loops bounded by the `size` clamp (4..128 → ≤16,384 verts) and `octaves` 1..6, so no memory-exhaustion path. Registration changes are purely additive.

Three low-severity findings, none blocking, but the first is worth addressing:

1. **Destructive regenerate on non-terrain nodes with lossy undo** (`terrain_handler.gd:110-134`, `_clear_generated` at ~375): regenerate accepts *any* Node3D and immediately frees children literally named `TerrainMesh`/`TerrainCollision`, while `_capture_mesh_state` snapshots only the mesh resource and first collision shape — not scripts, material overrides, or extra children. A misdirected `terrain_regenerate` on a hand-built node with a name-colliding child destroys data that undo only partially restores. Suggest requiring an MCP-terrain marker (e.g. metadata set at create time) before clearing, or deep-capturing prior children.
2. **No `scene_file` pin on the Python tool** — the GDScript handler honors the opt-in wrong-scene guard, but `src/godot_ai/handlers/terrain.py` never exposes it, so a scene switch racing the call writes terrain into the wrong open scene. Matches the existing `tilemap.py` gap, so noting rather than blocking.
3. **Non-finite floats pass validation** (`_validate_params`, ~152-163): NaN fails safely past the `<= 0.0` comparisons and inf `cell_size` is accepted — result is a garbage mesh, bounded work, not exploitable. Robustness nit.

---

dsarno (changes requested):
## Full review (correctness, tests, conventions)

Follow-up to the security review. Ran `pytest` on the PR head (1687 passed, 5 skipped) and `ruff` clean. The domain integration is exemplary — lazy registration, rollup pattern, catalog/domains/instructions sync, undo bundling, readiness gating, and test hygiene all follow the repo conventions, and the earlier CodeRabbit items are genuinely resolved at head (owner assignment via `_assign_owners` after `add_child`, the `_mcp_terrain_rebuild` meta cleaned up in the do-chain, frequency applied once, regenerate-undo collision restore tested). But there's one blocker:

1. **Blocker — triangle winding is inverted: the terrain renders back-face-up and is invisible from above.** In `_build_surface` (~236-243), `_emit_tri(st, triangles, verts, a, c, b)` / `(b, c, d)` produces a +Y cross product, i.e. counter-clockwise seen from above — but Godot's front face is **clockwise** (verified against `PlaneMesh::_create_mesh_array` in the engine: its up-facing plane emits triangles whose cross is −Y). So the mesh's front faces −Y; with a default cull-back material the terrain is culled from every viewpoint above the surface, while the hand-computed normals point up — normals and winding actively disagree. The collision faces are wound the same way (`backface_collision` is false, so fast bodies from above can tunnel). `terrain_create` → look down in the viewport → nothing renders, which defeats the tool's own "iterate on seeds and compare screenshots" purpose. Fix: emit `(a, b, c)` and `(b, d, c)` — the existing normal-accumulation order is then already correct. Please add a test asserting `(v1−v0).cross(v2−v0).y < 0` for the first triangle, and do the live viewport smoke `docs/verification.md` requires — no current test asserts orientation, which is exactly why this survived.

Minor, can ride along or follow up:

2. `regenerate` builds the whole surface twice — `_build_surface(p)` for the response counts, then again inside `_apply_rebuild` via `commit_action()`. Pass the built data in, or derive counts arithmetically (`size²`, `2·(size−1)²`).
3. `terrain_regenerate` on the Python side fills every default, so omitted params silently reset the terrain (create at size=96 → regenerate with just a new seed snaps back to size=48). At minimum document "restate all params" in `_DESCRIPTION`; better, persist creation params and merge.
4. Grid (0,0) samples `noise.get_noise_2d(0,0)` which is 0 for gradient noise regardless of seed — the corner (and, at some frequencies, a lattice of points) is pinned to `base_height` for every seed; the tests route around it rather than the handler fixing it. Sampling at a half-cell offset removes the pinned points cheaply.
5. `height_scale=0` passes validation but makes `_height_color` compute `inverse_lerp(-0.0, 0.0, …)` → NaN → all-snow tinting. Require `> 0` or guard the degenerate case.
6. Nits: `test_terrain_frequency_changes_shape` would pass even with squared frequency (distinct meshes either way) — a real pin is comparing against an independently-constructed `FastNoiseLite`; `docs/tool-surface.md:16` and `TOOLS.md:3` still say "~43 tools / 24 rollups" (now 44/25).

With the winding fixed plus an orientation test and a viewport smoke, this is ready — everything else about the PR is genuinely well-built.

---

dsarno:
@michaltomczykowski — the extension point this PR was held for has now landed, so here's the resolution we promised in the hold comment.

**What shipped:** #875 (merged as `0a7edb3`, closing #781) added `McpToolRegistry` — third-party Godot addons register their own MCP tools with data-only specs, no core changes. Specs with `promoted: true` surface as first-class MCP tools (`custom_<name>`) advertising your own `params_schema` verbatim, with per-tool enable/disable in the dock and per-project persistence. The registration surface is documented in `docs/plugin-architecture.md` (custom tools section) and `docs/TOOLS.md`.

**What this means for terrain:** we're closing this PR per the core-vs-custom policy from the earlier hold comment — not for quality, which remains excellent. The invitation stands and is now concrete: ship this as a standalone addon that registers `terrain_create` / `terrain_regenerate` (promoted) via `McpToolRegistry`. The conversion is smaller than it may look — roughly half this diff was core integration (`domains.py`, `tool_catalog.gd`, rollup wiring, server-instructions sync) that simply disappears in the addon form. `terrain_handler.gd` and your GDScript tests are the part that carries over.

**The review findings carry over as your addon's punch list** — the inverted triangle winding first (the terrain renders back-face-up and is invisible from above regardless of where the code lives; fix is emitting `(a, b, c)` / `(b, d, c)`), then the regenerate double-build, the regenerate-defaults reset, the (0,0) seed-pinned sample, and the `height_scale=0` NaN tint. The full details are in the changes-requested review above.

When the addon exists, we'd be glad to link it from the README as promised — it would be the first community custom-tool addon, and this codebase is exactly the validation case the registry was designed for. Happy to answer questions about the registration surface here or in a new issue.

Thanks again for the quality of this work — twice over, since it also pressure-tested the custom-tool design.

---

Issue #712, opened 2026-07-13
Audit Tier-2: connection/dispatcher resilience (pause-depth wedge, queued-commands-across-connections, worker-thread joins, sync main-thread probes)


dsarno:
Tier-2 items from `docs/audit-tier2-plan.md` (transport/lifecycle-resilience group). Verify each against current code before implementing.

- [ ] **Handler crash inside a `pause_processing` window permanently wedges the transport** (`connection.gd:95`, S3·concurrency) — restore pause balance at `McpDispatcher._call_handler`: snapshot `pause_depth` before invoking, restore after, per the backlog's proposed shape.
- [ ] **`_clear_on_disconnect` leaves already-queued commands to execute under the next connection** (`connection.gd:186`, S2·concurrency) — decide the cross-connection contract: clear `_command_queue` on disconnect, or document the carry-over as intended.
- [ ] **`_run_blocking` worker Threads are never joined at teardown** (`server_lifecycle.gd:118`, S2·concurrency) — track in-flight workers on the manager and add a blocking `drain_workers()` for teardown.
- [ ] **Three synchronous main-thread probe/discovery paths contradict #678's non-blocking-startup claim** (`server_lifecycle.gd:369`, S2·drift) — move `_set_incompatible_server`'s proof evaluation (and siblings) onto `_run_blocking` with main-thread-read inputs, mirroring `recover_strong_port_occupant`.
- [ ] **`project_handler`'s deferred finishers are instance-method coroutines** (`project_handler.gd:197`, Tier-1 S2·drift, deferred from PR #710) — static conversion cascades through `_run_project_liveness_decision`/`_run_project_response`; parameterize `connection` + `debugger_plugin` explicitly with `is_instance_valid` checks after each await, mirroring the script/filesystem handlers' pattern.
- [ ] **Disabling the plugin mid-walk resumes `_start_server` on a freed instance** (`plugin.gd:711`, S1·defense/report-only) — fire-and-forget the walk; move trace-finish stamping into the manager's existing validity-checked path.

Full evidence: `docs/audit-tier2-plan.md`. Lifecycle changes here require both `ci-stale-server-smoke` modes per the plan.

PR #862, opened 2026-08-07
fix: stop handshake verdicts from stealing the startup walk's worker slot

dsarno:
## Symptom

On v3.1.2/v3.1.3, whenever editor startup finds an older godot-ai server holding the HTTP port (e.g. a v3.1.1 attach-owned server on :8000), the editor Output shows:

```
SCRIPT ERROR: Trying to assign value of type 'Nil' to a variable of type 'Dictionary'.
  at: McpServerLifecycleManager._start_server_impl (res://addons/godot_ai/utils/server_lifecycle.gd:815)
  GDScript backtrace: [0] _start_server_impl (server_lifecycle.gd:815), [1] _run_blocking (server_lifecycle.gd:173)
```

A retry walk then completes the incompatible-server banner + `MCP | proof: status_name`, masking the failure from users.

## Root cause

The deferred startup walk (#678) suspends in `_run_blocking` while the plugin's connection dials the occupant's WS port in parallel. The handshake against the old server delivers a version-mismatch verdict from `_process` (`McpServerVersionCheck` → `handle_server_version_verified`), which fires `_set_incompatible_server` fire-and-forget. Its #712 diagnosis tail (`_evaluate_recovery_port_occupant_proof` via `_run_blocking`) then claimed the single active-worker slot out from under the suspended walk — violating the single-owner invariant the comment at the crash site documents. The walk's op lost the `_invalidate_async_startup` join guarantee, and its resume received `null` **without** a generation bump, so the only guard (`_async_stale`) passed and the typed `Dictionary` assignment crashed. The missing-handshake_ack verdict (`handle_server_version_unverified`) had the same race.

## Fix (two layers)

1. **Restore the single-owner invariant**: `_set_incompatible_server` takes a `caller_owns_worker_slot` parameter. Sync callers (both verdict arms, the force-restart failure arm) leave it false, so the head runs `_invalidate_async_startup()` first — cancelling and joining the suspended walk exactly the way the recovery click already does (#712) — before the tail claims the slot. The startup walk, which owns the slot and awaits the tail, passes `true`.
2. **Harden every `_run_blocking` call site** (12 sites): bail on `_async_stale(...) or result == null` before any typed use. The two `bool(await _run_blocking(...))` wrappers were restructured to untyped-first as well — `bool(null)` / `int(null)` are themselves SCRIPT ERRORs in GDScript ("Nonexistent 'bool' constructor"), and the port-probe site crashed *before* its staleness check. A future slot loss without a generation bump now unwinds instead of crashing.

## Regression test

`test_sync_verdict_mid_walk_cancels_walk_before_claiming_worker_slot` suspends a real deferred walk (worker-side nap guarantees suspension) and lands the verdict synchronously. It makes only synchronous observations because the test runner never awaits test coroutines. Stash-verified red on the unfixed code: it aborts with "Attempted to free a locked object" — the un-joined worker still executing a host method, the same use-after-free class the join guarantee exists to prevent (the macOS reload-churn wedge).

## Verification

- `ruff check` clean; `pytest`: 1682 passed, 5 skipped
- Full GDScript suite against a live editor: 2116/2143 passed, 0 failed (server_lifecycle 76/76; new test red pre-fix, green post-fix)
- Live mismatch fixture (v3.1.1 server on an isolated port vs v3.1.3 plugin, Godot 4.6.3): pre-fix reproduces the exact SCRIPT ERROR at line 815 on the first launch; post-fix, two runs latch the banner + proof with zero errors
- `script/local-self-update-smoke`: all checks PASS (v3.1.3 → synthetic v3.1.4, both sides carrying the fixed lifecycle; no crash, no `.ips`, temp dir consumed, no vNext `_exit_tree` during the update window)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server startup and recovery reliability when an incompatible server version is detected.
  * Prevented outdated or cancelled background operations from incorrectly changing server state.
  * Improved cancellation handling across server checks, shutdown, recovery, and restart flows.
  * Ensured competing startup work is cancelled before a new startup attempt proceeds.

* **Tests**
  * Added regression coverage for interrupted startup operations and incompatible server detection.

### DIFF
--- a/plugin/addons/godot_ai/utils/server_lifecycle.gd
+++ b/plugin/addons/godot_ai/utils/server_lifecycle.gd
@@ -151,9 +151,14 @@ var _active_blocking_thread: Thread = null
 ##
 ## Returns null (without joining) when `_invalidate_async_startup` took
 ## ownership of the thread mid-flight — the walk is stale at that point
-## and must bail at its next staleness check, so callers assign the
-## result to an untyped local BEFORE the staleness check (a typed
-## assignment would trip on the null first).
+## and must bail. Callers therefore assign the result to an untyped
+## local and bail on `_async_stale(...) or result == null` BEFORE any
+## typed use — a typed assignment (or a bool()/int() constructor, both
+## of which have no Nil form) trips on the null first. The null check is
+## not redundant with the generation check: a caller that loses the slot
+## without a generation bump — an invariant violation, but exactly what
+## a concurrent fire-and-forget `_run_blocking` user produces — must
+## still unwind instead of crashing on the Nil.
 func _run_blocking(work: Callable) -> Variant:
 	if not defer_blocking_work:
 		return work.call()
@@ -407,7 +412,12 @@ static func _managed_record_has_version_drift(record_version: String, current_ve
 
 # ---- Incompatible-server bookkeeping ----------------------------------
 
-func _set_incompatible_server(live: Dictionary, expected_version: String, port: int) -> void:
+func _set_incompatible_server(
+	live: Dictionary,
+	expected_version: String,
+	port: int,
+	caller_owns_worker_slot := false
+) -> void:
 	## Latches the incompatible diagnosis into manager state and asks
 	## the dock to re-sweep client rows so they don't show stale green.
 	## Threads the caller's `live` snapshot through the recovery proof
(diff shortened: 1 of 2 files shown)


dsarno:
Pushed `5958f13` reverting the CodeRabbit autofix commit (`fde96a7`) — it was the cause of the CI failure on the last run.

The autofix restructured the core of this fix: it changed the sync-verdict path to `_invalidate_async_startup(false)` — cancel *without* joining the suspended walk's worker thread — stashing the thread in a new `_deferred_join_thread` for a later `_wait_for_worker_completion()` poll. That discards exactly the join guarantee this PR restores: the un-joined worker can still be executing a host method during teardown, which is the use-after-free class (the macOS reload-churn wedge) the synchronous join exists to prevent. The PR's own regression test caught it — `test_sync_verdict_mid_walk_cancels_walk_before_claiming_worker_slot` aborted with `Attempted to free a locked object (calling or emitting)` on all four Godot jobs (macOS, Linux, Linux/4.5, Windows), the same signature the PR description documents for the unfixed code.

The branch tree is now byte-identical to `811ea9b`, the commit that passed CI in full. The blocking-main-thread concern behind the bot's suggestion is already addressed in the original design: the join is bounded by the walk op's own timeout, and it's the price of the invariant. Recommend dismissing the associated review suggestion rather than re-applying it.

---

Issue #691, opened 2026-07-12
[S3] Residual #678 async-startup seams not covered by #683


dsarno:
Surfaced by the codebase audit. #683 fixed the worker-thread **teardown join** (`_active_blocking_thread` + `wait_to_finish()`) and now propagates `connect_blocked` on the two `handle_server_version_*` handshake paths. These two related seams from the same #678 async-startup work remain:

### 1. Spawn-time process-env mutation races the dock's worker threads (`server_lifecycle.gd:~699`)
The startup walk's spawn step mutates the editor's **process-global** environment around `OS.create_process`: `OS.set_environment("GODOT_AI_DISABLE_TELEMETRY")` (~467), `GODOT_AI_OWNER_PID` (~479), `GODOT_AI_PLUGIN_SPAWNED` (~493), `PYTHONPATH` (~682), unset (~702-717); `plugin.gd` repeats the PYTHONPATH window for `start_dev_server`. Since #678 made the walk a multi-frame coroutine, this window runs **concurrently** with the dock's client-status refresh `Thread` (`mcp_dock.gd`), whose `check_status` path reads env off-thread via `_cli_finder`.

**Failure:** a refresh/configure/remove worker calling `OS.get_environment` while the main thread is inside `setenv`/`unsetenv` can crash the editor (glibc `getenv` returning a freed pointer) — rare but process-fatal. The tamer failure is semantic: a CLI child spawned by the worker during the window inherits `GODOT_AI_PLUGIN_SPAWNED`/`GODOT_AI_OWNER_PID`/modified `PYTHONPATH` (e.g. a `uvx`-launched status probe mis-configured).

**Fix (two-part, since `OS.create_process` has no per-child env param):** (1) make workers env-read-free — extend the phase-1 main-thread probe snapshot (`client_configurator.gd:262`) to pre-resolve config paths and env-derived values so no `OS.get_environment` runs off the main thread; (2) guard the `set → create_process → unset` sequence so no worker is dispatched during the window.

### 2. Startup-walk INCOMPATIBLE verdict doesn't reach the live connection (`server_lifecycle.gd` recovery-failure path, ~646)
[...]

PR #870, opened 2026-08-21
feat(clients): add Pi Coding Agent support

Rliop913:
## Summary

   - Add Pi Coding Agent as the 23rd supported MCP client.
   - Configure Pi through `~/.pi/agent/mcp.json` using the path-based lazy client registry.
   - Document the required `pi-codemode-mcp` extension:
     `pi install git:github.com/mitsuhiko/pi-codemode-mcp`
   - Clarify Pi code-mode tool and resource call syntax.
   - Clarify scene persistence and inline resource versus reusable `.tres` behavior.
   - Update stale `AGENTS.md` guidance to match the existing path-based client registration behavior.
   - Add Godot-side and Python regression coverage for the Pi descriptor and documentation.

   ## Testing

   - `ruff check .`
   - `pytest -q` — 1702 passed, 19 skipped
   - `script/ci-check-gdscript` — all GDScript files OK
   - Full Godot test suite — 2151 passed, 21 skipped, 0 failed
   - `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - Added Pi Agent as a supported MCP client.
  - Added automatic Pi Agent configuration with MCP settings, URL support, environment fields, installation detection, and legacy compatibility.
  - Added merged configuration handling with precedence, rollback protection, and safe project override detection.
  - Added guidance for installing and using Pi Coding Agent’s code-mode extension.

- **Documentation**
  - Updated setup guidance for 23 supported clients.
  - Clarified code-mode tools, resource handling, scene saving, resource creation, and typeless configurations.

- **Tests**
  - Added coverage for Pi Agent registration, configuration, merging, rollback, and platform support.


dsarno (changes requested):
## Security review

I did a full security-focused review of this PR, including a line-by-line audit of the pinned extension commit (`mitsuhiko/pi-codemode-mcp@0ad55d6`, all ~3.7k lines of source plus its dependency tree).

**The code in this PR is clean.** No unexpected network, file, or process activity; all writes go through `McpAtomicWrite` to the documented `~/.pi/agent/` paths; project-tier files (`.pi/mcp.json`, `.mcp.json`) are read-only and fail closed with an explicit message instead of mutating an inferred project root; the merged-config handling preserves user entries, reuses existing `mcp-servers`/`servers` aliases instead of shadowing them, and multi-file remove is transactional with rollback. `_instance_is_coherent` got stricter, not looser. The audit of the pinned extension itself also came back clean: two mainstream runtime deps (official MCP SDK, typebox), config I/O confined to `~/.pi/agent/`, and the entry this plugin writes is spawned as-is with no rewriting.

### Requested change

We don't want the project's README to instruct users to install third-party code, even pinned and review-first — that's a trust decision each user should make on their own. Please:

1. Replace the README install block (the `pi install git:github.com/mitsuhiko/pi-codemode-mcp@…` instruction) with neutral wording along the lines of: *"Pi has no built-in MCP support. You'll need a Pi MCP extension of your choice that reads `~/.pi/agent/mcp.json` — see Pi's package listing."* No named extension, no install command.
2. Rework or drop the "Pi Agent code-mode call syntax" section in `docs/client-configuration.md` — it documents one specific extension's call surface (`call("godot-ai", …)` / `listTools(...)`), which doesn't apply if the user picks a different adapter.

The descriptor, strategy, and test changes are good as-is; this request is docs-only. Happy to re-review once the wording is neutral.

---

dsarno (changes requested):
## Full review (correctness, tests, conventions)

Complements my earlier security review (its docs-only request still stands). The merged-tier machinery itself is carefully built: Configure/status/manual all agree on "later tier wins", alias precedence matches pi's map-presence resolution, legacy-key scrubbing is generic, `_write_transaction` rollback is correct, and `pi.gd` stays genuinely data-only with a thorough descriptor pin test. Two substantiated code issues need fixing, though:

1. **Major — `_manual_command.gd:128-130` regresses the manual-instructions escape hatch for *every* JSON client.** Pre-PR, `_build_json` was pure string synthesis. Now it calls `manual_target_details`, which ends in `_read_or_init(fallback_path)` — so a corrupt/unparseable config file makes "Run this manually" degrade to "Cannot build manual configuration instructions: …". That's exactly the state where the manual text is the designated escape hatch (auto-Configure already refuses on a broken file). Fix: on `manual_target_details` failure, fall back to `resolved_path` + `client.server_key_path` (optionally prefixing the parse warning) instead of returning the error.
2. **Major (convention) — `_project_override_message` (`_json_strategy.gd:523`) hardcodes "Pi" and "godot-ai" in a shared strategy**, violating the "no per-client knowledge inside strategies" rule this PR's own doc changes restate. Any future merge-tier client gets a message about Pi, and a non-default server name is misreported. All four call sites already have `client` and `server_name` in scope — pass `client.display_name` and `server_name` in.
3. **Test gaps**: (a) the fresh-install branch (`target_index = 0`, both tiers empty/absent) is untested — the existing precedence test pre-seeds both tiers; (b) no test for an unparseable merged tier (Configure/Remove error + status ERROR branch, `_json_strategy.gd:526-540`); (c) a test asserting manual instructions survive a broken config file would have caught finding 1.

For the record, I looked hard at CodeRabbit's worker-thread finding on `_project_candidate_paths` (`_json_strategy.gd:482,486`) and consider it **not substantiated**: `DirAccess.open` per-call instances are Godot's documented thread-safe usage, this same worker path already opens `DirAccess` pre-PR (`_path_template.gd:143`, and `_base.gd`'s #691-annotated `dir_exists_absolute`), and `globalize_path` is a pure string op over a value immutable after editor startup. No change needed there.

Nits, take or leave: `pi.gd:41`'s reflective `set("server_key_path_aliases", …)` next to direct assignments of the equally-new merge fields — use direct assignment; `_load_merge_tiers` reads each file twice (capture once, parse the captured text so rollback's `original_text` can't diverge); `_remove_merged` says "configuration removed" when nothing contained the entry; the Pi code-mode docs section splits the "Per-strategy command rendering" bullet list (moot if it's removed per the security review's request).

---

dsarno:
Thanks @Rliop913 — `1ae9ac4` addresses the reviews well. Removing the endorsed-extension install in favor of neutral "choose your own Pi MCP extension" wording resolves the third-party-code concern in the cleanest possible way, and the manual-panel fallback, thread-safe project-root resolution, and new tests all check out.

Two things before the final review pass:

1. **Please merge `main` into your branch** — five PRs landed today and this branch now conflicts. Two spots to watch while reconciling:
   - `_manual_command.gd`: #880 restructured `_build_cli` (the register line now renders inside a shared shell block via `_format_shell_block`/`_render_command_line`, preceded by the scope-sweep remove lines) — your `project_roots` threading needs to merge into that new shape.
   - `test_client_attach_config.gd`: #880 added a `suite_setup`/`suite_teardown` that pins `godot_ai/mcp_client_scope`; keep those when merging your new tests in.
   - Also note #867 also bills itself as the 23rd client — whichever lands second will need to bump the counts in README/docs.

2. One open thread from the Codex bot still looks live on the current head: the dock's Open/Reveal actions and post-write diagnostics resolve `config_path()` from `path_template` alone, so they point at `~/.pi/agent/mcp.json` even when the effective entry your strategy found lives in a higher-precedence tier. Worth either routing those through the selected tier or explaining why not.

Once the branch is conflict-free I'll do the full re-review of the tier-merge rework promptly.

---

dsarno (approved):
## Re-review at `be8753a` (post main-merge)

Full pass over the reconciled branch, as promised. Approving.

**Merge reconciliation — verified clean.** #880's shell-block manual-command structure survived with your `project_roots` threading layered in correctly, the #876 scope-pinning suite setup is intact (presence-tracked restore included), and the sweep-note README text carried through.

**Tier-merge rework — reviewed in full, sound design.** The merge machinery is descriptor-gated (`config_merge_path_templates` via dynamic `get()`), so all 22 existing JSON clients are provably untouched. Configure targets the effective highest-precedence global definition; Remove clears every global tier transactionally with byte-for-byte rollback on partial failure; status verifies the effective definition; project-tier overrides fail closed with an actionable message rather than guessing Pi's cwd — the right call given the dock genuinely cannot know it. `select_server_key_path` reproduces Pi's nullish precedence exactly, including the scalar-wins subtlety. Roots are resolved on the main thread (`capture_project_roots`) and passed into workers as immutable snapshots, refreshed before each aggregate sweep.

**Independently verified:**
- `ruff` clean; `pytest` **1786 passed** (frozen-instructions update included)
- `ci-check-gdscript` OK
- Live editor suite on this head: **2193/2221 passed, 0 failed, 69/69 suites** (one earlier run showed 2 `plugin_lifecycle` failures that reproduced as my own environment's shell cmdlines matching the server fingerprint — clean on re-run, not this PR)

**Two non-blocking notes:**
1. The one remaining Codex-bot point stands: the dock's Open/Reveal and the post-write diagnostic still resolve `config_path()` from `path_template` (the lowest tier), so they can point at `~/.pi/agent/mcp.json` when the effective entry lives in `.mcp.json` next door. The manual panel now targets the right tier, and both global tiers share a directory, so I'm fine landing this as-is — but a follow-up routing those through `manual_target_details` (or a one-line "accepted, here's why") would close the loop.
2. `pi.gd` is missing its trailing newline.

Thanks for the thorough response across all three rounds — dropping the endorsed-extension install for neutral wording was exactly the right resolution.

---

dsarno:
One small addendum found after approving: the write-failure leg at `test_client_attach_config.gd:962` uses `{"path": ""}` to force `_write_transaction`'s rollback path, but `McpAtomicWrite.write("")` creates its random-named temp file in the repo root before the rename fails — every suite run leaks two untracked `test_project/-XXXXXX` files containing `cannot-write` (found them in my checkout after verifying this branch). Pointing the failure path at something like `_scratch_dir.path_join("no_such_dir/x.json")` forces the same failure without the debris. One-liner; approval stands.

---

Rliop913:
wait im still working with 23 -> 24th, and codex things

dsarno:
No rush — we'll hold the merge until you give the word. One note on the Codex float-precision finding: the full-document `JSON.stringify` round-trip with `_narrow_integral_numbers` is the existing behavior of the base JSON strategy for all 22 current clients, so the merged Remove isn't introducing a new risk class — treat that one as pre-existing repo-wide behavior rather than something this PR must solve. The root-canonicalization one is worth a look though.

---

Rliop913:
Regarding the transition from module 23 to 24, I will make the changes after the DeepSeek harness is merged into the `main` branch. The reason is that merging the DeepSeek harness will inevitably cause new conflicts, and adding module 24 without module 23 would appear to violate the documentation agreement.

dsarno:
Small clarification that may save you a wait: there's no need to hold for the DeepSeek Harness PR. The convention is simply that whichever PR lands **first** takes the 23rd slot — this PR's current "23rd client" numbering is already correct as-is, and #867 will renumber to 24th when it lands (its author will be rebasing anyway; it currently has an unaddressed review blocker, so its timeline is uncertain). No documentation agreement is violated by merging this one first.

So the only things between this PR and merge are the Codex items you're already on — particularly the project-root canonicalization one. Once you've pushed those, we're good to go.

---

Rliop913:
@codex review [b2eb301](https://github.com/hi-godot/godot-ai/pull/870/commits/b2eb301758fafd9ad13b9b52dcdf445afdb3d317) commit that fixes all findings

dsarno (changes requested):
## Review of `b2eb301` (F1–F5 fixes)

The F1–F4 responses look right — last-definition-wins on project tiers, root canonicalization with a test seam, `effective_config_path` routing for Open/Reveal, and the `external_client_cwd` escape hatch are all good resolutions. But the F5 text surgeon (`_text_remove_server_entry`) has **two confirmed bugs** I verified by simulating the algorithm exactly as written:

**1. Removing a middle entry produces invalid JSON (severe).** The comma trim is unconditional on both sides: it walks back over a preceding comma *and* forward over a trailing comma. For an entry that is neither first nor last, both commas are consumed:

```
input:  {"mcpServers": {"a": {"command": "x"}, "godot-ai": {"command": "y"}, "b": {"command": "z"}}}
output: {"mcpServers": {"a": {"command": "x"}"b": {"command": "z"}}}   ← invalid JSON
```

Pi then rejects the whole file — the exact corruption class F5 exists to prevent, now reachable from the dock's Remove button on any config with ≥3 entries where `godot-ai` sits in the middle. Fix: make the trims mutually exclusive — consume the **trailing** comma when one exists, and only fall back to consuming the leading comma when there is no trailing one (i.e., the entry was last). Please add first/middle/last test legs; the current suite apparently only exercises first/last shapes.

**2. The scanner escapes its container and can delete from an unrelated sibling (silent data loss).** `_find_key_at_container_depth` keeps scanning after `inner_depth` goes negative (the target container closed), so a later sibling container re-balances the count back to 0 and its keys match:

```
input:  {"mcpServers": {"other": {...}}, "extensions": {"godot-ai": {"pinned": true}}}
remove: godot-ai from mcpServers
output: {"mcpServers": {"other": {...}}, "extensions": {}}   ← extensions.godot-ai silently deleted
```

Fix: terminate the scan (return `{}`) as soon as `inner_depth` drops below 0. Same guard applies to the `key_path` descent loop. (Relatedly: the `container_depth` parameter is never read in the function body — the depth semantics are all relative `inner_depth`. Worth deleting the parameter or the docstring claim so the next reader isn't misled.)

Everything else in the commit holds up, and the token-preserving approach is the right idea — it just needs these two scanner fixes plus middle-entry coverage before Remove can be trusted with real files.

---

dsarno:
Re-verified `006197f` by simulating the updated algorithm: **both findings from my review are correctly fixed** — middle/first/last removals all produce valid JSON, and the sibling-escape case now leaves the file untouched (the `inner_depth < 0` bail-out works). BOM skip and `authoritative_tier_path` both look right too.

One last one-liner before I can re-approve — Codex's round-3 finding is real, verified with the same simulation: `had_trailing_comma` probes the exact position after `value_end` without skipping whitespace, so valid-but-unusual formatting with a space before the comma corrupts on removal:

```
input:  {"mcpServers": {"godot-ai": {}   , "other": {}}}
output: {"mcpServers": {, "other": {}}}   ← invalid JSON
```

Fix: scan past `_is_json_ws` from `value_end` first, then check for the comma at that position (and when consuming it, remove the intervening whitespace too). With that plus a test leg for space-before-comma, this is ready.

---

dsarno (approved):
## Re-approval at `788ba2b`

Re-verified the whitespace-comma fix by simulation across nine shapes — first/middle/last/only entries, whitespace-before-comma in both middle and last position, pretty-printed, BOM-prefixed, and the sibling-escape guard: **all produce valid JSON with the entry gone (or the file untouched where it should be)**. The test legs landed too. The text surgeon is now sound against every corruption class found across four review rounds.

On Codex's round-4 P2s, my disposition as reviewer:

1. **Escaped property names** (`"godot-ai"`): the failure mode is fail-safe — Remove no-ops and post-verify reports it loudly, nothing corrupts. Real-world configs don't escape ASCII hyphens, and teaching the scanner a JSON string decoder buys almost nothing for meaningful complexity. Accept as a documented limitation (a one-line comment on `_text_remove_server_entry` would do); no code change needed.
2. **Stale cached launch context vs refreshed roots in the status worker**: legitimate but a settings-lag nit on a brand-new opt-in setting — fine as a follow-up, not a blocker.

Four rounds of bot findings have converged from file corruption → fail-safe no-op → refresh lag; that's the right place to stop iterating. Approving — thanks for the persistence, this landed in genuinely robust shape.

---

dsarno:
Merged — thank you @Rliop913! Pi Agent is the 23rd supported client. This one went through more review rounds than most, and you met every single finding — from the third-party-install rework through the tier-merge design to the text-surgeon hardening — with real fixes and real tests. The final Remove path is now verified against every corruption shape three reviewers could throw at it. Genuinely appreciated.

---

Issue #463, opened 2026-05-26
Claude Code running as (VS)Code Extension undetected


TheYellowArchitect:
I use Code, and have Claude Code as an extension.
I installed godot-ai plugin, and on configure list, I tried to enable:
Claude Code
Claude Desktop
VSCode

Claude Code was red, the other 2 became green.
After troubleshooting with Claude, it said it didn't have access to the MCP through VSCode.
But it did this edit in ~/.claude.json  (user scope)
`"mcpServers": { "godot-ai": { "type": "http", "url": "http://127.0.0.1:8000/mcp" } }` and it works now.

So I suggest adding this as a feature on setting it up when Claude Code becomes red.

dsarno:
Thanks for the detailed report, @TheYellowArchitect! This is fixed in #504 — the client configurator now falls back to `~/.claude.json` when the `claude` CLI isn''t on PATH (the Claude-Code-as-VS-Code-extension case you hit).

It''s live as of **v2.6.0**: https://github.com/hi-godot/godot-ai/releases/tag/v2.6.0 — update via the dock / AssetLib and the extension should be detected. Appreciate you flagging it! 🙏

PR #881, opened 2026-08-24
clients: #873 follow-ups — sweep transparency, scope-aware hints, hermetic scope tests

### LINKED ISSUE #876: tests: setting godot_ai/mcp_client_scope to project fails the GDScript suite
### Summary

Setting the new `godot_ai/mcp_client_scope` EditorSetting (#872 / #873) to `project` —
its headline value — makes the GDScript suite fail. The failure is in
`test_client_attach_config.gd`, a suite that has nothing to do with the scope feature
and never opts into it.

A developer who actually adopts the feature gets a red suite with no obvious connection
to the setting they changed.

### Reproduction

Verified on Godot `4.7.2.stable` (`ed1daf0bf`), headless editor, plugin `3.1.5`,
at `143e3ba` (current `main`):

| `godot_ai/mcp_client_scope` | full suite |
|---|---|
| `user` (default) | 2161 passed / **0 failed** / 36 skipped |
| `project` | 2160 passed / **1 failed** / 36 skipped |

```
client_attach_config.test_claude_cli_json_fallback_migrates_legacy_http_entry
  Expected 2, got 0
```

(`2` = `CONFIGURED_MISMATCH`, `0` = `NOT_CONFIGURED`.)

### Root cause

`test_client_attach_config.gd:813` `_claude_cli_clone()` builds its fixture by copying
fields off the **real** `claude_code` descriptor:

```gdscript
client.cli_register_template = registered.cli_register_template
client.cli_unregister_template = registered.cli_unregister_template
client.cli_status_args = registered.cli_status_args
```

Since #873 `cli_register_template` carries `{scope}`, so the clone now reports
`CliStrategy.uses_scope_token() == true`. It does **not** copy the new
`cli_scope_status_template`.

At `project` scope, in `_dispatch_check_status_with_cli_path_details`:

1. `_scope_diverges_from_json_fallback` (`client_configurator.gd:762`) returns `true`
   — token present, scope is not the default, and the test passes a non-empty
   `cli_path` (`"/fake/claude-not-invoked"`).
2. So the JSON-fallback read at `client_configurator.gd:672-677` is skipped — which is
   the whole point of that test ("proving the JSON file is authoritative for
   command-shape CLI clients").
3. The scope-probe gate at `client_configurator.gd:694-700` requires a non-empty
   `cli_scope_status_template`, which the clone lacks, so it does not fire either.
4. Control reaches `CliStrategy.check_status_details` with a binary that does not
   exist → `spawn_failed` → `NOT_CONFIGURED`.

`test_client_attach_config.gd` contains zero references to `SETTING_CLIENT_SCOPE`, so it
neither pins nor restores the setting. `test_clients.gd` has `_restore_client_scope()`,
but it restores the developer's *saved* value rather than the default — so an adopted
`project` setting persists across the run.

### Suggested fix

Make the suite hermetic rather than dependent on ambient editor state: pin
`godot_ai/mcp_client_scope` to `McpSettings.DEFAULT_CLIENT_SCOPE` for the duration of
`test_client_attach_config.gd` and restore the previous value in teardown.

More generally, any suite that clones a real descriptor is now exposed to this setting;
a shared save/pin/restore helper would stop the next one silently inheriting it.

### Environment

- godot-ai `3.1.5` @ `143e3ba`
- Godot `4.7.2.stable (official)`
- Windows 11

### LINKED ISSUE #877: clients: Configure's all-scope sweep silently rewrites .mcp.json in the editor's cwd, and the dock never shows the removes
### Summary

Configure's all-scope pre-cleanup (#873) runs `claude mcp remove --scope project godot-ai`
on **every** Configure, including at the default `user` scope. That command deletes the
`godot-ai` key from the `.mcp.json` in the spawned CLI's working directory — which is
whatever directory the editor process happened to inherit at launch, not necessarily the
Godot project.

The sweep itself is deliberate and documented (`README.md:122`: "clears `godot-ai` out of
every scope before writing the new one"). What is missing is any connection between that
line and the uncontrollable-cwd caveat documented separately, plus any surfacing of the
removes in the UI.

### Reproduction

Real `claude` 2.1.241, isolated `CLAUDE_CONFIG_DIR`, a throwaway git repo with a
committed `.mcp.json`:

```json
{
  "mcpServers": {
    "godot-ai": { "type": "stdio", "command": "uvx",
                  "args": ["godot-ai", "attach", "--hand-tuned-by-the-team"] },
    "some-other-server": { "type": "stdio", "command": "node", "args": ["other.js"] }
  }
}
```

```
$ claude mcp remove --scope project godot-ai
Removed MCP server godot-ai from project config
File modified: .../fakerepo/.mcp.json
exit 0
```

Result: the hand-tuned `godot-ai` entry is gone, the file is dirty in git, and
`some-other-server` was reformatted as collateral. Exit code is 0, and
`_cli_strategy.gd:57` discards the result, so the plugin cannot notice.

### Where

`plugin/addons/godot_ai/clients/_cli_strategy.gd:45-57` — the loop is guarded only on a
non-empty unregister template. There is no scope-equality check:

```gdscript
for pre_scope in _cleanup_scopes(client):
    var pre_args := _format_args(
        client.cli_unregister_template, server_name, server_url, {}, pre_scope
    )
    McpCliExec.run(cli, pre_args, _REMOVE_TIMEOUT_MS)
```

`_cleanup_scopes` returns all of `CLIENT_SCOPES` for any `{scope}` descriptor, pinned by
`test_clients.gd` `test_client_scope_pre_cleanup_sweeps_every_scope`. `claude_code.gd:23-30`
concedes the cwd problem: *"`--scope project` resolves `.mcp.json` against the spawned
CLI's working directory ... Godot has no API to set a child's cwd."*

### Scope of the problem (what it is *not*)

- Only the `project` pass can touch a version-controlled file. `--scope local` writes a
  per-project block inside `~/.claude.json`; it does not touch `.mcp.json`.
- Blast radius is one key named exactly `godot-ai`, not the whole file.
- `mcp remove` on an absent entry really is a harmless no-op, as the comment claims.

So this is not "Configure corrupts your repo". It is: at the default setting, Configure
has a silent, uncommunicated write to a file in an unrelated directory that a team may
be tracking in git.

### Suggested fix

Transparency rather than behaviour change:

1. **Show the removes in the dock's "Run this manually" text.**
   `clients/_manual_command.gd` `_build_cli` renders only `cli_register_template`, so the
   snippet a user is told to run is not what Configure runs. Rendering the sweep lines
   makes the side effect self-documenting and keeps the manual path faithful.
2. **Connect the two README bullets** so the sweep and the cwd caveat are read together.
3. Optionally, skip the `project` pass when the selected scope is `user` and no
   `{scope}` entry was ever written at project scope — though this trades the
   stranded-entry guarantee the sweep exists to provide, so (1) and (2) may be enough.

### Environment

- godot-ai `3.1.5` @ `143e3ba`
- Claude Code CLI `2.1.241`, Windows 11

### LINKED ISSUE #878: tests: scope-aware status probe never executes in the suite; two mutations survive
### Summary

The scope-aware status probe added in #873 — `CliStrategy.check_scope_status_details`
(`_cli_strategy.gd:247-275`) and its dispatch gate (`client_configurator.gd:694-700`) —
never executes during the test suite. Only the pure helpers `_scope_probe_verdict` and
`_scope_from_probe_output` are covered.

### Evidence

Instrumented a full headless run on Godot `4.7.2.stable`, plugin `3.1.5` @ `143e3ba`,
and counted the synthetic client's subprocess spawns in the editor log:

```
godot-ai-nonexistent-cli-xyz mcp list   3
godot-ai-nonexistent-cli-xyz mcp get    0
```

`mcp list` is `cli_status_args`; `mcp get` is `cli_scope_status_template`. Zero `mcp get`
spawns means the scope-probe branch was never taken.

Cause: `test_clients.gd:1531` `_make_scope_cli_client()` sets `cli_register_template`,
`cli_unregister_template` and `cli_status_args`, but never `cli_scope_status_template`.
The gate at `client_configurator.gd:694-700` requires that field to be non-empty, so
control always falls through to the plain `check_status_details`.

### Mutations that survive the whole suite

Both silently disable scope detection for the only client that has the feature:

1. `client_configurator.gd:699` — `McpSettings.client_scope()` → `McpSettings.DEFAULT_CLIENT_SCOPE`
   (the probe would always compare against `user`).
2. `_cli_strategy.gd:264` — `client.cli_scope_status_template` → `client.cli_status_args`
   (the probe would run `mcp list`, whose output has no `Scope:` line, and every verdict
   would degrade to the target check).

### Fairness note

Most of `check_scope_status_details` is a structural clone of the already-covered
`check_status_details` — same `cli.is_empty()` guard, same `command_launch_error`
fail-close, same `timed_out` / `spawn_failed` branches. The real decisions live in
`_scope_probe_verdict`, which *is* tested against verbatim `claude mcp get` output. So
this is duplicated plumbing left uncovered rather than a body of untested logic — but the
two mutations above are genuine holes, and the gate itself is what routes them.

### Suggested fix

- Set `cli_scope_status_template` on `_make_scope_cli_client` so the dispatch gate is
  actually reachable from a test, and assert the routed spawn-failure path.
- Add a case that would fail under mutation (1) — e.g. assert the probe is invoked with
  the *selected* scope rather than the default.
- Consider asserting on the rendered argv (`format_args(cli_scope_status_template, ...)`)
  so mutation (2) is caught without needing a live binary.

### Related

Coverage gaps in the same area, for whoever picks this up: `client_scope()`'s
`value == null` branch (`settings.gd:99`), `_client_scope_setting()`'s worker-thread
branch, `warm_client_scope()`, `format_args`'s `scope_override` parameter, the
`configure` all-scope cleanup loop, and `_note_unhonoured_scope`'s token guard all have
no direct test.

### LINKED ISSUE #879: cleanup after #873: dead PATH-walk clause, wrong path hint on cross-scope mismatch, stale descriptor comments
Three small follow-ups from reviewing #873 after merge. None change behaviour except (2).

### 1. `_scope_diverges_from_json_fallback`'s PATH walk is dead code, and its docstring says otherwise

`client_configurator.gd:767-768`:

```gdscript
var resolved := cli_path if not cli_path.is_empty() else CliStrategy.resolve_cli_path(client)
return not resolved.is_empty()
```

The docstring above it (`:757-761`) says *"the PATH walk only runs when the answer can
actually be true"*, implying the clause is load-bearing. It cannot change any outcome:

- The function is only consulted from `client_configurator.gd:674`, inside a condition
  that already requires `client.has_json_fallback()`.
- If the walk returns empty, the function returns `false` and the JSON read happens at
  `:672-677`.
- If it is forced to `true` instead, the JSON read is skipped, but `:682`
  (`if resolved_cli.is_empty() and client.has_json_fallback():`) takes the fallback
  anyway — same status, same file.

So the clause costs a second full `McpCliFinder.find` PATH sweep per status check (the
same walk runs again at `:679`) and buys nothing. Either drop it and return `true`, or
thread the already-resolved value down instead of re-walking. Either way the docstring
needs to stop claiming it is decisive.

### 2. Verification error names the wrong config file after a cross-scope mismatch

`client_configurator.gd:809`:

```gdscript
var path := client.resolved_config_path()
var path_hint := "" if path.is_empty() else " Inspect %s and remove the godot-ai entry by hand if needed." % path
```

`resolved_config_path()` is `path_template`, i.e. `~/.claude.json`. When the surviving
entry is the *project*-scope one, the message tells the user to inspect `~/.claude.json`
where there is nothing to find — the entry is in `<cwd>/.mcp.json`.

The probe already knows which scope actually resolved (`_scope_probe_verdict` puts it in
`error_msg`, e.g. `"registered at user scope, not project"`), so the hint can name the
right location instead of assuming the fallback file.

### 3. Stale descriptor comments in `clients/claude_code.gd`

These describe pre-#873 behaviour and now contradict the code they sit on:

- `:20-21` — *"Explicit scope: an unscoped `mcp remove` deletes from whichever scope
  matches first, which could eat a project-local entry the user made"* — sits directly
  above `cli_unregister_template`, which the Configure sweep now runs at **every** scope,
  including that project-local one.
- `:46` and `:56-57` — still spell the command as `claude mcp add --scope user`.
- `:51-53` — *"Status always reads this file — it is the CLI's own store for user scope"*
  — false at project/local scope, where `_scope_diverges_from_json_fallback` routes status
  to the CLI probe instead.

These are the primary documentation for the descriptor and were left untouched while
every line around them changed.

dsarno:
Fixes the four follow-up issues @azereki filed after #873 merged. Every specific claim in them was verified against `origin/main` (`143e3ba`) with file:line evidence before fixing — 12 of 13 confirmed exactly as written; the one nuance is noted under #879 below.

## #876 — suite red at `mcp_client_scope = project`

`test_client_attach_config.gd` now snapshots the setting in `suite_setup`, pins it to `McpSettings.DEFAULT_CLIENT_SCOPE` for the suite, and restores it in `suite_teardown`. The pin (rather than copying `cli_scope_status_template` into the clone) is deliberate: those tests exist to prove the JSON fallback file is authoritative, and that premise only holds at the default scope.

Regression proof: the full GDScript suite was run twice against a live editor — once at default scope and once with the ambient setting at `project`. Both green (2173/2200 passed, 0 failed; skips are environment-conditional).

## #877 — silent all-scope sweep (transparency only, no behavior change)

- The dock's "Run this manually" text now renders the pre-cleanup's remove commands (one per swept scope) ahead of the register line, via a new public `McpCliStrategy.cleanup_scopes()` view of the existing private helper — the snippet a user is told to run is now what Configure actually runs.
- The README's sweep sentence now states that the sweep runs on every Configure at any scope selection and links it to the CLI-cwd caveat it was previously divorced from.
- The sweep's behavior is intentionally unchanged — the stranded-entry guarantee from #873 stays, matching the issue's own lean toward option (1)+(2).

## #878 — scope probe now executes in the suite

- `_make_scope_cli_client()` gains `cli_scope_status_template`, making the probe gate reachable from a fixture.
- New `test_dispatch_routes_scope_status_through_probe_with_selected_scope` drives the real subprocess path with a fake recording CLI (a `.bat` on Windows — `McpCliExec` already wraps those in `cmd.exe /c` — a shebang script elsewhere) that echoes the recorded `claude mcp get` fixtures. It kills both surviving mutations from the issue: the fake CLI prints a user-scope entry, so a hardcoded default expected-scope reads green where the selected `project`/`local` scope must read `CONFIGURED_MISMATCH`; and the recorded argv is asserted to be `mcp get godot-ai`, not `cli_status_args`' `mcp list`.
- `test_project_scope_status_leaves_the_user_scope_file` got a comment noting it now exercises the scope-probe spawn-fail path rather than the plain probe (same verdict, different route), per the caveat in the issue analysis.

## #879 — cleanup trio

1. **Dead PATH walk dropped.** `_scope_diverges_from_json_fallback` loses the `resolve_cli_path` clause and its `cli_path` parameter; the docstring no longer claims the walk is decisive. When the CLI is unresolvable, the caller's #463 fallback branch reads the same file with identical arguments, so the clause could not change any outcome — it only cost a duplicate `McpCliFinder.find` per status check.
2. **Scope-aware verify hint.** The probe's mismatch verdict now carries `resolved_scope` as a structured key (no string-parsing of `error_msg`), and `_verify_post_state` picks its hint from it: a `project`-scope survivor points at the `.mcp.json` in the editor's launch directory (phrased, since `path_template` has no project-relative token), `local` at the per-project block in `~/.claude.json`, and the default keeps the historical `resolved_config_path()` hint. Pinned by `test_post_state_path_hint_follows_resolved_scope` plus the `resolved_scope` assertion in the #878 test.
3. **`claude_code.gd` comments refreshed.** The `--scope user` examples and the "Status always reads this file" claim now describe post-#873 behavior. One deviation from the issue: the `:20-21` unregister comment was **not** stale — it is the live rationale for the explicit `--scope` flag — but it was contradicted by the all-scope sweep, so it now names the sweep as the sanctioned exception instead of being deleted.

## Verification

- `ruff check` clean; `pytest`: 1780 passed, 5 skipped (no Python changes; confirms no breakage).
- `script/ci-check-gdscript`: all files OK.
- Live `script/ci-godot-tests` against a connected editor: **2173/2200 passed, 0 failed** at default scope **and** with ambient `mcp_client_scope = project` (the #876 repro).

Closes #876. Closes #877. Closes #878. Closes #879.

https://claude.ai/code/session_01HL8hj91sBJdWj8ZoXkmrDz

---
_Generated by [Claude Code](https://claude.ai/code/session_01HL8hj91sBJdWj8ZoXkmrDz)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved client configuration checks for project and local scopes.
  * Corrected post-configuration status reporting and diagnostic file paths.
  * Ensured stale registrations are removed across configured scopes.

* **Improvements**
  * Manual setup commands now include accurate cleanup commands in the correct order.
  * Scope-specific behavior and mismatch details are reported more clearly.

* **Documentation**
  * Clarified Configure sweep behavior, scope handling, cleanup commands, and project-scope caveats in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,10 +120,15 @@ opened on that project. `local` is also accepted for clients that support it.
 The default stays `user` so existing installs are unchanged on upgrade. To move
 an existing entry, change the setting and press **Configure** again — Configure
 clears `godot-ai` out of every scope before writing the new one, so the old
-entry doesn't linger. **Remove**, by contrast, only targets the scope that is
-currently selected, so switch back before removing if you changed the setting.
+entry doesn't linger. That sweep runs on *every* Configure, whatever scope is
+selected, and its `project` pass resolves `.mcp.json` against the CLI's working
+directory (first caveat below) — so a `godot-ai` entry in that file is removed
+even when you only ever use `user` scope. The dock's "Run this manually" text
+shows the exact remove commands the sweep runs. **Remove**, by contrast, only
+targets the scope that is currently selected, so switch back before removing if
+you changed the setting.
 
-Two caveats for `project` scope:
+Caveats for `project` scope:
 
 - The client CLI resolves the project config against **its own working
   directory**, which is the one the Godot editor was launched from — not
--- a/plugin/addons/godot_ai/client_configurator.gd
+++ b/plugin/addons/godot_ai/client_configurator.gd
@@ -672,7 +672,7 @@ static func _dispatch_check_status_with_cli_path_details(
 			if (
 				client.command_shape != Client.CommandShape.NONE
 				and client.has_json_fallback()
-				and not _scope_diverges_from_json_fallback(client, cli_path)
+				and not _scope_diverges_from_json_fallback(client)
 			):
 				var command_launch := _resolved_or_discovered_launch(client, resolved_launch, launch_context)
 				return JsonStrategy.check_status_details(client, SERVER_NAME, url, command_launch)
@@ -754,18 +754,15 @@ static func _note_unhonoured_scope(client: Client, result: Dictionary) -> Dictio
 ## detection — a stdout scan sees the command, not the full argv — which is the
 ## documented trade for a status that is merely coarse instead of wrong.
 ##
-## Ordering matters: the token check and the scope compare both short-circuit
-## for every non-`{scope}` descriptor and for the default user scope, so the
-## PATH walk only runs when the answer can actually be true. An unresolvable
-## CLI means configure took the #463 JSON fallback, which writes user scope
-## whatever the setting says — so the file is the right thing to read again.
-static func _scope_diverges_from_json_fallback(client: Client, cli_path: String) -> bool:
+## No CLI-resolvability check here (#879): when the CLI is unresolvable —
+## meaning configure took the #463 JSON fallback, which writes user scope
+## whatever the setting says — the caller's own fallback branch reads the
+## file with identical arguments, so re-walking PATH to pre-empt it changed
+## no outcome and cost a duplicate McpCliFinder.find per status check.
+static func _scope_diverges_from_json_fallback(client: Client) -> bool:
 	if not CliStrategy.uses_scope_token(client):
 		return false
-	if McpSettings.client_scope() == McpSettings.DEFAULT_CLIENT_SCOPE:
-		return false
-	var resolved := cli_path if not cli_path.is_empty() else CliStrategy.resolve_cli_path(client)
-	return not resolved.is_empty()
+	return McpSettings.client_scope() != McpSettings.DEFAULT_CLIENT_SCOPE
 
 
 static func _resolved_or_discovered_launch(
@@ -793,13 +790,13 @@ static func _verify_post_state(
 ) -> Dictionary:
 	if result.get("status") != "ok":
 		return result
-	var actual := _dispatch_check_status_with_cli_path_details(
+	var details := _dispatch_check_status_with_cli_path_details(
 		client, url, "", {}, resolved_launch
-	).get("status", Client.Status.NOT_CONFIGURED)
+	)
+	var actual: Client.Status = details.get("status", Client.Status.NOT_CONFIGURED)
 	if actual == expected:
 		return result
-	var path := client.resolved_config_path()
-	var path_hint := "" if path.is_empty() else " Inspect %s and remove the godot-ai entry by hand if needed." % path
+	var path_hint := _post_state_path_hint(client, str(details.get("resolved_scope", "")))
 	return {
 		"status": "error",
 		"message": "%s reported %s ok but verification still reads %s (expected %s).%s" % [
@@ -808,6 +805,33 @@ static func _verify_post_state(
 			path_hint,
 		],
 	}
+
+
+## Where to point the user when verification finds a surviving entry. The
+## scope probe's mismatch verdict carries `resolved_scope` (#879); without it,
+## `resolved_config_path()` (`path_template`) is the right file only for the
+## default user scope — naming ~/.claude.json for a project-scope survivor
+## sends the user to a file where there is nothing to find. `path_template`
+## has no project-relative token (see _note_unhonoured_scope), so the
+## project hint is phrased rather than resolved.
+static func _post_state_path_hint(client: Client, resolved_scope: String) -> String:
+	match resolved_scope:
+		"project":
+			return (
+				" The surviving entry is project-scoped: inspect the .mcp.json in the"
+				+ " directory the editor was launched from and remove the godot-ai entry"
+				+ " by hand if needed."
+			)
+		"local":
+			return (
+				" The surviving entry is local-scoped: inspect this project's block in %s"
+				+ " and remove the godot-ai entry by hand if needed."
+			) % client.resolved_config_path()
+		_:
+			var path := client.resolved_config_path()
+			if path.is_empty():
+				return ""
+			return " Inspect %s and remove the godot-ai entry by hand if needed." % path
 
 
 static func manual_command(id: String) -> String:
--- a/plugin/addons/godot_ai/clients/_cli_strategy.gd
+++ b/plugin/addons/godot_ai/clients/_cli_strategy.gd
@@ -320,10 +320,15 @@ static func _scope_probe_verdict(
 	if resolved.is_empty():
 		return _status_details(McpClient.Status.CONFIGURED)
 	if resolved != expected_scope:
-		return _status_details(
+		## `resolved_scope` rides along as a structured key so callers
+		## (`_verify_post_state`'s path hint) don't have to parse it back out
+		## of the human-facing message (#879).
+		var details := _status_details(
 			McpClient.Status.CONFIGURED_MISMATCH,
 			"registered at %s scope, not %s" % [resolved, expected_scope],
 		)
+		details["resolved_scope"] = resolved
+		return details
 	return _status_details(McpClient.Status.CONFIGURED)
 
 
@@ -333,6 +338,12 @@ static func _scope_probe_verdict(
 ## the JSON-fallback file is still a valid place to read status back from.
 static func uses_scope_token(client: McpClient) -> bool:
 	return client.cli_register_template.has(SCOPE_TOKEN)
+
+
+## Public view of the pre-cleanup sweep for the manual-command text, so what
+## the dock tells the user to run matches what Configure actually runs (#877).
+static func cleanup_scopes(client: McpClient) -> Array[String]:
+	return _cleanup_scopes(client)
 
 
 ## Scopes the configure pre-cleanup removes from. A descriptor without the
--- a/plugin/addons/godot_ai/clients/_manual_command.gd
+++ b/plugin/addons/godot_ai/clients/_manual_command.gd
@@ -66,13 +66,17 @@ static func _build_cli(
 			var parts: Array[String] = [short_name]
 			for arg in args:
 				parts.append(String(arg))
-			cmd = _format_shell_command(parts, shell_kind)
+			var lines := _sweep_lines(client, server_name, server_url, short_name, shell_kind)
+			lines.append(_render_command_line(parts, shell_kind))
+			cmd = _format_shell_block(lines, shell_kind)
 	else:
 		var args := McpCliStrategy.format_args(client.cli_register_template, server_name, server_url)
 		var parts: Array[String] = [short_name]
 		for arg in args:
 			parts.append(String(arg))
-		cmd = _format_shell_command(parts, shell_kind)
+		var lines := _sweep_lines(client, server_name, server_url, short_name, shell_kind)
+		lines.append(_render_command_line(parts, shell_kind))
+		cmd = _format_shell_block(lines, shell_kind)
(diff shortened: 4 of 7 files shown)


azereki:
Woops. @dsarno Did we stomp each other with [#880](https://github.com/hi-godot/godot-ai/pull/880)

dsarno:
Oops no worries go ahead with yours.  Thank you.

dsarno:
Closing in favor of #880, which was opened first and covers the same four issues — we fixed them in parallel and landed on essentially the same design on every point (suite-pinned default scope, sweep lines rendered in the manual text, `cli_scope_status_template` on the probe fixture, dropped PATH walk, probe-informed verify hint). #880's presence-tracked EditorSetting restores are a nicer touch than what this branch did, so it's the better version to land.

One small thing worth carrying over to #880: CodeRabbit flagged here that `_scope_probe_verdict()` returns on **target drift** before parsing the `Scope:` line. In #880's `_verify_post_state`, that path leaves `error_msg` empty, so a project-scoped entry with a stale launcher still falls back to the `~/.claude.json` hint. Parsing the scope before the target check (and including it in that mismatch verdict too) closes the gap — a one-liner plus a test leg.

Thanks @azereki — and sorry for the stomp!

---

PR #884, opened 2026-08-24
ci: let the ci-* runners target a non-default MCP port, document the headless opt-in

azereki:
Two gaps found while running the verification gauntlet for #880 on a machine whose editor pins a non-default port.

### `ci-*` runners hard-coded `:8000`

`ci-godot-tests`, `ci-quit-test`, `ci-reload-test`, and `ci-slow-suite-smoke` all pinned `http://127.0.0.1:8000/mcp`. Two failure modes:

- An editor whose `godot_ai/http_port` is anything else connects fine to its own server, and the runner still reports `No Godot session connected after 60 attempts` — it was polling a different server the whole time. A healthy editor log next to that error is the telltale.
- Worse: if some *other* editor is on 8000, the runner drives it. The multi-session pin guard only fires past one session, so a single foreign session is driven without question — during verification here, a run at the default port reached a live game-project editor on 8000 and sent it `scene_open main.tscn` (the play-mode guard refused it, so nothing was touched, but only because that editor happened to be playing).

`script/_ci_env.sh` now exports `MCP_SERVER_URL` (default unchanged, so CI needs no changes), the four bash runners consume it, and `ci-godot-tests` names the URL in its wait/no-session messages so the mismatch is self-diagnosing. The variable name is the one `ci-game-capture-smoke` and `local-game-capture-diag` already read.

### `docs/verification.md` omitted the headless opt-in

The doc's headless launch line lacked `GODOT_AI_ALLOW_HEADLESS=1`; without it the plugin logs `MCP | plugin disabled in headless mode`, starts no server, and nothing ever connects. The session-start hook and `ci.yml` both set it — the doc a fresh contributor follows did not. The launch line now carries the opt-in plus `GODOT_AI_DISABLE_TELEMETRY=true` (mirroring `_ci_env.sh`), with a short explanation and a pointer to [docs/server-lifecycle.md](docs/server-lifecycle.md). The doc also documents the `MCP_SERVER_URL` override and its symptom.

### Verification

Godot 4.7.2-stable (`ed1daf0bf`), headless editor pinned to `:8123`:

- `MCP_SERVER_URL=http://127.0.0.1:8123/mcp script/ci-godot-tests` — **2164/2198 passed, 0 failed, 34 skipped, 69/69 suites** (the #883 suite-count guard active)
- default resolution unchanged: sourcing `_ci_env.sh` with nothing set yields `http://127.0.0.1:8000/mcp`; with the var exported, the override wins
- `bash -n` clean on all five scripts; `ruff` clean; `pytest` 1766 passed / 19 skipped

### Possible follow-up (not in this PR)

The single-foreign-session case above suggests `ci-godot-tests` should also verify the connected session's `project_path` matches this checkout's `test_project/` before `scene_open`, instead of trusting any lone session. Left out to keep this change to the two gaps it fixes; can file an issue if wanted.

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring the MCP server URL when the editor uses a non-default port.
  - Automated checks now connect to the configured endpoint and provide clearer connection and session error guidance.

- **Documentation**
  - Documented the server URL configuration and troubleshooting steps.
  - Updated headless startup instructions to enable headless operation and disable telemetry.

### DIFF
--- a/docs/verification.md
+++ b/docs/verification.md
@@ -31,6 +31,19 @@ script/ci-godot-tests    # waits for the plugin, opens main.tscn, runs test_run,
                          # prints {suite}.{test}: {message} for each failure
 ```
 
+The probe above and every `script/ci-*` runner assume the server is on
+`:8000`, the plugin's default. If the connected editor's `godot_ai/http_port`
+is something else — the editor log names it in its
+`MCP | started server ... --port N` line — point them at it:
+
+```bash
+MCP_SERVER_URL=http://127.0.0.1:8123/mcp script/ci-godot-tests
+```
+
+The symptom of forgetting is a healthy editor log next to
+`No Godot session connected after 60 attempts`: the runner was polling a
+different server the whole time.
+
 **With several editors connected, pin the one you mean.** Multiple editors
 sharing port 8000 is a supported setup (see [worktrees](worktrees.md)), and
 `scene_open` would otherwise land in whichever session is active — yanking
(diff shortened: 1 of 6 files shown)


azereki:
Follow-up filed as #885 (lone-foreign-session project_path check across the ci-* runners).

dsarno (approved):
Reviewed the full diff at `29d8c4c`. Approving.

Both gaps are real — I hit each of them independently in my own environment this week: the missing `GODOT_AI_ALLOW_HEADLESS=1` cost me a silent "plugin disabled in headless mode" no-connect exactly as described, and the hardcoded `:8000` is precisely the polling-the-wrong-server trap the error messages now self-diagnose. The foreign-session near-miss (a lone session on 8000 getting driven without the multi-session guard firing) is a genuinely valuable catch, and scoping the `project_path` verification out as a follow-up was the right call — please do file that issue.

Verified: `bash -n` clean on all five scripts against the PR head; the `${MCP_SERVER_URL:-…}` default keeps CI byte-identical; all four runners consume the single `_ci_env.sh` definition; the branch is based on current main so it composes with the #883 suite gate.

On CodeRabbit's consistency finding: I'd treat it as a scoping question rather than a defect. `ci-start-server` starts its *own* server on 8000 by design — the override exists for the connect-to-an-existing-editor workflow where `ci-start-server` never runs, and the doc already frames it that way. If you want to close the loop cheaply, a one-line note in `verification.md` ("non-default ports assume the editor started the server; `ci-start-server` always uses :8000") would do; making `ci-start-server` port-aware is fine as a follow-up but shouldn't hold this.

---

dsarno:
@azereki If you haven't already, come to our discord and drop your username so we can give you a shout in upcoming announcement. Thanks again

PR #892, opened 2026-08-27
feat(resource): generate physics shapes for 3D meshes

### LINKED ISSUE #868: Proposal: collision_manage MCP tool for bulk 3D mesh collider generation
## Summary

Add a new core MCP tool, `collision_manage`, with one op: `generate`.

For each selected `MeshInstance3D`, it creates a physics-body sibling
(`StaticBody3D` / `RigidBody3D` / `CharacterBody3D` / `Area3D`) with a
`CollisionShape3D` sized to the mesh's world-space AABB. This is the
counterpart to the existing `resource_manage.physics_shape_autofit`
(which only sizes an existing shape) — this tool creates the full
body + shape hierarchy in one call.

## Motivation

Imported 3D models often arrive as many `MeshInstance3D` leaves under a
skeleton or rig. Agents today have to create a `StaticBody3D`, add a
`CollisionShape3D`, pick a shape type, and size it manually for every
mesh. A single bulk command matches how level-builders actually use the
engine.

## Proposed API

```json
{
  "op": "generate",
  "params": {
    "paths": ["/Root/Body", "/Root/Wing"],
    "shape_type": "box",
    "body_type": "static"
  }
}
```

- `paths`: scene paths to `MeshInstance3D` nodes (required).
- `shape_type`: `box` | `sphere` | `capsule` | `cylinder` (default `box`).
- `body_type`: `static` | `rigid` | `character` | `area` (default `static`).

Returns `{created: [...], undoable: true}`.

## Key details

- The generated body is positioned/rotated to match the mesh transform,
  but scale is **not** copied to the body — the shape size is already
  computed from the world-space AABB, so copying scale would double-apply it.
- Supports meshes whose geometry is offset from the node origin.
- All writes go through `EditorUndoRedoManager` as a single undo action.

## Implementation status

Branch: `feature/collision-manage`
Commits: `adfecee`, `8ccb146`, `f0f2ae6`, `0367208`

## Tests

- `test_project/tests/test_collision.gd` — GDScript handler tests against
  the live editor.
- `tests/unit/handlers/test_collision.py` — Python handler unit tests.
- `tests/integration/test_mcp_tools.py` — MCP tool dispatch integration test.
- `tests/unit/test_server_instructions.py` + `tool_catalog.gd` — surface parity.

Validation run:
- `uv run pytest -q` → **1704 passed**
- `uv run ruff check src/ tests/` → clean
- `test_run suite=collision` → **15 passed, 0 failed**

## Note on scope

This is a generic Godot engine capability, not a project-specific or
third-party integration. It sits alongside the recently merged GridMap/CSG
tools as a core authoring tool.

## Merge notes

If `feature/visual-shader-manage` lands first, this branch will need a
trivial rebase for the overlapping registration/catalog/instructions files.

michaltomczykowski:
## Summary

- add `resource_manage(op="physics_shape_generate")` for bulk collision-body generation from `MeshInstance3D` scene paths
- support `StaticBody3D` and `Area3D` generation, with static as the default
- fit box, sphere, capsule, and cylinder shapes from a body-local AABB
- validate the full request before mutation and keep the bulk operation in one undo/redo action

## Maintainer feedback addressed

- uses the existing `resource_manage` tool surface instead of introducing a single-operation `collision_manage` domain
- registers `physics_shape_generate` in the existing physics-shape subsystem
- avoids the world-AABB/double-rotation issue by creating an unscaled sibling body with the mesh position and rotation, transforming all eight mesh-local AABB corners into body-local space, and deriving the collision shape from those transformed corners
- preserves offset mesh origins, rotation, non-uniform scale, and negative scale
- intentionally limits body generation to static bodies and areas; dynamic body generation remains out of scope
- prevalidates shape type, body type, paths, node existence, and node type before opening the undo action, so invalid bulk requests create nothing

## Verification

- `ruff check src tests`: passed
- `pytest -v`: 1,805 passed, 19 skipped
- full Godot test suite: 2,285 passed, 0 failed, 21 skipped across 70/70 suites
- live tool smoke test: generated a visible collider through `resource_manage`, verified the resulting `CollisionShape3D` and undo availability, then removed the smoke-test node

Closes #868

## Summary by CodeRabbit

* **New Features**
  * Added physics shape generation for mesh nodes.
  * Supports box, capsule, cylinder, and sphere collision shapes.
  * Creates static or area physics bodies with fitted collision shapes.
  * Preserves alignment for rotated, scaled, offset, and top-level meshes.
  * Supports batches of up to 1,024 meshes with bounded processing and a single undo/redo action.
* **Bug Fixes**
  * Improved validation and error reporting for invalid inputs and batch sizes.
  * Added safeguards while the editor is importing or scenes are being edited.
* **Documentation**
  * Updated resource management operation listings to include physics shape generation.


dsarno (changes requested):
Thanks for the revision — the corner-transform pipeline, the `resource_manage` surface, and the prevalidate-before-mutate structure are what was asked for, and the sync path holds up. Requesting changes on the deferred path and a few contract gaps. Line numbers are this PR's `physics_shape_handler.gd` unless noted. Verified on `main` at c6ba797; the PR merges clean there.

**Blocking**

1. **Deferred path applies stale plan-time state (:186, :375).** `_finish_generate_deferred` captures mesh/parent/transform in the plan phase, then yields across chunks. `_create_generated_entry` reads `plan.mesh.name` with no `is_instance_valid` guard — a mesh deleted during the window aborts the coroutine, no response is ever sent (client waits out the 30 s timeout), and the detached bodies in `created_nodes` are never rolled back (Nodes aren't RefCounted, so they leak). A mesh moved or reparented in the window passes the checks and lands under the old parent at the old transform with `undoable: true`. Re-validate mesh validity, `mesh.get_parent() == entry.parent`, and the transform before apply.
2. **A scene-root `MeshInstance3D` passes validation (:160).** `/Main` resolves to the scene root (scene_path.gd:59-61), whose parent in the editor is the internal viewport, so `parent == null` never trips: the body is added among editor internals, `set_owner` fails, `body_path` comes back `""`, and the undo action references an editor-internal node. Reject `node == scene_root` explicitly.
3. **Parent-chain scale still reaches the body (:165).** Only the mesh's own scale is stripped; a body added under a parent with scale `(1,2,1)` inherits it in world space (non-uniform-scale warning, deformed sphere/capsule/cylinder). The only scaled-parent test sets `top_level = true`. Compute the body transform in global space as a top-level sibling (the corner pipeline already takes an arbitrary `mesh_to_body`), or document the constraint.
4. **Null/degenerate AABB commits a zero-size collider as success (:176).** A `MeshInstance3D` with `mesh == null` yields `BoxShape3D.size = (0,0,0)` / radius 0 and `{created, undoable: true}`. Reject empty bounds in the plan step, same as the wrong-type case.
5. **Duplicates and retries stack colliders (:121, :186).** No de-dup of `paths`, no existing-`<Name>Collider` sibling check, and `add_child(body, true)` renames clashes to `WallCollider2` silently. A repeated path or a retry after a lost reply doubles every collider with both calls reporting success.
6. **`scene_file` is rejected at the Python surface (physics_shape.py:11).** The GDScript `_validate_generate_request` reads `scene_file` and enforces `EDITED_SCENE_MISMATCH`, but the Python handler has no such parameter and `_meta_tool` passes `**params` through, so the kwarg fails with INVALID_PARAMS and the guard is unreachable — every node op threads `scene_file` (node.py `_mutation_params`). Thread it through, or drop the dead guard and document that the op can't be pinned.
7. **Undocumented 16-path cap through `batch_execute` (resource.py:62).** `dispatch_direct` strips `_request_id`, so batch calls take the sync fallback capped at 16 and abort the whole batch; the docstring promises 1024. Document the bound or lift it.
8. **`shape_type` vocabulary differs from `autofit` (:83).** `generate` rejects `"BoxShape3D"` that `autofit` (same file, :460-470) normalizes and whose tests use. One shared normalizer for both ops.

**Should fix**

9. **Timeout budget (:42).** At 1024 paths the deferred path performs ~254 `process_frame` awaits against a fixed 30 s wall clock; an unfocused editor (~10 FPS by default) burns ~25 s before any real work, and on expiry `_rollback_generated_deferred` discards everything. `PHYSICS_SHAPE_GENERATE_TIMEOUT_SEC = 35.0` is also a second bare literal, not derived from the plugin budget the way custom.py does. Chunk by a time budget, fold the trivial `is String` loop into validation, and derive the Python timeout.
10. **The production path has no GDScript coverage (test_physics_shape.gd:20).** The suite builds `PhysicsShapeHandler.new(_undo_redo)` with no connection, so all twelve `test_generate_*` cases hit the ≤16-path sync fallback; every real MCP call takes `_finish_generate_deferred`. AGENTS.md asks for direct coverage of editor/undo semantics in `test_project/tests/`; `test_editor.gd:434` has a `_StubConnection` that can drive it. Also `test_generate_rejects_empty_non_array_and_non_mesh_paths` (:902) asserts param errors before its no-scene skip while the handler checks `require_edited_scene` first, so it fails on EDITOR_NOT_READY instead of skipping.

**Non-blocking cleanup (all verified)**

- `_transform_aabb` (:415) is `Transform3D * AABB` — identical output including negative scale (checked headless); `editor_handler.gd:871` already uses the operator.
- `_send_deferred_if_pending` (:251) duplicates the guard `McpConnection.send_deferred_response` already applies and suppresses its "dropped late response" log; keep the mid-loop `_deferred_request_pending` early-aborts, send directly.
- The per-entry `is String` check exists twice (:121, :290) — once in `_validate_generate_request` is enough.
- Per AGENTS.md "include the failing parameter value": the five `paths` messages (:69, :98, :104, :125, :299) omit the received type/count/index, unlike `:157` in the same file.
- Pre-existing but now divergent: `autofit`'s `_measure_bounds` (:618) still multiplies `aabb.size` by a possibly negative `get_scale()`; `BoxShape3D` rejects negative sizes and the response reports the negative size with `undoable: true`. Routing it through the new corner pipeline fixes both paths at once.

dsarno:
Deferring to milestone 4.1: the head is still d16770c, the same commit reviewed by dsarno on September 2, so the requested fixes have not landed.

Remaining blockers: revalidate mesh lifetime, parent and transform after deferred yields and roll back failures; reject a scene-root mesh; handle inherited parent scale correctly; reject null/degenerate mesh bounds; prevent duplicate paths and retry-created colliders; expose scene_file through the Python handler; document the 16-path batch_execute limit; and share shape-type normalization with autofit. Also address the timeout/chunking budget and add production deferred-path success/rollback tests (including the no-scene test ordering). The review's cleanup remains: use Transform3D * AABB, remove duplicated deferred-send/type checks, include failing parameter values, and fix negative-scale autofit through the corner pipeline.

No contributor code was rewritten. — Codex

michaltomczykowski:
Hey, will take me few days to adjust.

dsarno:
Thanks @michaltomczykowski — this landed on `main` via #1028, which reworked your branch into a step-driven job (frame-budgeted generation, stale-plan detection, rollback) and addressed the review items. You are credited in the CHANGELOG entry for `resource_manage(op="physics_shape_generate")`, and it ships in the next release (4.0.4). Closing this PR as superseded by #1028.

PR #925, opened 2026-08-28
fix: bound send_command's write as well as its response wait

### LINKED ISSUE #910: [bug] send_command's deadline covers the response wait but not the write

**Location:** `src/godot_ai/transport/websocket.py:546-547`

**Risk:** The body is:

```python
await ws.send(request.model_dump_json())
return await asyncio.wait_for(future, timeout=timeout)
```

The deadline wraps only the response wait. Once the server's WebSocket transport is write-paused by TCP backpressure — an editor that has stopped reading its socket, with roughly a megabyte of outbound bytes queued — `await ws.send(...)` blocks for the whole stall and the per-command timeout never starts.

Nothing else bounds it. `asyncio.wait_for` appears in `src/` only at this response wait, the handshake recv, and two unrelated call sites; the websockets keepalive ping goes through the same paused drain; and the `finally` that fails in-flight futures on disconnect only runs after the connection actually drops. The AI client's tool call hangs with no error rather than timing out with something actionable.

**Reproduction:** With a fake websocket whose `send()` awaits an event that is never set, `send_command` with a 0.2 s budget was still running when a 2.0 s test backstop cancelled it.

**Fix:** Wrap both statements in a single deadline:

```python
async with asyncio.timeout(timeout):
    await ws.send(request.model_dump_json())
    return await future
```

The existing `except asyncio.TimeoutError` and the `finally` that pops `_pending` both continue to work unchanged.

---

*Suggested labels: bug, severity: medium, area: server*

azereki:
## Summary
- `send_command` now applies one deadline to both `ws.send` and the response wait, so a write-paused editor cannot hang a tool call forever.
- Existing TimeoutError handling and `_pending` cleanup are unchanged.
- Fixes #910.

## Test plan
- [x] `ruff check` on the touched files
- [x] `pytest -q tests/integration/test_websocket.py::TestPendingFutureCleanup` (3 passed, including write-stall timeout)

### DIFF
--- a/src/godot_ai/transport/websocket.py
+++ b/src/godot_ai/transport/websocket.py
@@ -543,9 +543,13 @@ class GodotWebSocketServer:
         ## raise / TimeoutError / cancellation it prevents Futures leaking
         ## into _pending forever.
         try:
-            await ws.send(request.model_dump_json())
-            return await asyncio.wait_for(future, timeout=timeout)
-        except asyncio.TimeoutError:
+            # One deadline for the write and the response wait. A paused
+            # WebSocket drain used to block forever in `ws.send` because
+            # only the recv side was wrapped (#910).
+            async with asyncio.timeout(timeout):
+                await ws.send(request.model_dump_json())
+                return await future
+        except TimeoutError:
             raise TimeoutError(
                 f"Command {command} timed out after {timeout}s on session {session_id}"
             )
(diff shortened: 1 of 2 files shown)


azereki:
Duplicate of #923. This PR also targeted issue #910 (send_command deadline misses the write). Superseded by #923, which is equally green on CI and is the more complete fix: it bounds the write and distinguishes send-leg vs response-leg timeouts so a write-stall is not retried as a silent unanswered request.

PR #927, opened 2026-08-28
fix: record descendant ownership inside reparent and duplicate undo

### LINKED ISSUE #904: [bug] reparent_node and duplicate_node fix descendant ownership outside the undo action
**Severity:** High · Verified present on `main` at v3.2.4 (`b4bc574`). Originally surfaced by an audit of v3.1.5 and re-confirmed against current `main` before filing this correction; line numbers above are v3.2.4.

**Location:** `plugin/addons/godot_ai/handlers/node_handler.gd:165` (reparent) and `:392` (duplicate) — both immediately after their `commit_action()`

**Risk:** Both call `_set_owner_recursive(node, scene_root)` **after** `commit_action()`, so the descendant-ownership fixup is not part of the recorded action and cannot be replayed.

**reparent — breaks on undo.** Undo replays `remove_child`, and Godot's `Node::remove_child` runs `_propagate_after_exit_tree()`, which clears `owner` on every node of the pruned subtree whose owner sits outside that subtree. The action's `add_undo_method(node, "set_owner", scene_root)` restores the moved node itself, but nothing restores its descendants.

**duplicate — breaks on redo.** Undo is safe here: it is just `parent.remove_child(dup)`, so the copy disappears cleanly. Redo replays `add_child(dup)` and `set_owner(dup, scene_root)` but never the recursive pass, so the restored duplicate's descendants come back with a null owner.

A node with a null owner is not serialized into the scene file on save, so in both cases the subtree is silently dropped the next time the scene is written.

**Reproduction:** Create a parent with a child, reparent the parent, undo, and read the child's `owner` — it is null. For duplicate: duplicate a node that has a child, undo, redo, and read the copy's child's `owner` — null. Two GDScript tests covering exactly these fail today; a third asserting duplicate's *undo* path passes, confirming the split.

**Fix:** Move the recursive owner fixup inside the `create_action`/`commit_action` block with matching do and undo methods, so both directions replay it.

---

*Suggested labels: bug, severity: high, area: handlers*

azereki:
## Summary
- Record descendant `set_owner` inside the `reparent_node` and `duplicate_node` undo actions instead of calling `_set_owner_recursive` after `commit_action`.
- Reparent undo now restores each descendant's prior owner (Godot `remove_child` clears owners on the pruned subtree). Duplicate redo now restores owners on the copy's children.
- Add GDScript regressions in `test_node.gd` for reparent undo and duplicate undo/redo owner restoration.

## Test plan
- [ ] `test_reparent_preserves_descendant_owner_on_undo` — parent+child, reparent parent, undo, child's owner remains the scene root
- [ ] `test_duplicate_restores_descendant_owner_on_redo` — duplicate a node with a child, undo, redo, copy child's owner is the scene root
- [ ] Existing `test_duplicate_node_basic` still covers undo removing the copy
- Live GDScript suite was not run because no editor is connected.

Fixes #904

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed undo and redo behavior for reparenting nodes, preserving descendant ownership correctly.
  * Fixed duplicated node hierarchies so descendant ownership is restored when the action is redone.

* **Tests**
  * Added regression coverage for owned and unowned descendants across reparent and duplicate undo/redo workflows.

### DIFF
--- a/plugin/addons/godot_ai/handlers/node_handler.gd
+++ b/plugin/addons/godot_ai/handlers/node_handler.gd
@@ -148,21 +148,27 @@ func reparent_node(params: Dictionary) -> Dictionary:
 
 	var old_parent := node.get_parent()
 	var old_idx := node.get_index()
+	## Snapshot descendants before commit: remove_child clears owner on any
+	## child whose owner sits outside the pruned subtree, so both do and undo
+	## must restore those owners as part of the recorded action (#904).
+	var descendants := _collect_descendants(node)
 
 	_undo_redo.create_action("MCP: Reparent %s" % node.name)
 	_undo_redo.add_do_method(old_parent, "remove_child", node)
 	_undo_redo.add_do_method(new_parent, "add_child", node, true)
 	_undo_redo.add_do_method(node, "set_owner", scene_root)
+	for child in descendants:
+		_undo_redo.add_do_method(child, "set_owner", scene_root)
 	_undo_redo.add_do_reference(node)
 	_undo_redo.add_undo_method(new_parent, "remove_child", node)
 	_undo_redo.add_undo_method(old_parent, "add_child", node, true)
 	_undo_redo.add_undo_method(old_parent, "move_child", node, old_idx)
 	_undo_redo.add_undo_method(node, "set_owner", scene_root)
+	for child in descendants:
+		var prior_owner: Node = child.owner if child.owner != null else scene_root
+		_undo_redo.add_undo_method(child, "set_owner", prior_owner)
 	_undo_redo.add_undo_reference(node)
 	_undo_redo.commit_action()
-
-	# Re-set owner for all descendants (reparent can break ownership chain)
-	_set_owner_recursive(node, scene_root)
 
 	return {
 		"data": {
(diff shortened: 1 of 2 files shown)


dsarno (changes requested):
Requesting one small correctness fix before merge.

In 
eparent_node, the descendant undo loop snapshots child.owner but substitutes scene_root when the original owner is null:

`gdscript
var prior_owner: Node = child.owner if child.owner != null else scene_root
`

That means undo cannot restore the exact pre-operation state for an intentionally unowned descendant; it becomes scene-owned instead. Please preserve the nullable owner unchanged and add a Godot-side do/undo/redo regression test containing both a scene-owned and an unowned descendant.

The overall direction is good, and the combined integration branch passed the live Godot suite; this is a narrow undo-integrity edge case.

azereki:
@dsarno Addressed the nullable-owner undo case.

`reparent_node` now snapshots `child.owner` unchanged (null stays null) instead of substituting `scene_root`. `test_reparent_restores_owned_and_unowned_descendants_on_undo_redo` covers a scene-owned descendant and an unowned descendant through do/undo/redo.

dsarno:
Thanks — this fix, including the nullable-owner undo case from the review round, has been ported onto the v4 branch (#943) with attribution in the source comment. This PR no longer applies cleanly there (both files conflict with the v4 rewrite), so I'll close it in favor of #943 when that merges rather than ask for a rebase. Credit stands.

dsarno:
Superseded: this fix (including the nullable-owner case from review) is on main via #943 (c6ba797) with attribution in the source comment. Closing rather than asking for a rebase — thank you.

PR #928, opened 2026-08-28
fix: treat backslash paths as sensitive writes

### LINKED ISSUE #905: [security] Backslash separators bypass the .godot/ and addons/godot_ai/ write guards
**Severity:** High · Verified present on `main` at v3.2.4 (`b4bc574`). Originally surfaced by an audit of v3.1.5 and re-confirmed against current `main` before filing this correction; line numbers above are v3.2.4.

**Location:** `plugin/addons/godot_ai/utils/path_validator.gd:155-172` (`_reject_sensitive_write`); the segment split is `:163`, the loaded-plugin clause `:171`

**Risk:** The two segment-based clauses of `_reject_sensitive_write` derive their segments from `path.trim_prefix("res://").split("/", false)`, which splits on forward slash only.

`res://.godot\uid_cache.bin` yields the single segment `.godot\uid_cache.bin`, which matches neither the `.godot` any-depth check nor anything else. `res://addons\godot_ai\plugin.gd` yields one segment, so the `segments.size() >= 2` loaded-plugin check never fires. Containment still passes because `String.simplify_path()` normalizes backslashes, and Godot's FileAccess normalizes them again on Windows — so the write lands where the blocklist meant to refuse it: the editor's metadata cache, or a `.gd` file in the running plugin's own tree (which the guard's comment notes can SIGABRT the editor mid-call).

**Not affected:** the `project.godot` and `override.cfg` clauses at `:157` and `:159`, which use `String.get_file()` — that is backslash-aware, so the startup-execution surface stays protected.

**Reproduction:** `path_error("res://.godot\\uid_cache.bin", "path", true)` and `path_error("res://addons\\godot_ai\\plugin.gd", "path", true)` both return null (accepted). The forward-slash forms of both are correctly refused.

**Fix:** Normalize separators (`path.replace("\\", "/")`) before the segment split in `_reject_sensitive_write`, and add Windows-path cases to the validator tests.

---

*Suggested labels: kind/security, severity: high, windows, area: handlers*

azereki:
## Summary
- Normalize backslashes to `/` in `_reject_sensitive_write` before the segment split so Windows-style paths cannot skip the `.godot/` and `addons/godot_ai/` write guards.
- Add GDScript coverage for backslash, mixed-separator, and `path_error(...)` forms of those writes.
- Leave the `project.godot` / `override.cfg` clauses unchanged; they already use `get_file()`.

## Test plan
- [ ] Run `test_run suite=path_validator` against a live Godot editor (skipped here: no editor connected; Godot was not launched)
- [ ] Confirm write of `res://.godot\uid_cache.bin` is rejected
- [ ] Confirm write of `res://addons\godot_ai\plugin.gd` is rejected
- [ ] Confirm mixed-separator write `res://addons\godot_ai/plugin.gd` is rejected
- [ ] Confirm `path_error(...)` returns `VALUE_OUT_OF_RANGE` for both backslash paths
- [ ] Confirm existing forward-slash forms still reject as before
- [ ] No Python files changed; ruff not run

Fixes #905

### DIFF
--- a/plugin/addons/godot_ai/utils/path_validator.gd
+++ b/plugin/addons/godot_ai/utils/path_validator.gd
@@ -158,9 +158,12 @@ static func _reject_sensitive_write(path: String) -> String:
 		return "Refusing to write res://project.godot (project manifest)"
 	if file_lower == "override.cfg":
 		return "Refusing to write res://override.cfg (startup config override)"
-	# Reject the `.godot/` editor-metadata dir at any depth. Split drops empty
-	# segments so a trailing slash can't hide a segment from the check.
-	var segments := path.trim_prefix("res://").split("/", false)
+	# Reject the `.godot/` editor-metadata dir at any depth. Normalize
+	# backslashes first — a Windows-style `res://.godot\uid_cache.bin`
+	# would otherwise be a single segment and skip both this check and the
+	# loaded-plugin clause below. Split drops empty segments so a trailing
+	# slash can't hide a segment from the check.
+	var segments := path.replace("\\", "/").trim_prefix("res://").split("/", false)
 	for segment in segments:
 		if segment.to_lower() == ".godot":
 			return "Refusing to write under res://.godot/ (editor metadata)"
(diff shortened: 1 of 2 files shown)


azereki:
Duplicate of #922. This PR also targeted issue #905 (backslash bypasses write guards). Superseded by #922, which is equally green on CI and is the more complete fix: it normalizes with `simplify_path()` so `.` / `//` segments cannot shift the anchored plugin-tree check the way a separator-only replace would leave open.

PR #930, opened 2026-08-28
fix: parse Zed settings.json comments as JSONC

### LINKED ISSUE #914: [bug] Zed's stock commented settings.json is rejected as malformed

**Location:** `plugin/addons/godot_ai/clients/_json_strategy.gd:469` (`_read_or_init`)

**Risk:** Zed ships `settings.json` with a `//` comment header on first launch. `_read_or_init` parses with Godot's `JSON.parse`, which has no comment support, so a default Zed install fails at read time.

The row goes permanently red and its Configure/Retry button can never succeed, while the message — "Refusing to overwrite …: JSON parse error … Fix or move the file" — implies the user's own file is corrupt when it is exactly what Zed wrote.

Scope note: VS Code and VS Code Insiders point at `mcp.json`, which is not seeded with comments, so in practice this is Zed-specific rather than affecting every JSONC client.

**Reproduction:** A config file beginning with `// Zed settings` followed by valid JSON returns `JSON parse error on line 0: Unexpected character`.

**Fix:** Strip `//` and `/* */` comments before parsing for clients whose format is documented as JSONC. The existing BOM-stripping code immediately above is the natural place for it, and the dock's "Run this manually" fallback already covers the case where a write is genuinely unsafe.

---

*Suggested labels: bug, severity: medium, area: telemetry*

azereki:
## Summary
- Zed ships `settings.json` with a `//` comment header. Godot's `JSON.parse` has no comment support, so a default install failed Configure permanently.
- `_read_file_text` now strips `//` line comments and `/* */` block comments from the parse copy only (after the existing BOM strip). Sequences inside JSON strings are left intact; `original_text` is unchanged so remove/rollback stay byte-for-byte.
- Added GDScript tests for the Zed header, block comments, in-string `//`, and invalid JSON still erroring.

## Test plan
- [ ] `test_json_strategy_tolerates_zed_jsonc_header` / `test_read_file_text_parses_zed_jsonc_header`
- [ ] `test_read_file_text_strips_jsonc_block_comments`
- [ ] `test_strip_jsonc_preserves_comment_markers_inside_strings`
- [ ] `test_read_file_text_jsonc_still_errors_on_invalid_json`
- [ ] Existing BOM / unparseable-file tests still pass
- Live GDScript tests were not run (Godot was not launched, per issue instructions).

Fixes #914

## Summary by CodeRabbit

- **New Features**
  - Configuration updates preserve existing formatting, comments, UTF-8 markers, and large numeric values.
  - JSONC files are handled more reliably, including commented and empty files.
  - Unrelated entries remain unchanged when servers are added or removed.

- **Bug Fixes**
  - Improved streamed-response handling for keepalive lines, carriage returns, and UTF-8 markers.
  - Extended timeout support for lengthy operations.

### DIFF
--- a/plugin/addons/godot_ai/clients/_json_strategy.gd
+++ b/plugin/addons/godot_ai/clients/_json_strategy.gd
@@ -446,8 +446,9 @@ static func _arrays_equal(left: Variant, right: Variant) -> bool:
 ## Callers must NOT treat the error path as an empty config — doing so blows
 ## away the user's other MCP entries on the next write. The `original_text`
 ## is the exact captured source so transactional rollback can restore
-## byte-for-byte; the UTF-8 BOM and JSONC comments are stripped only from
-## the parsing copy.
+## byte-for-byte; the UTF-8 BOM is restored on `original_text` when the
+## decoder skipped the on-disk EF BB BF, and JSONC comments are stripped
+## only from the parsing copy.
 static func _read_file_text(path: String) -> Dictionary:
 	if not FileAccess.file_exists(path):
 		return {"exists": false, "ok": true, "data": {}, "original_text": ""}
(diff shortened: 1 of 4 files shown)


dsarno (changes requested):
Requesting changes because the current JSONC acceptance creates a user-config data-loss path.

`_strip_jsonc()` removes all comments for parsing, then `configure()` rewrites the entire document with `JSON.stringify`. The tests verify that semantic keys survive but do not verify that the Zed header or user comments/formatting survive, so clicking Configure silently destroys them.

There is also a fail-open malformed-input case: an unterminated trailing block comment such as `{"theme":"x"} /* unfinished` is stripped to EOF, becomes valid JSON, and is rewritten rather than rejected.

Please either use a token-preserving update for the target server entry, or narrowly preserve the known stock header/comment text. At minimum, reject unterminated block comments and add byte-for-byte preservation tests for unrelated comments/content. This is a data-integrity concern rather than an exploit concern.

azereki:
@dsarno Took the token-preserving path.

Configure/remove now splice only the target server entry in the original text, so Zed headers, comments, and unrelated formatting survive instead of a `JSON.stringify` rewrite. Unterminated block comments fail closed. Byte-for-byte tests cover headers/comments, empty-object comments, BOM round-trip, and glued-comment rejection.

dsarno:
Thanks for doing the token-preserving splice — that addressed the data-loss concern properly.

A heads-up that changes the picture: the v4 branch (#943) made a different call for Zed. `clients/zed.gd` sets `automatic_config_edits = false` with the rationale that owning a second JSON tokenizer to preserve comments makes the shared strategy harder to trust; Configure/Remove now hand the user the exact manual entry instead of writing the file. So on v4 the permanent-red/dead-Configure failure in #914 is gone by construction, and this PR's writer path conflicts with that decision (and with `script/ci-reload-test`).

What v4 does *not* fix: the status read still goes through `_read_file_text` → `JSON.parse`, so a stock Zed file still shows a parse error in the row even though nothing will be written. If you're willing, the useful residual is a much smaller PR against `v4/architecture-simplification`: comment-tolerant *read* for status only (fail closed on unterminated block comments, as you already do), with no write path. I'd take that.

dsarno:
Update: #943 has merged to `main` (c6ba797), so the read-only status-parse follow-up suggested above would target `main` directly.

azereki:
Superseded by #954.

Agreed on the v4 call — with `automatic_config_edits = false` the dead-Configure failure is gone by construction, and the writer here conflicts with that. #954 is the residual you asked for: comment-tolerant read only, against `main`.

Status, `manual_target_details`, and `authoritative_tier_path` strip comments from a throwaway parse copy; `configure`/`remove` keep the strict parser. `_status_allows_comments` requires both `config_allows_comments` and `not automatic_config_edits`, so re-enabling automatic edits on a JSONC descriptor restores the parse error rather than quietly permitting a comment-dropping rewrite. Unterminated block comments still fail closed.

The splice, the removal comment handling, and the `script/ci-reload-test` changes are all dropped. One thing carried over as a note rather than a fix: the BOM finding from this branch (Godot's text decoder eats a leading BOM, so `original_text` loses it and Remove would strip it from disk) looks like a live bug on `main`, but it is write-path, so #954 leaves it alone.

PR #931, opened 2026-08-28
fix: honor a later telemetry opt-out on an already-running server

### LINKED ISSUE #913: [robustness] Telemetry opt-out cannot reach a server the plugin did not spawn

**Location:** `src/godot_ai/telemetry.py:177`, consumers at `:330`, `:396`, `:422`

**Risk:** `TelemetryConfig.enabled` is written exactly once, in `__init__`, and every consumer reads `self.config.enabled`. There is no setter and no per-record environment re-check.

The plugin's only opt-out channel to the server is `GODOT_AI_DISABLE_TELEMETRY` injected into the environment of a spawn. So a server the plugin adopted rather than spawned — a hand-started dev server, a CI-started one, or an attach-owned backend — keeps whatever telemetry state it was born with for its whole life. The dock's "Apply and Restart Server" does not help: it toggles the plugin, not that process.

The user unchecks the box, the dock shows telemetry off, and the running server keeps sending. That is a trust problem as much as a correctness one.

Spawn-time opt-out itself works correctly — this is only about a process that is already running. Sibling of #530, which covered the falsey-env-var case at injection time; this is a different mechanism.

**Reproduction:** Build a collector with telemetry enabled, POST once, set `GODOT_AI_DISABLE_TELEMETRY=true` in the environment, and record again — a second POST goes out. The same holds through the `plugin_event` route.

**Fix:** Re-read the setting per send, or accept a runtime toggle over the existing WebSocket channel. Either way the dock should report the **live server's** telemetry state rather than the local EditorSetting, so the two cannot silently disagree.

---

*Suggested labels: robustness, severity: medium, telemetry, area: telemetry*

azereki:
## Summary
- Telemetry record/send paths re-read `GODOT_AI_DISABLE_TELEMETRY` / `DISABLE_TELEMETRY` so an adopted process stops sending after a later opt-out.
- `/godot-ai/status` now publishes `telemetry_enabled`, and the dock tooltip reports that live server state when it disagrees with the EditorSetting.
- Fixes #913.

## Test plan
- [x] `ruff check` on the touched Python files
- [x] `pytest -q tests/unit/test_telemetry.py tests/unit/test_server_status.py` (61 passed)
- [ ] Live dock tooltip against an adopted server (no editor connected in this pass)

## Summary by CodeRabbit

- **New Features**
  - Telemetry status is now shown in the editor based on the running server’s current state.
  - The editor indicates when its local telemetry setting differs from the server’s setting.
  - Telemetry opt-out environment changes now take effect without restarting the running process.

- **Bug Fixes**
  - Clarified that editor telemetry settings only update servers launched by the plugin.
  - Older server versions continue to report status without unsupported telemetry details.

- **Documentation**
  - Added guidance explaining telemetry behavior, status reporting, and runtime environment changes.

### DIFF
--- a/docs/TELEMETRY.md
+++ b/docs/TELEMETRY.md
@@ -88,6 +88,19 @@ spawns). Attach entries written before the toggle changed read as
 `configured_mismatch` — re-run Configure so the client relaunches the bridge
 with the current preference.
 
+Apply & Restart does **not** mutate the environment of a server the plugin
+adopted rather than spawned (a hand-started dev server, CI backend, or
+attach-owned process). That process keeps whatever telemetry state it was
+born with until *its* environment changes or it is replaced. `/godot-ai/status`
+publishes `telemetry_enabled` from a live env re-read so the dock can show
+when the checkbox and the running server disagree; it is a report, not an
+opt-out channel.
+
+Record/send paths in an already-running Python process re-read
+`GODOT_AI_DISABLE_TELEMETRY` / `DISABLE_TELEMETRY` on each event, so an
+in-process env change in that server takes effect without reconstructing
+the collector.
+
 ### Via environment variable
 
 Set either environment variable to `true` / `1` / `yes` / `on`:
--- a/plugin/addons/godot_ai/mcp_dock.gd
+++ b/plugin/addons/godot_ai/mcp_dock.gd
@@ -1559,8 +1559,8 @@ func _on_telemetry_toggled(pressed: bool) -> void:
 
 
 ## Report the running server's telemetry state, not just this editor's
-## checkbox. An adopted backend ignores the EditorSetting until its own
-## environment is changed (#913).
+## checkbox. Apply & Restart injects opt-out into a server this plugin
+## spawns; it cannot change the environment of an adopted process (#913).
 func _live_telemetry_tooltip(local_enabled: bool) -> String:
 	if _plugin == null or not _plugin.has_method("_probe_live_server_status"):
 		return ""
--- a/plugin/addons/godot_ai/mcp_dock.gd
+++ b/plugin/addons/godot_ai/mcp_dock.gd
@@ -1548,12 +1548,33 @@ func _load_telemetry_setting() -> void:
 		)
 	else:
 		_telemetry_toggle.disabled = false
-		_telemetry_toggle.tooltip_text = ""
+		_telemetry_toggle.tooltip_text = _live_telemetry_tooltip(enabled)
 
 
 func _on_telemetry_toggled(pressed: bool) -> void:
 	_telemetry_pending_enabled = pressed
+	if _telemetry_toggle != null:
+		_telemetry_toggle.tooltip_text = _live_telemetry_tooltip(pressed)
 	_refresh_tools_ui_state()
+
+
+## Report the running server's telemetry state, not just this editor's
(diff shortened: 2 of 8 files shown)


dsarno (changes requested):
Requesting either an implementation change or a narrower stated contract.

`TelemetryConfig.live_enabled()` re-reads `os.environ` inside the already-running Python process. That passes tests which mutate `os.environ`, but the Godot editor cannot mutate the environment of an adopted server process. Applying the dock checkbox reloads the plugin; it does not deliver the opt-out to that backend. The new tooltip correctly discloses this mismatch, but the issue/PR currently claim the later opt-out is honored.

Please either:

1. narrow the PR to “report the adopted server's live telemetry state and honor in-process env changes,” updating the issue/title/tests accordingly; or
2. add an authenticated, instance-bound runtime opt-out channel.

Please do not put a mutation on the unauthenticated loopback status/lease surface. This is a privacy/trust-contract concern, not evidence of data exfiltration in the patch.

azereki:
@dsarno Took option 1 (narrower contract), no mutation on the unauthenticated loopback surface.

`live_enabled()` still re-reads in-process env; `/godot-ai/status` publishes `telemetry_enabled`; the dock tooltip reports that live state and does not claim the checkbox opts out an adopted process. Docs and tests match that scope. Intended title: `fix: report live telemetry state and honor in-process env opt-out`. pytest 61 passed; live status shows `telemetry_enabled: false` under the opt-out env.

dsarno:
The narrowed contract from your 2026-08-29 comment is the right scope. Two things before I re-review:

- This now conflicts in all 8 files with the v4 branch (#943), which is where it will land — `telemetry.py:177` is still the set-once `self.enabled` there, so the fix is still needed. Please rebase onto `v4/architecture-simplification`.
- Please update the PR title/description to the narrowed claim ("report live telemetry state and honor in-process env changes") so the merge commit doesn't say the later opt-out is honored for an adopted process.

dsarno:
Update: #943 has merged, so the rebase target is now plain `main` (merge c6ba797), not the v4 branch. Everything else in the previous comment stands.

azereki:
Superseded by #955.

Two things happened to this PR:

1. **Its implementation is already on `main`.** The v4 architecture simplification (#943) ported the whole patch — `live_enabled()`, the record/send gates, `telemetry_enabled` on the status route, the projection passthrough, the dock tooltip, and both Python test files including the suggested fix to `test_status_route_telemetry_enabled_rereads_env`. That is why this shows as conflicting: it is based on v3.2.4, and those files were rewritten wholesale. A rebase would leave nothing.

2. **The blocking review point was never addressed.** The request was either a narrowed contract or "an authenticated, instance-bound runtime opt-out channel". The docs were narrowed, but the actual gap stayed open — unchecking the box still could not reach an adopted server.

#955 takes the second option. It adds a `telemetry_opt_out` event on the mutually authenticated per-backend WebSocket, dispatched only past the handshake. Nothing is added to the unauthenticated status/lease surface, which stays read-only. The latch is opt-out-only and one-way, so on a shared backend the most restrictive editor wins and no one can re-enable telemetry over the wire.

It also carries the two GDScript projection tests from this PR that the port dropped, and includes a live end-to-end run against an externally started server showing `telemetry_enabled` flip from `true` to `false` — the exact #913 scenario.

PR #932, opened 2026-08-28
fix: let Configure all run while the server is INCOMPATIBLE

### LINKED ISSUE #916: [ux] The INCOMPATIBLE panel prescribes "Configure all", which the dock refuses in that state

**Location:** `plugin/addons/godot_ai/utils/mcp_server_state.gd:113-114`; `mcp_dock.gd:2460-2461` (`_on_configure_all_clients` and its gate), `:2126`, `:2294` (the per-client gates), `:3551` (`_server_blocked_client_message`); message built at `utils/server_lifecycle.gd:627-633`

**Risk:** `McpServerState.blocks_client_health(state)` is literally `state == INCOMPATIBLE`.

The crash panel's body in that state — the usual text after a plugin update, when the port is held by an older backend an AI client's attach bridge keeps alive — reads: *"If AI-client attach bridges are keeping it alive, run Configure all to repin them, then restart those client apps."*

Clicking Configure all hits the `_server_blocks_client_health()` early return, which paints **every** client row `ERROR` with `_server_blocked_client_message()` — and that returns the same paragraph. The user reads "run Configure all", runs it, and gets 21 red rows each repeating "run Configure all".

The working sequence is: Restart Server first, then Configure all, then restart the client apps. Nothing states that ordering, and the message names the blocked step first.

To be clear about what is *not* broken: Restart Server is genuinely offered here. `_evaluate_recovery_port_occupant_proof`'s `status_name` tier grants recovery proof to any live occupant answering `/godot-ai/status` as `godot-ai`, so `can_recover_incompatible` is true and the button renders. Confirmed by reproducing the state against a live editor, which logged `MCP | proof: status_name`.

**Fix:** Preferred: let the write path run while INCOMPATIBLE. `ClientConfigurator.configure` already takes an explicit url and launch context, and `capture_launch_context()` reads the plugin version live, so it writes the correct new pin even while the old server still holds the port — the health gate should suppress status *interpretation*, not the *write*.

Minimum: reorder the message so Restart Server is named first, and render the Configure buttons disabled with a tooltip naming the prerequisite rather than letting the click fail into 21 red rows.

---

*Suggested labels: bug, severity: medium, area: self-update*

azereki:
## Summary
- Configure / Configure all now write client pins even when the occupant is INCOMPATIBLE, instead of painting every row ERROR.
- Health refresh skips interpretation in that state rather than turning 21 rows red with the same paragraph.
- The crash-panel repair text names Restart Server before Configure all.
- Fixes #916.

## Test plan
- [x] Updated GDScript assertions for message order and `blocks_client_health` docs
- [ ] Live INCOMPATIBLE dock: Configure all should write pins without 21 red rows (no editor connected in this pass)

## Summary by CodeRabbit

* **Bug Fixes**
  * Client rows are no longer incorrectly marked as errors when the server’s health status prevents client-health checks.
  * Client configuration remains available even when the server is in an incompatible state.
  * Updated repair guidance to recommend restarting the server before selecting “Configure all.”

* **Tests**
  * Added coverage to verify the revised repair-action order and server-state behavior.

### DIFF
--- a/plugin/addons/godot_ai/mcp_dock.gd
+++ b/plugin/addons/godot_ai/mcp_dock.gd
@@ -2453,7 +2453,14 @@ func _on_refresh_clients_pressed() -> void:
 
 
 func _on_configure_all_clients() -> void:
-	if ClientRefreshStateScript.should_disable_client_actions(_refresh_state):
+	## Per-row Configure already bypasses the RUNNING gate. INCOMPATIBLE
+	## skips health interpretation (#916), so Configure all must do the same
+	## even if a status sweep is still in flight — `_set_incompatible_server`
+	## does not reset `_refresh_state`.
+	if (
+		ClientRefreshStateScript.should_disable_client_actions(_refresh_state)
+		and not _server_blocks_client_health()
+	):
 		return
 	for client_id in _client_rows:
 		var status: Client.Status = _client_rows[client_id].get("status", Client.Status.NOT_CONFIGURED)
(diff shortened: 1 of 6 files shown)


azereki:
CodeRabbit outside-diff note on Configure all during RUNNING+INCOMPATIBLE was valid. Per-row Configure already bypasses the RUNNING gate; Configure all and the button now do the same when the server blocks client-health checks. Added test_configure_all_dispatches_while_incompatible_refresh_running. Local commit bf3516b3 on fix/916-incompatible-configure. The inline comment-update thread was already resolved in 13795d4.

PR #936, opened 2026-08-29
Fix version-check lifecycle-manager refcount cycle

uno1982:
McpServerLifecycleManager.arm_version_check() builds `_version_check = McpServerVersionCheck.new(self)`, and McpServerVersionCheck._init() stores `_manager = <that same manager>`. That is a RefCounted <-> RefCounted cycle, which GDScript's reference counting cannot collect. Once arm_version_check() has run even once, both instances and both scripts stay alive for the rest of the session. Surfaces at editor close as:

    WARNING: N ObjectDB instances were leaked at exit
    ERROR: N resources still in use at exit
    Resource still in use: res://addons/godot_ai/utils/server_lifecycle.gd
    Resource still in use: res://addons/godot_ai/utils/server_version_check.gd
    (+ the two scripts' method/property StringNames as orphans)

disarm_version_check()'s docstring already says it "releases the connection / manager references", but disarm() only nulled _connection.

- disarm_version_check(): also drop `_version_check`. arm_version_check() already reconstructs it lazily (`if _version_check == null`), and every get_version_check() / is_awaiting_server_version() reader is null-safe, so the observable contract is unchanged (a disarmed check and a nulled one both report not-awaiting).
- teardown_for_editor_exit(): call disarm_version_check() up front. The existing call sites only cover terminal-diagnosis paths, so a plain editor close never broke the cycle.
- plugin.gd::_exit_tree(): null `_lifecycle` after teardown, matching the nulling every sibling field already gets (the manager holds the plugin back via ServerLifecycleManager.new(self)).

Verified: `godot --headless --path test_project --import` parses clean; editor open/close (`--editor --quit`) reports zero leaked ObjectDB / resources / RIDs from the plugin, vs the counts above before.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed resource leaks when closing the editor.
  * Improved cleanup of server lifecycle and version-check resources during shutdown.
  * Prevented “resource still in use” warnings and leaked object messages from appearing at editor exit.
  * Ensured version checks can be initialized again when needed.

### DIFF
--- a/plugin/addons/godot_ai/plugin.gd
+++ b/plugin/addons/godot_ai/plugin.gd
@@ -622,6 +622,10 @@ func _exit_tree() -> void:
 	if _headless_disabled:
 		_server_started_this_session = false
 		_headless_disabled = false
+		## `_lifecycle` is built in _init(), before the headless guard in
+		## _enter_tree() runs, so it exists even on this path — null it here
+		## too (the full teardown below is skipped).
+		_lifecycle = null
 		return
 
 	if _custom_tool_registry != null:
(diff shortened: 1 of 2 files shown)


uno1982:
@coderabbitai full review

dsarno:
Thanks for the precise diagnosis. On the v4 branch (#943) the `McpServerVersionCheck.new(self)` instance no longer exists — `server_lifecycle.gd` now calls a static `ServerVersionCheck.evaluate(actual, expected)`, so there's no RefCounted↔RefCounted pair to leak, and `plugin.gd` nulls `_lifecycle` on `_exit_tree`. This PR conflicts with that rewrite in both files. I'll close it as superseded once #943 merges. If you still see `ObjectDB instances were leaked` from these two scripts on v4, please reopen with the counts.

dsarno:
Superseded: on main via #943 (c6ba797) the McpServerVersionCheck instance no longer exists — server_lifecycle.gd uses a static ServerVersionCheck.evaluate(), so the RefCounted cycle is gone and plugin.gd nulls _lifecycle on _exit_tree. Closing; please reopen with counts if the leak still shows on main.

PR #966, opened 2026-09-03
perf(windows): walk process ancestry in one PowerShell spawn per proof

dsarno:
## Why

The Windows Godot CI job went from 2.6 min (v3, and every main run June through August) to 6.4 min on main and 11.5 min on the #949 branch. All of it is the real-editor updater regression step, and inside that step the same tests run 3-5x slower on Windows than on Linux:

| Test | Linux | Windows |
|---|---:|---:|
| signed update restarts live server (private-https) | 19 s | 108 s |
| final-v3 capsule replaces tree, repins, starts | 14 s | 45 s |
| clean installer holds first start (cold / offline / wedged) | 10 s each | 36 s each |
| whole step | 180 s | 500 s |

Every Windows process proof paid a ~1.2 s `powershell.exe` spawn per field per hop: `process_descends_from` walked up to 16 parents one spawn each, and `pid_cmdline_is_godot_ai` walked up to 5 with a command-line spawn and a parent spawn per hop. One server adoption (brand check, ancestry check, fingerprint, then the re-verify) cost 10-15 spawns.

## What

`windows_process_chain` runs one PowerShell script that follows `ParentProcessId` inside a single process (each `Get-CimInstance` there costs tens of ms) and returns `pid|parent|commandline` per hop, command line last so an embedded pipe cannot split a record. A pure parser stops at the first malformed line, at a hop whose pid is not the parent the previous hop promised, and at `max_depth`, so a partial dump can only shorten a chain, never invent ancestry. `chain_descends_from` and `chain_has_godot_ai_brand` mirror the per-hop walks exactly, including the last resolved parent id counting as an ancestor.

Unchanged: the single-field helpers, `pid_alive`, `process_fingerprint`, and the identity string the fingerprint hashes. Non-Windows paths are untouched.

## Verification

- Full GDScript suite on Godot 4.7.beta3 (macOS): 2160 passed, 0 failed, 25 skipped. `port_resolver` suite: 21 passed, the Windows-only live walk skipped.
- Five new tests: parser (pipe in command line, garbage and wrong-hop truncation, depth cap), `chain_descends_from` (self, ancestor, last parent, absent, sentinel), `chain_has_godot_ai_brand` (depth), and a Windows-only live walk of the editor's own ancestry cross-checked against `process_descends_from`.
- The Windows path itself is exercised by this PR's CI: the Godot / Windows job and its updater regression step. Baseline on main for that step is 228 s; the number to compare is in this PR's run.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Windows process ancestry detection, including reliable fallback handling when system process information is unavailable or incomplete.
  - Added bounded ancestry checks to prevent incorrect matches from unrelated or overly deep process chains.
  - Improved identification of related Godot AI processes across supported platforms.
  - Improved reliability of Windows process inspection when PowerShell command execution is constrained.

- **Tests**
  - Expanded coverage for Windows process-chain parsing, ancestry validation, malformed data, depth limits, and live editor-process verification.
  - Improved diagnostic output for upgrade-path test failures.

## Status

- First run: the Windows updater regression step timed out at 12 minutes (728 s vs the 228 s baseline), three per-test 240 s timeouts, no diagnostics because that step ran pytest quietly. Cause: the first version walked ancestry to the root and routed the brand check through a fixed five-hop walk, both slower than the per-hop code they replaced.
- Second push bounds the walk at the ancestor, restores the one-spawn brand check, falls back to the per-hop walk on an empty chain, and makes the step verbose. The number to compare is the Windows regression step in the latest run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -200,10 +200,15 @@ jobs:
         shell: bash
         timeout-minutes: 12
         run: |
+          # -v -s --durations=0: quiet pytest prints nothing until the step
+          # ends, so a step killed by its timeout leaves no trace of which
+          # test hung. Verbose output with the harness's phase lines makes a
+          # slow or hung Windows editor attributable from the log alone.
+          pytest_args=(-v -s --durations=0 tests/integration/test_self_update_upgrade_paths.py)
           if [ "${{ runner.os }}" = "Linux" ]; then
-            xvfb-run -a env GODOT_BIN=godot pytest -q tests/integration/test_self_update_upgrade_paths.py
+            xvfb-run -a env GODOT_BIN=godot pytest "${pytest_args[@]}"
           else
-            GODOT_BIN=godot pytest -q tests/integration/test_self_update_upgrade_paths.py
+            GODOT_BIN=godot pytest "${pytest_args[@]}"
           fi
 
       - name: Start MCP server
--- a/plugin/addons/godot_ai/utils/port_resolver.gd
+++ b/plugin/addons/godot_ai/utils/port_resolver.gd
@@ -428,13 +428,18 @@ static func windows_process_chain(
 ) -> Array[Dictionary]:
 	if pid <= 1 or max_depth <= 0:
 		return []
+	## No double quotes anywhere in this script: OS.execute on Windows wraps
+	## an argument in quotes without escaping quotes already inside it, so an
+	## embedded `"` ends the argument early and PowerShell sees a mangled
+	## command. Every working script in this file is single-quote only for
+	## the same reason; string joins replace interpolation.
 	var script := (
 		"$current = %d; $stop = %d; $depth = 0; "
 		+ "while ($current -gt 1 -and $current -ne $stop -and $depth -lt %d) { "
-		+ "$p = Get-CimInstance Win32_Process -Filter \"ProcessId = $current\" "
+		+ "$p = Get-CimInstance Win32_Process -Filter ('ProcessId = ' + $current) "
 		+ "-ErrorAction SilentlyContinue; "
 		+ "if (-not $p) { break }; "
-		+ "\"$($p.ProcessId)|$($p.ParentProcessId)|$($p.CommandLine)\"; "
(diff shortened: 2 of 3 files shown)


dsarno:
Closing: measured, and it does not pay.

**Result.** With the ancestry walk working (third push, Windows Godot job green including the Windows-only live test), the updater regression step took 279 s against main's 228 s baseline. Per test, the batched walk is within runner noise of the per-hop code it replaced and slightly worse on the capsule migration:

| Windows test | per-hop (main / #949 runs) | batched walk (this PR, run 3) |
|---|---:|---:|
| signed update restarts live server | 92-108 s | 92 s |
| final-v3 capsule replaces tree, repins, starts | 45 s | 54 s |
| clean installer holds first start, cold / offline / wedged | ~36 s each | 35 / 37 / 45 s |
| whole step | 228 s | 279 s |

The premise was wrong: the per-hop code already stops at the first matching hop, typically one to three spawns, and finds the brand on the server process itself in one spawn. A single PowerShell process with several `Get-CimInstance` queries inside it is not cheaper than that on a CI runner, so there is nothing to batch.

**Two findings worth keeping from the three runs.**

1. `OS.execute` on Windows wraps an argument in quotes but does not escape quotes already inside it, so any PowerShell script passed through it must be single-quote only. The first two pushes had double-quoted strings in the script; PowerShell saw a mangled command, exited non-zero, and the walk came back empty (run 1: proofs failed closed, three 240 s timeouts; run 2: a wasted spawn per check before the per-hop fallback). Every working script in `port_resolver.gd` already follows the single-quote convention, without saying why.
2. The existing Windows `process_fingerprint` script has the same embedded-quote shape (`ForEach-Object { "$($_.CreationDate)|$($_.CommandLine)" }`). If it fails the same way, that would explain why its "restricted host" fallback to `Get-Process ... StartTime` exists at all, and it would mean every Windows fingerprint costs two spawns instead of one. Unverified here; worth one `Measure-Command` on a Windows box before touching it, since fingerprints are compared against stored values.

The branch stays for reference. The `-v -s --durations=0` change to the regression step is already on the #949 branch and lands with it.

PR #967, opened 2026-09-04
Surface cross-worktree claims at session start

drexthealpha:
`.claude/CLAUDE.md` already tells a session to *"verify the active worktree and connected Godot `test_project/` before editing or testing plugin code."* That check is per-session and asks the agent to look. It has no way to know a session in a sibling `.claude/worktrees/<name>` is mid-change on the same handler.

**What's in the PR**

- `.knos/decisions.md` — the standing rules from `AGENTS.md` and `.claude/CLAUDE.md`, in a file a fresh clone reads without installing anything. Every line cites the file it came from. All worktrees of one repo resolve to the same record, because it keys on `git rev-parse --git-common-dir` rather than `--show-toplevel`.
- One line in `.claude/settings.json` — a second `SessionStart` hook, after the existing `session-start.sh`, printing who is working on what. `|| true` so a machine without knos installed loses nothing and the session still starts.

**The worktree point is the reason this is worth anything here.** `--show-toplevel` returns the worktree root, so anything keyed on it treats every `.claude/worktrees/<name>` as a separate project. `--git-common-dir` is shared across all of them. That is the difference between a per-worktree record and a per-repo one.

**Disclosure:** I wrote Knos. The record is seeded from your own docs rather than mine so it is checkable rather than promotional. Close it if you'd rather not carry a third-party hook — the `.knos/decisions.md` half works with no hook at all, since it is plain markdown in the tree.

## Summary by CodeRabbit

* **Documentation**
  * Added a project decisions record covering development guidance, security expectations, protocol behavior, middleware ordering, skill naming, and worktree verification.
  * Clarified that recorded decisions exclude secrets and private paths, can be read without additional setup, and that live claim services are optional.
  * Documented that active work items are empty and recorded claims expire after 30 minutes.

* **Chores**
  * Added a session-start status check to display project state when a session begins.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -6,6 +6,10 @@
           {
             "type": "command",
             "command": "$CLAUDE_PROJECT_DIR/.claude/hooks/session-start.sh"
+          },
+          {
+            "type": "command",
+            "command": "knos status 2>/dev/null || true"
           }
         ]
       }
--- a/.knos/decisions.md
+++ b/.knos/decisions.md
@@ -1,0 +1,25 @@
+# Decisions and current work
+
+<!-- Written by `knos export`. Commit this file. -->
+
+A second clone reads this on its first question — it is one of the decision
+records knos looks for. Nothing here is private: secrets and private paths
+never reach it.
+
+
+## Decisions
+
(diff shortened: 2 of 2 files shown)


drexthealpha:
Closing this as not a fit for this repo's current contribution patterns. Thanks for the review time.

PR #986, opened 2026-09-07
fix: avoid writes before acquiring Windows startup lock

dsarno:
Concurrent attach clients can fail before acquiring the Windows startup lock: two first openers can observe an empty file, and one can lock byte zero before the other flushes its initialization byte. Windows rejects that write, so the second client fails instead of waiting for the existing lock. Remove the unnecessary initialization write; both native lock backends support locking an empty file.

This addresses the concrete startup race found while investigating the Windows Python 3.14 failure on #983. That earlier CI log did not retain the gathered exceptions, so the original three-client assertion now includes them.

The cross-process regression uses an independent native holder to lock an empty file. On Windows it first proves that the old initialization write is denied, then verifies that the corrected contender reports a bounded contention timeout and successfully acquires after release. Existing cancellation, error, and platform-lock tests remain covered.

Validation: complete CI Ruff scope; 2,531 Python tests passed with 23 platform skips, including real-editor integrations; full Godot suite 2,235 passed, zero failures, 24 skips across 72 suites; live concurrent-client smoke adopted the same editor backend twice and rejected a mismatched client version. Windows-native validation runs in this PR's CI.

References: [Python msvcrt locking](https://docs.python.org/3/library/msvcrt.html#msvcrt.locking), [Windows byte-range locking](https://learn.microsoft.com/en-us/windows/win32/fileio/locking-and-unlocking-byte-ranges-in-files).

Implemented and validated by Codex.

### DIFF
--- a/src/godot_ai/attach/ensure.py
+++ b/src/godot_ai/attach/ensure.py
@@ -221,14 +221,9 @@ class AdvisoryFileLock:
             handle = self.path.open("a+b")
         except OSError as exc:
             raise self._lock_error("open", exc) from exc
-        try:
-            handle.seek(0, os.SEEK_END)
-            if handle.tell() == 0:
-                handle.write(b"\0")
-                handle.flush()
-        except OSError as exc:
-            handle.close()
-            raise self._lock_error("prepare", exc) from exc
+        # Both backends can lock an empty file. Do not initialize its first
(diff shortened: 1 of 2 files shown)


dsarno:
@coderabbitai review

dsarno:
@coderabbitai review

PR #1043, opened 2026-09-10
fix(update): activate verified plugins without restarting the editor

dsarno:
Updating Godot AI activates the verified replacement inside the running editor and preserves open scenes, unsaved changes, selection, and undo history. A source-free runner serves both the v4 updater and the v3 migration capsule, waits for filesystem scans, retains scripts needed by existing undo callbacks, and loads a fresh canonical script graph. Signature verification, exact-tree checks, locking, and authenticated connections remain in place.

The update crash is now reproduced and fixed. Godot can leave its shared native ProgressDialog under the Update confirmation after an import. Plugin teardown destroyed that borrowed child; the next class-registration scan crashed. Teardown now returns the same native dialog to the editor root before removing the dock. Failure-dialog dismissal applies the same ownership rule. A real-editor regression verifies destruction of the old confirmation, survival of the native dialog, and registration of two new script classes.

Published 3.2.5 uses the new capsule without an editor restart. Published 4.0.4 still restarts once through its shipped updater; subsequent updates use the in-editor path. Compatibility files retain published script paths. Unsafe user-resource references refuse activation with a reason, and ambiguous failures retain repair evidence.

The follow-up also distinguishes unavailable Windows process evidence from an exited process, exposes port selection for unrecoverable incompatible servers, reuses the Windows reservation query while retaining bind checks, and skips linked child directories during backup scans. Test fixtures now enforce complete status-response framing, start the attached-client timeout at the actual gate, and restore endpoint settings.

Validation of the final source:

- Full Windows Python suite: **2,644 passed, 58 skipped**; one dependency deprecation warning. Ruff passed its full CI scope.
- Full rendered Godot suite after reload stress: **2,330 passed, 0 failed, 33 skipped**, across 74 suites. All three native-dialog regressions passed.
- Reload stress: **10/10 isolated reloads**; a fresh rendered concurrent run made **1,185 calls with 0 unexpected errors and 2/2 reloads**, with 14 tolerated reload transients.
- The natural import-under-confirmation reproduction crashed before the fix. With the fix, the same native object moved to the editor root before confirmation destruction, and the update completed in the same PID. Removing only the rescue call from a copied regression fixture failed with `plugin disable destroyed native shared progress`.
- Local signed fixtures passed 3.2.5 to the fixed version, then another update in the same editor; 4.0.4 passed with its expected first restart. Authenticated reads and writes passed after activation. Disposable editors exited normally, and generated caches were removed.

Earlier native checks covered attached-client and undo continuity. This latest run does not requalify every earlier release contract. The earlier editor-exit and undo findings remain tracked in #1044 and #1045; the captured progress-dialog crash establishes this mechanism without claiming every historical exit shares it. Configure all remains tracked separately in #1047.

Local fixtures use test signing keys and version substitutions. Cross-platform qualification of the exact signed 4.1.0 artifacts remains required before publication. [Fresh CI passed all 22 jobs](https://github.com/hi-godot/godot-ai/actions/runs/34625416480) on the final progress-dialog fix. All six existing review threads are resolved; CodeRabbit skipped a fresh review because reviews are disabled for this stacked base branch.

The remaining poll-loop review suggestion was checked against Godot's native HTTP client: normal body completion occurs in `read_response_body_chunk()`, after the guard, while `poll()` retains BODY or reports an error. The guard remains, with passing length/chunked/interruption regressions. See [the pinned engine source](https://github.com/godotengine/godot/blob/4.7.2-stable/core/io/http_client_tcp.cpp#L403-L418).

Stacked on #1042, which is stacked on #1039. Merge in that order.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,8 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
+          # Unit fixture preparation archives published v3 tags, including v3.2.5.
+          fetch-depth: 0
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,6 @@ on:
       - "docs/**"
       - "**.md"
   pull_request:
-    branches: [main, codex/v4-release-promotion]
     paths-ignore:
       - "docs/**"
       - "**.md"
@@ -284,6 +283,8 @@ jobs:
             tests/integration/test_godot_editor_restart.py \
             tests/integration/test_test_run_cache_warning.py \
             tests/integration/test_migration_bridge_failures.py \
+            tests/integration/test_linux_listener_tools.py \
+            tests/integration/test_self_update_status_driver.py \
             tests/integration/test_self_update_upgrade_paths.py \
             tests/integration/test_qualification_https.py \
             tests/integration/test_plugin_reload_persistence.py \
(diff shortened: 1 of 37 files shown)


dsarno:
@coderabbitai full review

PR #1049, opened 2026-09-11
feat: add VisualShader introspection, node catalog, and editing

michaltomczykowski:
Adds the read/edit half of the VisualShader surface on top of #1041.

Target milestone: **4.1**.

> **Stacked on #1041.** Until that PR merges, this diff includes its three commits; review the last commit only. When #1041 lands, GitHub recalculates the diff against `main` and it shrinks to this PR's changes.

## Ops

| Op | Purpose |
|----|---------|
| `visual_shader_get(path)` | `shader_type`, per-stage nodes (id, type, position, supported params), connections, varyings. Resource form: `godot://visual_shader/{path}`. |
| `visual_shader_node_catalog(filter="", offset=0, limit=100)` | Instantiable `VisualShaderNode*` classes with the params this tool accepts, plus legacy aliases. |
| `visual_shader_edit(resource_path, operations)` | Ordered, fully validated edits: `add_node` / `remove_node` / `replace_node` / `set_node_params` / `set_node_position` / `connect` / `disconnect` / `add_varying` / `remove_varying`. |
| `visual_shader_create_graph(..., varyings=[...])` | Varying declarations (`vertex_to_frag_light` / `frag_to_light` × `float` / `int` / `uint` / `vector2..4` / `boolean` / `transform`). |

## Contract details

- `edit_graph` applies every operation to an in-memory copy; any failure returns an error and the file is untouched. Success saves through the same same-directory temp + rename as `create_graph`.
- Numeric node IDs address existing nodes (as reported by `visual_shader_get`); string IDs added in the same call are allocated integers, returned in `added` as `{id, node_id, stage}`, and can be referenced by later operations in that call.
- Engine limitation: `VisualShader` exposes varying names through dynamic `varyings/<name>` properties but has no mode/type getter; `get_graph` recovers mode/type from the serialized resource values and falls back to name-only.
- `visual_shader_node_catalog` has no port introspection to expose — `VisualShaderNode` does not bind input/output port APIs to GDScript — so it reports class names, supported params, and aliases.

## Validation

- focused Python suites (visual shader, server instructions, resource-form lint, resources, tool domains) — 57 passed
- live Godot 4.7 `test_run` — 2302 passed, 0 failed, 46 skipped, including 21 `visual_shader` tests (9 new)
- live `batch_execute` smoke: created a graph with varyings, read it back, catalogued nodes, added/connected/moved a node, added a varying, and re-read the result

## Summary by CodeRabbit

- **New Features**
  - Added VisualShader graph creation, inspection, editing, and node catalog operations through material management.
  - Added read-only access to VisualShader graphs via `godot://visual_shader/{path}`.
  - Added support for shader stages, nodes, parameters, connections, and varyings.
  - Added validation, atomic saving, filtering, pagination, and safeguards against invalid edits.

- **Documentation**
  - Added comprehensive VisualShader authoring guidance, API details, limits, and workflows.

- **Tests**
  - Added extensive coverage for VisualShader creation, editing, retrieval, validation, persistence, and integration.


michaltomczykowski:
Verified against the current code: `get_graph` validates through the project's canonical `McpPathValidator.path_error`, the same as every other load handler (`material_manage get`, `resource_manage load`, `script_manage read`, …).

Symlink resolution is a documented, accepted limitation of that validator — its header states containment is lexical (`globalize_path` + `simplify_path` prefix match) and deliberately does not resolve symlinks, because GDScript exposes no realpath and the loopback trust boundary is the primary control:

> Known limitation: containment is lexical … It does NOT resolve symlinks — GDScript exposes no realpath. A symlink *inside* the project that points outside it can therefore defeat the under-root check. This matches the engine's own `res://` resolution and is accepted; the loopback trust boundary is the primary control.

So this is not specific to this PR: every load handler shares it, and a fix belongs in the validator (or a realpath seam), not in `visual_shader_get` alone. Happy to take that on as a separate hardening PR if maintainers want it.

michaltomczykowski:
@coderabbitai, based on the above — the symlink limitation is documented and accepted in `McpPathValidator`'s own contract, and every load handler shares it — please withdraw this finding.

dsarno:
Reviewed incremental `2729d153..a038006b`. Readback and supported-node discovery fit the core graph capability; I would qualify those first, then editing/varyings separately. Two concrete blockers emerged from the exact handler in native Godot 4.7.2.

- `_connect_nodes` line 683 formats two placeholders with three values (`stage`, `target`, `to_port`). It logs a formatting error and makes distinct destination inputs collide. A two-constant graph connected to separate output inputs succeeds on #1041 and fails on this head.
- A declared varying cannot be used by `VisualShaderNodeVaryingGetter` through `params.varying_name`: the property is rejected. Declaration/readback tests do not prove a functioning cross-stage varying.

Please use one consistent occupied-input representation and test multi-edge creation plus disconnect/remove/reconnect sequences; `_collect_inputs` and disconnect currently mix key shapes too (source finding beyond the reproduced formatting failure). Either complete varying-node configuration with an actual value-flow test or defer that scope. The invalid-identifier issue raised on #1041 is also inherited. These were native handler probes, not exhaustive PR qualification.

michaltomczykowski:
Addressed the three current CodeRabbit findings on `9ed613e`:

- **Connection dedupe key (`_connect_nodes`)** — the key formatted three values through two placeholders, so every connection collapsed to one literal and the second connection in a single `create_graph`/`edit_graph` call was rejected as duplicate input. Reproduced live before the fix ("invalid, duplicate-input or cyclic connection" for two connections into distinct output ports), then fixed; `_collect_inputs`, `_connect_nodes` and the disconnect release now share one `stage:target:port[:source]` shape.
- **Disconnect release** — both keys recorded by `_collect_inputs` are erased, so a disconnect followed by a reconnect to the same input works in one `edit_graph` call.
- **Varying node params** — `varying_name`/`varying_type` are accepted on VaryingGetter/Setter (with the identifier guard `parameter_name` already gets) and reported through `get_graph`; the docs list them.

Regressions: two connections in one create/edit call, disconnect-then-reconnect the same input, and a declared varying referenced from a varying node.

The earlier round's four findings are addressed on the current head (`replace_node` IMPLICIT merge, stage-scoped aliases + duplicate rejection, mode-aware `_edit_stage`, exact `_varying_from_text` match).

The outside-diff symlink-containment finding is intentionally not changed here: `path_validator.gd` documents lexical containment as an accepted limitation ("does NOT resolve symlinks — GDScript exposes no realpath … accepted; the loopback trust boundary is the primary control"), and it applies to every path-based read handler rather than this PR.

Validation on `9ed613e`: `ruff check src/ tests/` clean; targeted Python tests pass; live Godot 4.7.2 `test_run` — visual_shader 32/32 (including the three new regressions), full suite 2361 passed / 34 skipped / 0 failed.

dsarno:
Maintainer verification passed for ecdb176a763c69191ed83df09380078ffdd52a89, tree 31e2ff9a584e943e6d5c5e9019ae8bd225910d71, against main 5d483bdbad19a9486918e65dd519db153ed6fc0f. Independent source review passed. The maintainer correction rebuilds remaining input edges after node removal; its regression checks the replacement connection and stored color.

Ruff passed, Python 2803 passed / 122 skipped, live Godot 2445 passed / 0 failed / 31 skipped, five authenticated public graph calls passed, and native updater 13 passed / 11 skipped. Connected varying rendering was also verified after save/load. Current CI is green. The final ancestry reconciliation preserved the exact tested source tree.

Issue #1053, opened 2026-09-11
physics_shape_generate: hands-on test results — exact fits, honest errors, and 3 gaps (convex/trimesh, sibling layout, no rigid body)


dsarno:
Hands-on test of `resource_manage(op="physics_shape_generate")` — the op introduced by #892 (@michaltomczykowski) and shipped via #1028 in 4.0.4. Driven live over MCP against Godot 4.7-beta3 on 2026-09-11, in a scratch `Node3D` scene with primitive meshes at non-default sizes so fits could be checked exactly.

## What works (well)

Every generated shape matched its source mesh exactly:

| Mesh | Requested | Generated | Fit |
|---|---|---|---|
| BoxMesh size (2,1,3) | auto | BoxShape3D size (2,1,3) → StaticBody3D | exact |
| SphereMesh r=0.75 | sphere / static | SphereShape3D r=0.75 → StaticBody3D | exact |
| CapsuleMesh r=0.4 h=2.2 | capsule / area | CapsuleShape3D r=0.4 h=2.2 → Area3D | exact |
| CylinderMesh r=0.6 h=2 | cylinder | CylinderShape3D r=0.6 h=2 → StaticBody3D | exact |
| TorusMesh inner 0.5 / outer 1.5 | auto | BoxShape3D size (3,1,3) | correct AABB |

Error surface is exemplary — every negative case failed honestly with an actionable message:

- Polygon2D target → `WRONG_TYPE: Node at … is Polygon2D — must be MeshInstance3D`
- MeshInstance3D with no mesh → `RESOURCE_NOT_FOUND: … has no mesh resource — there are no bounds to fit`
- bad `shape_type` / `body_type` → `VALUE_OUT_OF_RANGE` listing the valid values
- re-run on an already-wrapped mesh → `VALUE_OUT_OF_RANGE: … already has a collider sibling at … — remove or rename it first`

Multi-path batches are **atomic**: `paths=[freshMesh, meshlessNode]` failed on the meshless node and created nothing for the fresh one (a subsequent solo run on it succeeded, which it couldn't have if a collider had been left behind). All mutations report `undoable: true`.

## Gaps / suggestions

[...]

PR #1063, opened 2026-09-16
add physics shape generate follow-ups from #1053

### LINKED ISSUE #868: Proposal: collision_manage MCP tool for bulk 3D mesh collider generation

michaltomczykowski:
**Stacked on #1040** — the first 7 commits are that PR's; only `e42d318` belongs to this one. Until #1040 merges, review the last commit (the earlier commits are unchanged from #1040).

Completes the remaining `resource_manage(op="physics_shape_generate")` gaps reported in #1053:

- **`shape_type: convex | trimesh`** — derived from the mesh via `Mesh.create_convex_shape()` / `Mesh.create_trimesh_shape()`, with the mesh's own scale baked into the vertices and the generated `CollisionShape3D` left at an identity transform, so Godot never has to scale a hull or concave shape. A faceless mesh (e.g. `PointMesh`) fails the whole request before any mutation. `trimesh` is refused for `rigid`/`character` bodies (a concave shape only simulates on a static body or an area), and both hull types are refused under a non-uniformly scaled parent chain, like the round primitives.
- **`body_type: rigid | character`** — the two body types issue #868's spec asked for.
- **`reparent_mesh` (default `false`)** — wraps the mesh under the generated body (`Body → [MeshInstance3D, CollisionShape3D]`) while preserving its world transform, so moving the body moves the visual. Undo restores the mesh's original parent, sibling index, transform and owner; a failed batch unwraps any mesh it had already wrapped; `top_level` meshes are refused.
- **Float-noise snap** — collision offsets below `1e-6` become exactly zero, so a centered `CapsuleMesh` no longer carries a `1.19e-07` Y offset.

Validation:

- `ruff check` clean (CI scope).
- `pytest -m "not editor"`: 2546 passed, 43 skipped, 74 deselected. The 24 failures / 107 errors are pre-existing environment rows (this box's `C:\Windows\System32\bash.exe` is a WSL relay with no `/bin/bash`, plus openssl/qualification harness rows) and fail identically on the base branch.
- Live Godot 4.7 (headless editor on the branch, `dev_venv` server): full `test_run` **2336 passed, 0 failed, 44 skipped**; `physics_shape` suite **61/61**.
- Live MCP smoke of every new option: reparent wrap + world transform, convex/trimesh sub-resources present in the saved `.tscn`, rigid/character node types, `trimesh`+`rigid` refused with an actionable message and no partial creation, capsule offset exactly zero.

Closes #868. Refs #1053 — items 4 (`overwrite`), 5 (mesh-class-aware auto-selection) and 6 (inline sub-resources in `node_get_properties`) are intentionally out of scope here; happy to file them as separate issues.

## Summary by CodeRabbit

- **New Features**
  - Generate convex and trimesh collision shapes from meshes.
  - Support rigid and character body types.
  - Optionally reparent meshes under generated bodies while preserving transforms and undo behavior.
  - Snap very small collision offsets to zero.

- **Bug Fixes**
  - Improve handling of disconnected, abandoned, or incomplete generation requests.
  - Skip invalid or deleted scene nodes during searches and hierarchy inspection.
  - Reject unsupported mesh and scaling combinations with clearer validation.

- **Documentation**
  - Updated physics shape generation options, limitations, and undo behavior.


michaltomczykowski:
Both review findings are fixed in `d52bab7`:

- **Refused-entry leak** — the null-shape branch in `_create_generated_entry` now frees the detached `collision` and `body` before returning. `test_generate_degenerate_hull_frees_detached_nodes` locks it with an orphan-set diff (`Node.get_orphan_node_ids()`), driving the entry builder directly because the planning pre-check keeps faceless meshes out of the handler. The test also proves the orphan diff sees detached nodes before relying on it.
- **Mirrored trimesh winding** — `_fit_mesh_shape` now swaps two vertices per triangle when the baked scale has a negative determinant. Verified against the engine before/after: mirrored signed volume `+6.0` → `-6.0`, matching the unmirrored generation; `test_generate_mirrored_trimesh_restores_winding` compares the signed volumes of the mirrored and unmirrored colliders so the assertion doesn't hard-code Godot's winding convention.

Docstring coverage: docstrings added to the new test functions and the inline `respond` helper.

Re-validated on the fix commit:

- `ruff check` clean; `pytest -m "not editor"` — targeted `physics_shape` unit + MCP dispatch rows pass.
- Live Godot 4.7 `test_run`: full suite **2338 passed, 0 failed, 44 skipped**; `physics_shape` **63/63**.

michaltomczykowski:
Fixed in `213ffd1`:

- `_restore_reparented_mesh` now reads the captured `mesh_parent` **untyped** (the typed `Node` local raised on a freed object before `is_instance_valid()` could inspect it — same pattern as `_generate_rollback`).
- When the captured parent is invalid, the mesh is reattached to wherever the body currently lives (or detached when the body has no parent) before returning, so the rollback's `body.free()` can no longer take the user's mesh with it.

`test_generate_rollback_keeps_mesh_when_original_parent_is_gone` drives the restore directly with a freed captured parent and asserts the mesh survives under the body's parent after the body is freed.

Re-validated: live Godot 4.7 `test_run` full suite **2339 passed, 0 failed, 44 skipped**; `physics_shape` **64/64**.

dsarno:
Reviewed all three incremental commits `47afec57..213ffd10`, separately from the unresolved #1040 base. Convex/trimesh generation is a good extension of the existing core resource op, but the dynamic-body default needs correction.

`body_type="rigid"` still defaults to `reparent_mesh=false`. A native probe using the exact entry-construction helper shows the generated body falling about 2.75 units over 45 physics frames while the visual mesh remains stationary. Character bodies have the same detached hierarchy when moved. This is the body/visual coupling problem called out in the original #868 maintainer feedback.

Require wrapping for the newly supported dynamic body types, choose a safe dynamic default, or defer them; preserve the published static/area default. Test actual motion as well as editor transforms and undo/redo. Mesh-derived hull work also needs a measured per-mesh bound: the frame budget is checked between items, while one convex/trimesh operation and its vertex loops are synchronous. No large-mesh hang was reproduced in this review. #1040 remains a blocking dependency.

michaltomczykowski:
Thanks @dsarno — rebased onto #1040's current head (`3c037fb`) and fixed on `5d34899`.

**Dynamic bodies always wrap.** `rigid`/`character` now default `reparent_mesh=true`, and an explicit `false` is refused with guidance ("must own its visual mesh"); `static`/`area` keep the published `false` default. The rebase also taught `_applied_stale_reason` that a wrapped mesh's expected parent is the generated body — without that, #1040's new applied-entry revalidation would have failed every wrapped entry.

**Per-mesh bound.** Planning measures the extracted triangle count and refuses `convex`/`trimesh` above 10,000 triangles, naming the count and the cap. Measured on Godot 4.7.2 (Windows, headless): 10k triangles cost ~4.9 ms face extraction plus ~7.9 ms convex / ~11.6 ms trimesh, about one 60 Hz frame; 50k already reaches ~28 ms plus ~99 ms trimesh, which is what the cap prevents.

**Tests.** In-suite (synchronous — the runner does not await coroutines): the dynamic default wraps, the mesh's world transform is preserved, moving the body the way a physics step would carries the mesh, and undo/redo restores the unwrapped layout; an explicit `reparent_mesh=false` with a dynamic body is refused; a non-boolean flag is a type error; an over-cap mesh is refused before any hull work. The rebased #1040 regressions are included.

**Actual motion smoke** (your probe shape): the handler's exact entry helper, real physics for 45 frames — wrapped body_y 7.250 / mesh_y 7.250, detached body_y 7.250 / mesh_y 10.000. The wrapped visual falls with the body; the detached one stays behind.

Validation on `5d34899`: `ruff check src/ tests/` clean; physics_shape 70/70; full suite 2357 passed / 31 skipped / 0 failed. CI is running on the new head.

dsarno:
Published the maintainer correction at a98154be1fa0164332840f5a04a4e8a9663285f2, preserving the original contributor commits.

Convex/trimesh generation now checks bounded mesh metadata before extracting geometry and rechecks before applying a deferred item. Shape validation happens before body allocation. Unsupported mesh types fail explicitly. The limits bound inputs; they do not promise a universal engine-call duration.

Verified the exact tree with Ruff, Python (2783 passed / 122 skipped), live Godot (2424 passed / 0 failed / 31 skipped), actual editor Undo/Redo for rigid and character ownership, three public-tool checks, and native updater tests (13 passed / 11 skipped). CI is now running on this head; merge remains pending those results.

PR #1064, opened 2026-09-16
extend animation track-value coercion for typed properties

michaltomczykowski:
Extends animation track-value coercion so existing animation operations accept the JSON shapes for eight more property types (the `docs/implementation-plan.md` Tier 1 "animation gaps").

## Coercion

`AnimationValues.coerce_for_type` now coerces:

- `Transform3D` — `{basis, origin}` (canonical round-trip) or `{position, rotation_degrees?, scale?}` (scale applied in local axes)
- `Quaternion` — `{x,y,z,w}` / `[x,y,z,w]`, normalized at the animation boundary; zero-length refused
- `Basis` — `{x,y,z}` axis vectors / 3×3 nested array
- `Vector3i` — `{x,y,z}` / `[x,y,z]` with int32 destination-range validation
- `Rect2` / `AABB` — `{position, size}` / arrays
- `NodePath` / `StringName` — strings

Strict within each shape: missing components are refused with a typed message, and non-finite components are refused (`1e40` overflows float32 to INF).

## Scope

Per review, this PR is the coercion work only. The four content presets (bounce/orbit/sweep/drift) and the generalized `preset_pulse` were moved out of core: they are preserved on `feat/animation-presets` (pushed to the fork) for a separate review, and the recipes are documented in `docs/animation-recipes.md` with a custom-tool addon example.

## Validation

- `ruff check` clean; `pytest -m "not editor"` 2673 passed / 45 skipped.
- Live Godot 4.7.2: `animation` suite 97/97; full `test_run` 2343 passed / 0 failed.
- Live MCP smoke: Vector3i `[2147483648,0,0]` and `[1e40,0,0]` refused; Quaternion `[0,0,0,2]` stored as `(0,0,0,1)` and interpolates unit; `[0,0,0,0]` and `[1e40,0,0,1]` refused — no engine script errors.

## Summary by CodeRabbit

- **New Features**
  - Added support for animation values including Vector3i, Quaternion, Basis, Transform3D, Rect2, AABB, NodePath, and StringName.
  - Added validation for finite values, integer ranges, and valid quaternion rotations.
  - Added documented bounce, orbit, sweep, and drift animation recipes, including custom-tool packaging guidance.
  - Added an animation showcase demonstrating motion effects and typed-value tracks.

- **Documentation**
  - Updated the animation operation list to reference motion recipes separately from built-in presets.

- **Bug Fixes**
  - Invalid typed animation values are now rejected instead of being stored as raw input.


michaltomczykowski:
All four findings are fixed in `9aff41f`:

1. **Preset baselines** — `preset_bounce` now pops from `target.scale` (peak/dip multiply that baseline), and `preset_sweep` starts at the target's current rotation (`rotation.y` for 3D, `rotation` for 2D) and adds the turn from there. Both targets are read untyped, since `scale`/`rotation` live on Control/Node2D/Node3D rather than the `Node` base. Regression tests use a `(2,3,4)` scale and a `0.5` rad starting rotation.
2. **Orbit shape** — `preset_orbit` samples `_ORBIT_SEGMENTS = 16` angular segments (17 keys) instead of four quarter-turn keys; chord error is now under ~2% of the radius. The test asserts a 45° sample lies on the circle, which the four-key diamond could not produce.
3. **Drift loop guard** — `preset_drift` refuses `loop_mode="linear"` with an actionable message pointing at `pingpong`/`none`; the tool description was updated to match. The test asserts the refusal.
4. **Local-axis scale** — `parse_transform3d` uses `scaled_local(scale)`. Verified against the engine first: for a 45° rotation with scale `(2,1,1)`, `scaled()` produces `x=(1.414, 0, -0.707)` (skewed), `scaled_local()` produces `x=(1.414, 0, -1.414)` (parallel to the rotated axis). The test asserts the x axis stays parallel to the unrotated x axis and carries the scale length, so it does not depend on `scaled_local` itself.

Re-validated: `ruff check` clean; animation/error-code pytest rows pass (56); live Godot 4.7 full `test_run` **2341 passed, 0 failed, 44 skipped**; `animation` suite **106/106**.

dsarno:
Reviewed exact head 9aff41f77a1bfa4ea8be49baae59010ee27f47e0, including native Godot4.7.2 coercion and Animation interpolation probes. Needs changes and a scope split.

The typed-value fixes for existing animation operations belong in core. The nonuniform Transform3D round-trip passed and is a useful improvement. Bounce/orbit/sweep/drift are content recipes built from existing primitives; the current evidence does not justify adding roughly500 lines of preset implementation to core. Prefer custom tools or documented recipes. Generalized typed pulse is closer to an existing primitive but should be reviewed separately.

Two reproduced correctness gaps in the typed-value portion:

- New Vector3i conversion accepts [2147483648,0,0] and returns (-2147483648,0,0); [1e40,0,0] becomes (0,0,0). Validate destination component range before conversion. Existing scalar integer truncation is a separate policy question, not the basis of this finding.
- Quaternion [0,0,0,2] is accepted and stored, but actual Animation.value_track_interpolate emits a normalization error and returns identity. Finite JSON [1e40,0,0,1] becomes an infinite quaternion. Define the rotation-value contract at the animation boundary: reject invalid rotations or deliberately normalize valid nonzero ones. A generic mathematical Quaternion parser need not impose rotation semantics, but the animation consumer must produce usable values.

Add destination-range and actual interpolation regressions, including valid controls. The existing stored-type and nonuniform Transform3D tests are useful; they do not exercise these numeric boundaries.

This was a focused native helper/interpolation probe and source review, not full AnimationPlayer/MCP or branch qualification. I found no specific dependency-creation/undo atomicity blocker in the reviewed paths and am not claiming one.

michaltomczykowski:
Thanks for the native probes — both gaps are fixed and the scope is split at HEAD (`a562837`):

**Scope split**
- #1064 is now the typed-value coercion only (title and description updated). `animation_presets.gd`, `animation_handler.gd`, `plugin.gd`, the Python handlers, `docs/TOOLS.md` and the preset test rows are back to main.
- The four recipes and the generalized pulse are preserved on `feat/animation-presets` (pushed to the fork) for a separate review; nothing was lost.
- The recipes are documented in `docs/animation-recipes.md` — core-op keyframe recipes for bounce/orbit/sweep/drift (Control pivot-offset note, drift's loop-mode guidance) plus a complete custom-tool addon example for wrapping project-specific recipes.

**Vector3i range** — `parse_vector3i` validates every component against int32 bounds before conversion: `[2147483648,0,0]` and `[1e40,0,0]` are refused with `INVALID_PARAMS` ("components must be finite and within int32 range"); in-range fractional truncation is unchanged.

**Quaternion contract** — the generic parser refuses non-finite components, including values that only overflow at float32 construction (`1e40`). The animation boundary owns rotation semantics: zero-length is refused ("zero-length quaternion is not a rotation") and valid nonzero quaternions are normalized, so `Animation.value_track_interpolate` slerps unit values instead of emitting a normalization error and returning identity.

**Regressions** — coercion-level range/finiteness cases, plus actual `value_track_interpolate` tests through a new typed-property fixture (`mcp_animation_typed_target.gd`): a stored Vector3i track interpolates and out-of-range keys are refused without adding a track; `(0,0,0,2)` stores as unit and slerp stays unit; zero-length and infinite rotations are refused. Interpolation rounds Vector3i components through float32, so the track control uses representable values while the int32 bounds are covered at the coercion level. The other new parsers (Basis/Transform3D/Rect2/AABB) reject non-finite components too.

Re-validated: `ruff check` clean; `pytest -m "not editor"` 2673 passed / 45 skipped; live `animation` suite 97/97; full `test_run` 2343 passed / 0 failed; live dispatcher smoke of all five repro cases with no engine script errors.

dsarno:
Re-reviewed exact head `a5628377fc22841408a4a65276344f66407fa03e` with native Godot 4.7.2 helper and interpolation probes.

The scope split addresses the core/custom-addon concern. The original Vector3i repros now reject correctly, both int32 endpoints remain valid, and Quaternion `[0,0,0,2]` now stores a unit value and interpolates without an engine error.

One numeric-boundary defect remains in the new normalization path. Finite JSON `[0,0,1e20,1e20]` passes `parse_quaternion`, but its float32 squared length overflows. `coerce_for_type` then returns `{"ok": Quaternion(0,0,0,0)}`. Actual `Animation.value_track_interpolate` emits `The end quaternion (0, 0, 0, 0) must be normalized` and returns identity. `[0,0,1e30,1e30]` reproduces the same result. Checking component finiteness before normalization does not guarantee a usable normalized result.

Please make normalization safe across the accepted component range and validate the resulting rotation before reporting success. Include a large finite input regression through actual interpolation. The existing small-magnitude review finding is also reproducible: `[0,0,1e-7,1e-7]` is nonzero but is currently rejected as zero-length. Both ends should follow the stated contract for normalizing valid nonzero rotations, or use an explicit narrower contract with accurate errors.

This is a focused native helper/interpolation follow-up with the addon disabled, not full MCP/AnimationPlayer or branch qualification. Import and probe processes exited normally; the interpolation errors above are the reproduced finding, not a passing behavioral result.

michaltomczykowski:
Fixed at HEAD (`eefd4fe`).

**Root cause.** `length_squared()` on float32 components: `(0,0,1e20,1e20)` squares to INF, so the `< 1e-12` check was false and `normalized()` divided by INF, returning `(0,0,0,0)` as a success. At the other end `(0,0,1e-7,1e-7)` squares to `2e-14`, below the threshold, so it was refused as zero-length.

**Fix.** Normalization now scales by the largest component first — components are ≤ 1 after scaling, so neither overflow nor underflow can occur — refuses only an exactly zero-length quaternion, and validates the result (finite components, unit length within float32 precision) before reporting success, so an unusable rotation fails at coercion instead of reaching slerp as zero.

**Regressions.** Coercion: `(0,0,1e20,1e20)` and `(0,0,1e30,1e30)` normalize to unit `(0,0,√½,√½)`; `(0,0,1e-7,1e-7)` is accepted; `(0,0,0,0)` still refuses with "zero-length"; `1e40` still refuses as non-finite. Interpolation: a track keyed with `(0,0,1e20,1e20)` and `(0,0,1e30,1e30)` is stored and `Animation.value_track_interpolate` returns the actual unit rotation at t=0 and a unit value at t=0.5 — with the old code the t=0 read was the engine's identity fallback after its normalization error.

Also fixed the CodeRabbit doc finding in the same commit: the orbit recipe now gives a two-component value for 2D/Control targets (three-component for 3D), since typed coercion rejects the wrong shape.

Validation on `eefd4fe`: animation suite 97/97, full suite 2345 passed / 31 skipped / 0 failed. CI is running.

dsarno:
Independent review and the combined live verification passed. Scale-first normalization handles very large and tiny nonzero quaternion inputs without overflow/underflow, while zero and nonfinite values are rejected. The regressions inspect stored typed values and actual animation interpolation.

Ruff, full Python, live Godot, and explicit native updater checks passed on the combined current-main candidate. No further contributor changes requested.

PR #1065, opened 2026-09-17
add physics_manage for body configuration and collision layers

michaltomczykowski:
Adds the `physics.*` helpers tracked in `docs/implementation-plan.md` ("layer/mask/body helpers still pending") as one new `physics_manage` rollup — no new named verbs.

## Ops

- **`body_configure(path, collision_layer=None, collision_mask=None, gravity=None, gravity_scale=None, mass=None, linear_damp=None, angular_damp=None, continuous_cd=None, freeze=None, priority=None, monitoring=None, monitorable=None, physics_material_override=None, scene_file="")** — any subset in **one undo action**.
  - `collision_layer` / `collision_mask` accept a bitmask int **or** an array of layer names defined via `layers_set` (`["player", "enemy"]`).
  - Class applicability is checked against the node's own property list: `mass` on a `StaticBody3D` is `PROPERTY_NOT_ON_CLASS`, not a silent write.
  - `physics_material_override` takes a `res://` `PhysicsMaterial` path or `""` to clear.
  - Returns `applied` + `previous` so the caller can see exactly what changed.
- **`body_get(path)`** — reads the configuration back, including `collision_layer_names` / `collision_mask_names` (unnamed bits report as `layer_N`).
- **`layers_get(dimension="3d")`** — the project's named physics layers as `[{index, bit, name}, …]`.
- **`layers_set(dimension, layers)`** — `{layer_index: name}` for indexes 1-32, saved to `project.godot` (`undoable: false` with a reason, mirroring `project_manage.settings_set`).

## Surface wiring

`domains.py`, `server.py` (registration + instructions rollup block), `tool_catalog.gd`, `plugin.gd` (one lazy handler + four commands), `docs/TOOLS.md`, and the tool-surface counts in `AGENTS.md` / `docs/tool-surface.md` (47 tools now).

## Validation

- `ruff check` clean; `pytest` domain/instructions/command-parity/physics rows pass (51).
- Live Godot 4.7: full `test_run` **2332 passed, 0 failed, 44 skipped**; `physics` suite **13/13**.
- Live MCP smoke: `layers_set` → `layers_get` roundtrip with bit metadata, name-array `body_configure` (bits `1<<19 | 1<<20`), `body_get` with resolved names, Area3D `monitoring`/`priority`, `mass` refused on Area3D, unknown layer name refused naming the defined set. Project layer names were restored and `project.godot` reverted after the smoke.

One guardrail updated: `test_dispatcher.gd`'s lazy-handler count 30 → 31.

## Summary by CodeRabbit

* **New Features**
  * Added physics management tools for reading and configuring 2D and 3D bodies and areas.
  * Added support for retrieving and updating project-wide physics layer names.
  * Physics body changes support validation and undo; layer updates persist to project settings.

* **Bug Fixes**
  * Improved validation for physics layer names, including duplicate and ambiguous assignments.

* **Documentation**
  * Updated tool catalogs, capability guidance, and counts to include the new physics management tools.


michaltomczykowski:
Both findings are fixed in `59222c1`:

1. **Non-string layer names** — `layers_set` now requires each name to be a `String` and returns `WRONG_TYPE` otherwise, instead of stringifying arbitrary values into `project.godot` (a dict would have been persisted as its debug repr). Test covers `{"11": 42}` and asserts nothing was written.
2. **Duplicate names** — `layers_set` refuses an update that would leave one name on two indices, reporting the conflicting indices, and `_layer_name_index` now maps each name to *every* index carrying it so `_coerce_layers` fails an ambiguous name with the conflicts named rather than silently using the last one.

Two refinements beyond the literal suggestion, because the obvious strict version was a trap:

- The uniqueness check runs on the **post-update state but only for the names the call touches**. A strict whole-map check would have blocked unrelated layer edits whenever a hand-edited `project.godot` already contained duplicates — and the error tells the user to fix duplicates with `layers_set`, which would then be unable to run. Moving a name between indices in one call (`{12: "moved", 13: "taken"}` where 12 currently holds `taken`) is likewise allowed.
- While testing this I hit a genuine test-helper bug: `_save_layer_names` replaced the per-dimension restore record, so two tests in the suite overwrote each other and left `physics_probe_*` names in `project.godot` between runs. It now merges and keeps the earliest captured value, which is what made the second full run pass cleanly.

Re-validated: `ruff check` clean; physics pytest rows pass (18); live Godot 4.7 `physics` suite **16/16**; full `test_run` **2335 passed, 0 failed, 44 skipped**, with `layers_get` confirmed empty afterwards (no leaked names).

dsarno:
Reviewed exact head `59222c1c`. I recommend an optional custom tool first, rather than another core rollup: body properties, batched undoable writes, material assignment, and project layer settings already have core paths. The distinct convenience here is named-layer resolution and curated presentation; that alone does not yet justify a second body-property vocabulary.

Native Godot 4.7.2 probes also found two correctness blockers:

- `layers_set({"1.9":"Unexpected"})` persists layer 1, and key `"2junk"` persists layer 2. Validate the entire integer representation and collisions before writing. Fractional bitmasks are also silently truncated.
- `body_configure(mass=-5, gravity_scale=2)` returns success with `applied.mass=-5`, while Godot rejects the mass and keeps 1 but applies gravity_scale=2. This is partial mutation with an inaccurate success response.

Validate semantic constraints before committing, and assert actual stored values and no mutation on invalid requests. These findings used actual handlers and an editor undo manager in an isolated fixture; the whole PR was not qualified after the blockers were reproduced.

dsarno:
A concrete migration example for the custom-tool recommendation is your [Godot AI Terrain Tools addon](https://github.com/michaltomczykowski/godot-ai-terrain-tools), which followed the same path from [the core terrain PR #856](https://github.com/hi-godot/godot-ai/pull/856#issuecomment-5410618075).

Its [plugin.gd registration code](https://github.com/michaltomczykowski/godot-ai-terrain-tools/blob/main/addons/godot_ai_terrain_tools/plugin.gd) is a useful starting point. It registers tool specs, handles registry readiness/reload, and unregisters its source on disable. The supported contract is documented in [Custom tools: third-party addons](https://github.com/hi-godot/godot-ai/blob/main/docs/plugin-architecture.md#custom-tools-third-party-addons).

For this PR, keep the physics handler and relevant tests in a standalone addon, and replace the proposed Python wrappers and core catalog registration with custom-tool specs. The operations can be invoked through `custom_manage`, with selected high-use operations optionally promoted to named `custom_*` tools. This lets the named-layer and body-configuration conveniences evolve without requiring another core rollup.

Keep parameter validation in the handler; advertising a schema does not validate the values automatically. Set readiness and undo metadata per operation. Scene mutations should retain editor undo support, while project-settings file writes should report their actual non-undoable behavior. The malformed layer/mask inputs and partial body mutation described above still need fixes and stored-value/no-mutation regression coverage in the addon.

dsarno:
Closing following the maintainer decision that these physics helpers belong in an optional addon, outside the core tool surface. The existing core operations already cover the underlying body properties and project settings; the named-layer and configuration conveniences can evolve independently as custom tools. Concrete migration guidance and the terrain addon precedent are linked here: https://github.com/hi-godot/godot-ai/pull/1065#issuecomment-5734996110 . The validation and partial-mutation findings still need to be addressed wherever this ships: https://github.com/hi-godot/godot-ai/pull/1065#issuecomment-5733973514 . Thank you for contributing this work.

PR #1066, opened 2026-09-17
add theme texture/font/icon slots, per-node stylebox override, and richtext

michaltomczykowski:
Closes the remaining theme/UI gaps from `docs/implementation-plan.md` Tier 1 ("pixel-art / custom typography (`theme_set_font`, `theme_set_stylebox_texture`) still pending", "`node_stylebox_override`", "`ui_set_richtext`"). Four ops on the existing `theme_manage` / `ui_manage` rollups — no new tools.

## `theme_manage`

- **`set_stylebox_texture(theme_path, class_name, name, texture_path, region?, margins?, axis_stretch_horizontal?, axis_stretch_vertical?, modulate_color?, draw_center?)`** — composes a 9-slice `StyleBoxTexture` from an imported `Texture2D`. `region` is `{position, size}` or `[x,y,w,h]`; `margins` is `{all|left|top|right|bottom}`; stretch modes are `stretch | tile | tile_fit`.
- **`set_font(theme_path, class_name, name, font_path)`** / **`set_icon(theme_path, class_name, name, texture_path)`** — load, class-validate, and assign `Font` / `Texture2D` resources to theme slots (undoable).
- **`stylebox_override(path, slot, patch)`** — duplicates the stylebox the Control resolves for `slot`, applies the same `StyleBoxFlat` patch vocabulary as `set_stylebox_flat`, and attaches it via `add_theme_stylebox_override`. Undo restores the previous override or removes it. The zero-border angular frame / flash-the-bar-background pattern without mutating the shared theme. Non-`StyleBoxFlat` slots are refused with the resolved class named.

The `StyleBoxFlat` property parser was extracted into one `_apply_flat_props` helper now shared by `set_stylebox_flat` and `stylebox_override`.

## `ui_manage`

- **`set_richtext(path, text, bbcode=True)`** — sets a `RichTextLabel`'s text with BBcode parsing on by default (`bbcode=False` writes literal text), undoable in one action.

## Validation

- `ruff check` clean; theme/richtext/parity/error-code pytest rows pass.
- Live Godot 4.7: full `test_run` **2332 passed, 0 failed, 45 skipped**; `theme` suite **40/40**, `ui` suite **46/46**.
- Live MCP smoke: `set_stylebox_texture` with per-side margins + tile stretch, `set_font` refusing a non-`Font` resource, `stylebox_override` on a Panel, `set_richtext` verified through `node_get_properties` (`bbcode_enabled` true, markup text stored). Fixture files removed and `project.godot` reverted after the smoke.

Two test-suite fixes along the way: theme fixtures are now suite-scoped (a per-test delete left a stale `ResourceLoader` cache), and the new undo tests clear the undo history first so earlier suites' actions can't be undone instead.

## Summary by CodeRabbit

- **New Features**
  - Added RichTextLabel editing with optional BBCode rendering and undo support.
  - Added theme tools for texture-based styleboxes, font and icon assignment, and undoable stylebox overrides.
  - Added strict validation for theme values, with failed changes leaving existing settings and undo history unchanged.

- **Documentation**
  - Updated tool listings and operation descriptions for the new UI and theme capabilities.

- **Tests**
  - Added coverage for UI/theme operations, validation errors, parameter handling, and undo behavior.


michaltomczykowski:
All three findings are fixed in `d9242b1`:

1. **File encoding** — `test_theme.gd` is now UTF-8 **without** a BOM, and the one remaining mangled character (a `→` that had been read through cp1252) is restored. The earlier `df729d4` repaired the seven em-dashes but left the BOM and that arrow; both are gone now, and I verified no unexpected non-ASCII remains.
2. **Axis-stretch docs** — the `set_stylebox_texture` doc comment lists `tile_fit`, matching `_parse_axis_stretch` and its error message.
3. **Strict `bbcode`** — `set_richtext` returns `WRONG_TYPE` for a non-boolean `bbcode` instead of coercing: `"false"` is truthy in GDScript, so the old `bool(...)` would have silently rendered markup as literal text. The default `true` still applies only when the key is absent; the test covers the string case.

Re-validated: `ruff check` clean; theme/richtext pytest rows pass (24); live Godot 4.7 `theme` suite **40/40**, `ui` suite **46/46**, full `test_run` **2333 passed, 0 failed, 44 skipped**.

dsarno:
Reviewed `d9242b17`. Theme font/icon/texture slots and per-Control stylebox overrides are reasonable core additions to the existing domains. There is a concrete validation blocker before approval.

In a native Godot 4.7.2 probe of the exact handler, `_apply_texture_margins(sb, {"left": 2, "right": {"bad": true}})` emits `SCRIPT ERROR: Invalid call. Nonexistent float constructor`, but returns the empty string used for success, with left=2 and right=0. `set_stylebox_texture` therefore proceeds past its error check and can commit/save the partly prepared stylebox.

The shared flat parser has the same failure shape: a valid red background plus `border: {"all": {"bad": true}}` produces an int-constructor error yet returns `{ok: true}`. That flat-side weakness predates this extraction, but the new override operation inherits it and the new texture helper repeats it.

Please validate the numeric values before applying them, using structured errors for invalid/nonfinite inputs. Add regressions that assert refusal, unchanged actual theme/override and undo state, and no engine script error. The native evidence here exercises the helpers directly; it is not an end-to-end MCP test.

michaltomczykowski:
Thanks for the native probe — both failure shapes are fixed at HEAD (`b7245de`):

**Root cause.** `_apply_sides` / `_apply_texture_margins` / the shadow fields converted arbitrary Variants with `int()`/`float()`. A dict/array raised the engine conversion error and yielded 0, but the helper still returned the empty-string success sentinel, so the op proceeded to commit and save a partly prepared stylebox. `bool(...)` had the same silent-coercion problem for `anti_aliasing`/`draw_center`.

**Fix.**
- New `_parse_number_field(dict_name, key, raw, value_type)`: accepts int/float and numeric strings, refuses non-numeric input with `WRONG_TYPE` (`'border.all' must be a number, got Dictionary`) and non-finite results with `VALUE_OUT_OF_RANGE` (`'margins.all' must be finite, got inf`) *before* any conversion, so no engine script error can fire.
- `_apply_sides` now parses every key into a value map first (`all`, then side overrides) and only then assigns; it returns a structured result and takes an `Object` so `StyleBoxTexture` shares it. `_apply_texture_margins` delegates to it instead of duplicating the unchecked `float()` path.
- Shadow `size`/`offset_x`/`offset_y` use the same parser; `anti_aliasing`/`draw_center` require real booleans (`WRONG_TYPE` otherwise), consistent with the `bbcode` fix.

**Regressions** (`test_theme.gd`), each asserting refusal + unchanged stored slot/override + no undo action; the runner's script-error capture additionally fails the test if an engine conversion error returns, so "no engine script error" is asserted implicitly:
- `test_set_stylebox_flat_rejects_invalid_numbers` — your `border: {"all": {"bad": true}}` + valid `bg_color` case, plus `margins.all = INF`, `corners.top_left = []`, `shadow.size = {...}`, `shadow.offset_x = NAN`, and `anti_aliasing: "false"`.
- `test_set_stylebox_texture_rejects_invalid_margins` — your `{"left": 2, "right": {"bad": true}}` case, `margins.all = INF`, and `draw_center: "false"`.
- `test_stylebox_override_rejects_invalid_patch_numbers` — the override path with a valid `bg_color` plus a non-numeric border, and a variant where a pre-existing override survives untouched.
- `test_side_helpers_refuse_non_numeric_values_without_mutation` — direct helper probes like yours: the error comes back and `texture_margin_left`/`border_width_top` stay 0.

**Live verification** (Godot 4.7.2, plugin loaded from this branch): `theme` suite 44/44, `ui` 46/46, full `test_run` 2349 passed / 0 failed; `ruff` clean; `pytest -m "not editor"` 2680 passed / 45 skipped. A live dispatcher smoke of both repros returned `WRONG_TYPE: 'border.all' must be a number, got Dictionary`, with no engine script error, the theme file still slot-less, and the Panel's `theme_override_styles/panel` still null.

michaltomczykowski:
Both findings are fixed in `7f36231`:

1. **Non-finite region components (major)** — `_parse_rect2` now validates every component for all three input shapes (`Rect2`, `{position, size}`, `[x,y,w,h]`) through a `_is_finite_rect2` helper and returns null when any is non-finite, so `set_stylebox_texture` refuses with `VALUE_OUT_OF_RANGE` instead of storing a NaN/INF `region_rect`. Regression: `test_set_stylebox_texture_rejects_non_finite_region` covers the array and dict shapes plus direct `_parse_rect2` probes, asserting no stored slot and no undo action.
2. **Unpatched shadow axis** — missing offset components now start from `sb.shadow_offset` instead of 0, so a patch that sets only `offset_x` keeps the resolved stylebox's `offset_y` on the `stylebox_override` path (a fresh `StyleBoxFlat` in `set_stylebox_flat` is unaffected). Regression: `test_stylebox_override_preserves_unpatched_shadow_offset`.

Re-validated: ruff clean; `pytest -m "not editor"` 2680 passed / 45 skipped; live theme suite 46/46; full `test_run` 2350 passed / 0 failed.

dsarno:
Reviewed `7f36231e`. The original dictionary-input failures are fixed: native checks now return `WRONG_TYPE`, preserve the existing override/margins, and leave undo history untouched. Partial shadow-axis patching also works and undoes correctly. The narrow theme slots and per-Control override remain reasonable core scope.

One numeric boundary remains: `_parse_number_field` validates the GDScript double before it narrows to the native property type. On actual Godot 4.7.2, `stylebox_override` with `shadow.offset_x=1e40` succeeds and installs an override with offset `(inf,5)`. `set_stylebox_texture` with `margins.left=1e40` likewise succeeds and stores infinity. `border.all=4294967296` and `corners.all=4294967296` succeed but store zero. I checked actual override/slot values and real undo entries, with valid controls and no script errors.

Please reject values outside the native destination's representable range before commit, with finite-large-number regressions asserting structured refusal, unchanged stored resource and unchanged undo history. Checking `is_finite` before narrowing does not cover these cases. This was direct native handler verification, not full MCP/branch qualification.

michaltomczykowski:
Fixed at HEAD (`dcc139f`). The probe was right: the previous check validated the GDScript double, but the engine narrows per destination — and the destinations differ:

- `shadow_size` is a genuine C++ int32, but `border_width_*` and `corner_radius_*` take ints while storing `real_t`: `2147483647` rounds up to 2³¹ and the int getter reads back `-2147483648` (and `16777217` silently rounds to `16777216`). `margins` / `shadow_offset` are float32, and Color components are float32.
- `_parse_number_field` now takes a storage kind and validates against it: int32 range for `shadow_size` / Theme constants / font sizes; float32 narrowing via `PackedFloat32Array` for margins, offsets and colors (refusing a non-finite result and storing the narrowed value); and float32 round-trip + int32 range for the float-backed border/corner ints.
- `set_constant` / `set_font_size` share `_parse_int32_value` (`4294967296` used to store `0`, and the font size `16`), and the theme color parser refuses components that overflow float32 to INF.

Regressions cover each storage kind: `border.all=4294967296` and `2147483647`, `corners.all=4294967296`, `margins.all=1e40`, `shadow.size=4294967296`, `set_constant`/`set_font_size=4294967296`, `set_color {r:1e40}`, and `stylebox_override shadow.offset_x=1e40`. Each asserts structured `VALUE_OUT_OF_RANGE`, an unchanged stored slot/override and no undo action, with valid controls (`border.all=16777216`, `margins.all=3e38`, `shadow.size=2147483647`) still storing. The runner's script-error capture keeps "no engine script error" asserted.

Live verification (Godot 4.7.2, plugin from this branch): `theme` suite 52/52; full `test_run` 2357 passed / 0 failed; `ruff` clean; `pytest -m "not editor"` 2680 passed / 45 skipped; live dispatcher smoke of all five cases returned structured `VALUE_OUT_OF_RANGE`, the theme file stayed slot-less and the Panel's `theme_override_styles/panel` stayed null.

dsarno:
Verified `dcc139f0` against the native cases from the previous review. `shadow.offset_x=1e40`, texture `margins.left=1e40`, and border/corner `4294967296` now return `VALUE_OUT_OF_RANGE` without changing the stored override/slot or adding undo history. The original malformed inputs still refuse correctly.

Valid controls also pass: a partial shadow-offset patch preserves the other axis and undo restores the original override; a texture margin of2 is stored and undo removes the new slot. The final native editor run exited normally with no engine/script errors. The previously reported narrowing blocker is resolved; narrow theme authoring remains reasonable core scope.

This was bounded direct-handler/editor verification, not the full branch or MCP qualification. One nonblocking existing edge remains: an integer string beyond signed64-bit range can print an engine conversion diagnostic before refusal; the previous conversion expression does the same, so I am not treating it as a regression introduced by this revision.

dsarno:
The earlier numeric fixes remain verified. A broader native pass on `dcc139f0` found three additional issues in a rendered Godot 4.7.2 editor:

1. **Save failures report success.** I made only a disposable theme file read-only, then called `set_font`, `set_icon` and `set_stylebox_texture`. Each returned success/undoable, and the cached Theme contained the new slots, but the file bytes stayed identical and a `CACHE_MODE_IGNORE` reload contained none of them. The new slot callbacks discard `guarded_save` errors; texture inherits the same existing pattern. Please surface persistence failures with coherent resource/undo state, and test disk readback rather than default-cache loads.
2. **Per-Control overrides enter global undo history.** After clearing both histories, a red Panel override returned undoable, but the current scene's undo returned false and left it installed. Global undo removed it and global redo restored the actual red override. Please associate the action with the target Control's scene history and test that history specifically; the helper that falls back to global undo masks this problem.
3. **Unknown top-level patch keys silently mutate.** `patch={"bg_colour":"red"}` returns success, installs an unchanged override and adds an undo entry. Reject unknown keys at the patch boundary before installing the duplicate. The existing nested `shadow.wobble` test does not cover this case.

Positive controls: writable font/icon/texture changes persisted their actual requested values, and all three resource undo/redo steps worked with cold disk reads. RichTextLabel changes also passed actual scene undo/redo, including text, BBCode and parsed text. Both owned editors exited normally; the temporary file attributes were restored.

This was focused native handler verification, not full end-to-end MCP qualification. The core scope still makes sense; these correctness fixes are the remaining blockers from this pass.

michaltomczykowski:
All three fixed at HEAD (`102aad4`).

**1. Save failures report success.** Root cause: the slot writers ran their `guarded_save` inside the undo action's do method and discarded the return. Each writer now applies and persists the change as a **pre-flight before the undo action exists**; if the save fails it restores the previous slot in memory and returns `INTERNAL_ERROR` ("Failed to save theme to …; the change was rolled back"), so the file, the cached resource and the undo history all stay at the pre-request state. The action is then registered with `commit_action(false)` — the do method already ran as the pre-flight, so it is not re-run and a second save's error cannot be swallowed.

Testing note: a read-only *file* does not reproduce the failure in-editor — Godot's safe-save writes a temp file and replaces the target, so the read-only attribute is bypassed. The regression denies writes on a dedicated theme **directory** instead (icacls deny ACE on Windows, owner permission bits on Unix), asserts the denial is effective, then calls `set_color`, `set_stylebox_flat` and `set_font`: each returns `INTERNAL_ERROR`, the cached color/stylebox are restored, the slot that did not exist stays absent, the file is byte-identical, and no undo action is committed. (A read-only file does fail headless, where safe-save is off.)

**2. Per-Control overrides enter global undo.** The action was created on the manager and its methods were bound to the handler `RefCounted`, which routes it to `GLOBAL_HISTORY`. `stylebox_override` now resolves the Control's scene history (`get_history_undo_redo(get_object_history_id(scene_root))`) and commits bound callables there, so the scene undo removes the override and scene redo restores it. Regression undoes/redoes through the scene history specifically and asserts the global history stays empty.

**3. Unknown top-level patch keys.** `_apply_flat_props` validated nested dicts only; `patch={"bg_colour": …}` installed an unchanged override plus an undo entry. The patch boundary now rejects unknown top-level keys with `INVALID_PARAMS` and the valid list before anything is installed; regression covers the typo, the untouched override and the absent undo action.

Validation on `102aad4`: theme suite 55/55 (incl. the three regressions), full suite 2361 passed / 31 skipped / 0 failed, `ruff check` clean. CI is running.

dsarno:
Maintainer verification passed on refreshed head e4f9ce8a. The shared theme mutations now use the editor's scene-aware undo manager. Actual authenticated MCP calls verified font, icon, texture slots, stylebox overrides, and rich text; stored values and real editor-menu Undo/Redo matched the intended changes.

Final local qualification passed Ruff, 2,818 Python tests, 2,495 live GDScript tests, and the full editor-marked complement (67 passed, 14 skipped). All 25 current checks are successful. Independent source review and public smoke passed with retained shutdown warnings; no clean-console claim is intended.

The earlier SDK stream-end normalization failure was fixed separately in #1087 and included in this refresh. The earlier Windows startup delay remains unexplained; its diagnostic rerun passed, which establishes intermittency rather than a fix. Neither failure is being discarded.

Merging this verified theme/UI scope now.

PR #1067, opened 2026-09-17
add navigation_manage for regions, mesh configuration, baking, and path queries

michaltomczykowski:
Adds the pending `navigation.*` family from `docs/implementation-plan.md` as one new `navigation_manage` rollup — no new named verbs. Both dimensions share the same ops, selected by `dimension` ("3d" default) or inferred from the node class.

## Ops

- **`region_create(parent_path="", dimension="3d", name="", scene_file="")`** — creates a `NavigationRegion3D` with a fresh `NavigationMesh` (or `NavigationRegion2D` + `NavigationPolygon`), undoable.
- **`mesh_configure(path, <property>=<value>, …, scene_file="")`** — curated mesh/polygon parameters in one undo action. 3D: `agent_radius`, `agent_height`, `agent_max_climb`, `agent_max_slope`, `cell_size`, `cell_height`, `border_size`, `region_min_size`, `region_merge_size`, `edge_max_length`, `edge_max_error`, `detail_sample_distance`, `detail_sample_max_error`, `geometry_collision_mask`, `vertices_per_polygon`, `geometry_parsed_geometry_type`, `geometry_source_geometry_mode`, `geometry_source_group_name`. 2D: `agent_radius`, `cell_size`, `border_size`, `parsed_collision_mask`, `parsed_geometry_type`, `source_geometry_mode`, `source_geometry_group_name`. Source selectors accept names (`mesh_instances | static_colliders | both`, `root_children | groups_with_children | groups_explicit`). Cross-dimension names report `PROPERTY_NOT_ON_CLASS`/`VALUE_OUT_OF_RANGE` instead of being stored.
- **`bake(path)`** — synchronous bake, then the baked resource is pushed to the navigation server (`region_set_navigation_mesh` / `region_set_navigation_polygon`) and the map is force-synced. Undo restores the pre-bake resource.
- **`agent_create` / `agent_configure`** — `NavigationAgent3D`/`2D` with `radius`, `height` (3D), `max_speed`, desired distances, `avoidance_enabled`, `navigation_layers`, `simplify_path`, `debug_enabled`, one undo action.
- **`obstacle_create` / `obstacle_configure`** — `NavigationObstacle3D`/`2D` with `radius`, `height` (3D), `avoidance_enabled`.
- **`path_get(from_point, to_point, dimension="3d", optimize=True, navigation_layers=1)`** — read-only path query. The map is resolved from the scene's first navigation region, so 2D queries work in a 3D-rooted scene and vice versa.

## Notes

- `map_force_update` is deprecated in Godot 4.7 as incompatible with asynchronous navigation iterations; the force-sync path disables async iterations for the call and restores the previous setting.
- The editor only rebuilds map geometry on physics frames, so a `path_get` issued in the *same* frame as a bake can return an empty path; frame-separated calls (and live games) get the baked path. Documented in both op descriptions.

## Validation

- `ruff check` clean; domain/instructions/command-parity/error-code pytest rows pass (33).
- Live Godot 4.7: full `test_run` **2331 passed, 0 failed, 44 skipped**; `navigation` suite **12/12**.
- Live MCP smoke: region create → `mesh_configure` enum-by-name → bake (`polygon_count 2`, `vertex_count 4`) → agent/obstacle create+configure → **`path_get` returned a real 2-point path** → unknown property refused with the settable list.

One guardrail updated: `test_dispatcher.gd`'s lazy-handler count 30 → 31.

## Summary by CodeRabbit

- **New Features**
  - Added 2D and 3D navigation authoring through the `navigation_manage` tool.
  - Create and configure navigation regions and meshes, bake navigation data, and query paths.
  - Added optional map synchronization, explicit region selection, validation, timeout handling, and undo/redo support.
  - Navigation baking now runs asynchronously with deferred result reporting.

- **Removed**
  - Removed navigation agent and obstacle creation and configuration operations.

- **Documentation**
  - Updated tool catalogs, capability counts, taxonomy, and usage details to reflect navigation support.


michaltomczykowski:
Both findings are fixed in `efc60cb`:

1. **Bake undo/redo identity** — `_do_bake` now duplicates the region's resource and bakes into the copy, so redo never mutates the resource the undo method restores. The old flow baked in place, then the first redo baked into the snapshot, and a second undo restored a baked mesh instead of the pre-bake state. `_do_bake` takes the original resource as a parameter, and the action holds an `add_undo_reference(original)` so the pre-bake resource stays alive while the action is in history. The regression test drives **bake → undo → redo → undo** and asserts the pre-bake polygon count after both undos.
2. **Enum-name gating** — enum strings are only accepted for the property that owns the vocabulary (`geometry_parsed_geometry_type` / `parsed_geometry_type`, `geometry_source_geometry_mode` / `source_geometry_mode`). A plain integer property no longer silently accepts `"both"` as `2` (`vertices_per_polygon` now reports `WRONG_TYPE`), and a bad enum name lists only that property's valid names rather than both vocabularies.

Re-validated: `ruff check` clean; navigation pytest rows pass (6); live Godot 4.7 `navigation` suite **13/13**; full `test_run` **2332 passed, 0 failed, 44 skipped**.

dsarno:
Reviewed exact head `efc60cb1`. Navigation baking and explicit-map path queries are plausible core primitives. Region/agent/obstacle creation and curated property setters mostly compose existing tools; I would narrow the proposal before adding a broad new domain.

Native Godot 4.7.2 verification found blockers in the bake path:

- `_do_bake` explicitly requests synchronous baking. A single 20×20 plane at cell size 0.05 took about 34 ms, over the 4 ms editor budget; the input work is not bounded.
- The undo action first targets the RefCounted handler, selecting global history, then mixes scene objects. The editor reports history mismatch (expected 0, got 1); scene undo can target the preceding unrelated action.
- Redo rebakes current geometry. After undo and an external source translation, redo produces different vertices instead of restoring the prior baked result.

Use a bounded deferred bake with lifecycle/cancellation checks, then commit a scene-anchored swap between retained before/after resources. Verify actual editor undo routing and exact vertices. Path queries should select a region/map explicitly; silently choosing the first region and toggling shared async-map policy needs a clearer contract. The multi-map concern is source review; the bake timing/history/redo findings were reproduced natively.

michaltomczykowski:
Thanks for the native 4.7.2 verification — all four findings are fixed at HEAD (`5a53423`):

1. **Sync bake / unbounded input** — `bake` no longer calls `bake_navigation_mesh(false)`. It bakes on the region's own background thread, replies out-of-band (`_deferred`, 30 s budget) and is driven by a static frame loop with per-frame cancellation checks (dispatcher request gone, deadline, region freed/removed, edited-scene change). Direct callers (batch_execute/tests) are refused rather than blocking the frame budget. One honest caveat: Godot 4.7 still parses source geometry on the main thread inside `bake_navigation_mesh` (engine requirement — the editor's own bake button does the same); the Recast bake that produced the 34 ms is off-thread.
2. **History routing** — the action is now scene-anchored: its only do/undo targets are `region.set_navigation_mesh(...)`, so it lands in the edited scene's history (no RefCounted handler target, no global history, no mismatch). Regression test: after bake, a scene-history undo must target the bake and a second one the preceding region-create.
3. **Redo re-bakes** — before/after resources are both retained by the action (`add_do_reference`/`add_undo_reference`) and `commit_action(false)` records the already-applied swap, so redo restores the exact baked resource. Test: bake → undo → translate the source mesh → redo → exact `get_vertices()` equality.
4. **Path queries** — `path_get` takes an explicit `region_path` (the map comes from that exact region; empty falls back to the scene root's world map — the first-region scan is gone) and an explicit `force_sync` (default `false`; a plain read never touches the shared async-iteration policy). `bake` has the same explicit `force_sync` (default `true`). A 2D query in a 3D-rooted scene must now name its region, which the tests document.

Also narrowed as you suggested: dropped `agent_create/configure` and `obstacle_create/configure` (they compose `node_create` + `node_set_property`); the rollup is now `region_create`, `mesh_configure`, `bake`, `path_get`. Tool/lazy-handler counts unchanged (47 / 31).

Re-validated on 4.7.2: `ruff check` clean; `pytest -m "not editor"` 2680 passed / 45 skipped; live full `test_run` 2350 passed / 0 failed (navigation 18/18, including scene-history routing, exact-vertex redo, abandonment/deadline restores, threaded start, force-sync restore); live dispatcher smoke `region_create → mesh_configure → path_get(region_path, force_sync)`.

michaltomczykowski:
Both findings are fixed in `e095572`:

1. **Initial step ordering (major)** — confirmed, and thanks: `bake()` stepped the job before returning the sentinel, but `_dispatch` registers the deferred request only after the handler returns. The pending check therefore saw an unregistered request, the job self-abandoned, and the call timed out with no response. The synchronous step is gone; `_drive_bake_job` now performs the first step after its registration yield (the `physics_shape`/`scene_open` deferred pattern). Added `test_bake_does_not_step_before_the_dispatcher_registers`, which asserts `bake()` performs zero pending checks before registration (its recording fixture exposes no tree, so a synchronous suite run cannot orphan a `ScriptWork` entry — the dispatcher quiescence tests caught that variant during development).
2. **Point format (minor)** — `[x,y[,z]}` → `[x,y[,z]]` in the `path_get` description.

Also updated the PR title per the title check, and added docstrings for the touched functions the coverage check flagged.

Re-validated: `ruff check` clean; `pytest -m "not editor"` 2680 passed / 45 skipped; live full `test_run` 2350 passed / 0 failed (navigation 19/19, including the new ordering test).

dsarno:
Reviewed revised head e095572d with another isolated Godot 4.7.2 editor probe. The normal bake now works asynchronously, scene-history undo targets the bake, and redo restores the identical baked resource and vertices after source geometry changes. The explicit region selector and removal of agent/obstacle wrappers are improvements.

There are still blockers:

- Two bake calls before the next frame are both accepted. One reports settled with 0 polygons while the other reports 2; Godot logs "NavigationMesh is already baking". Reserve ownership of the region/working resource before yielding.
- Actually freeing the region hits "Trying to assign invalid previously freed instance" at the typed assignment in _bake_step before its validity check. The native helper call leaves phase=start and result empty; _bake_restore has the same ordering problem. The risk of keeping the driver/work receipt pending follows from its loop and has not been reproduced through a full MCP request.
- mesh_configure(cell_size=-1,agent_radius=.7) reports both applied, but Godot rejects cell_size and only applies radius. Validate the complete request before mutation.
- Threading removes the Recast wait, but a 600x600-subdivided plane still spends 27.5 ms in the initial bake step parsing source geometry. The 30 s request deadline does not bound that main-thread call.

Resource ownership also needs guarding across waits. A direct `_bake_restore` probe overwrites a newer resource assignment with the old job's resource. This establishes the helper's behavior; the full in-flight replacement interleaving remains a source-review concern. Please cover concurrent admission, actual freed nodes, and replacement/abandonment alongside the normal undo/redo case.

My core-scope recommendation remains baking plus explicit-map queries. Region creation and curated configuration can use existing generic tools or an optional setup tool. These are native handler probes with a recording connection, not end-to-end MCP or full PR qualification; both owned editors exited normally.

michaltomczykowski:
All four blockers fixed at HEAD (`f7e75ae`), with the scope narrowed as you recommended.

**Scope.** `navigation_manage` is now `bake` + `path_get`; `region_create` and `mesh_configure` are removed (plugin handler, plugin registrations, Python handlers/description, docs, tests). Region and mesh setup composes `node_create` + `set_property`. That also retires the `mesh_configure(cell_size=-1, …)` partial-application finding.

**1. Concurrent admission.** `bake` now reserves the region in a static in-flight table before the driver's first yield — the region's own `is_baking()` only flips after a frame, which is why two same-frame calls were both accepted. A second request is refused with "already baking", and every terminal path (commit, abort, abandonment, a driver that cannot start) releases the reservation. The regression asserts the reservation is live when the driver starts and that a second call is refused while it is held.

**2. Freed region.** `_bake_step` and `_bake_restore` read the region into an untyped local, `is_instance_valid`-check it, and only then take the typed reference — the old typed assignment raised "Trying to assign invalid previously freed instance" before the check could run. The regression frees the region mid-job and asserts a clean `NODE_NOT_FOUND` abort with the reservation released.

**3. Main-thread parse.** The parse inside `bake_navigation_mesh()` cannot be preempted, so it is now measured: the start step records `parse_ms`, the result reports it, and a parse that alone exceeds the deadline aborts with `DEFERRED_TIMEOUT` instead of waiting on an overspent bake. The docs state that the deadline bounds the deferred wait, not the engine call.

**4. Resource ownership across waits.** Restore and commit only act while the region still holds this job's working duplicate. A replacement made mid-bake is left untouched, and the commit path aborts with `RESOURCE_NOT_FOUND` ("the bake result was discarded") instead of swapping in the stale bake. Regressions cover replacement + restore, replacement + commit, and replacement + dispatcher abandonment.

Validation on `f7e75ae`: `navigation` suite 19/19; full `test_run` 2351 passed / 0 failed / 31 skipped; `ruff check` clean; command-parity and server-instruction rows pass. CI is running.

michaltomczykowski:
Both CodeRabbit findings fixed at HEAD (`bf9ea13`).

**1. Restore the pre-bake resource when the driver cannot start.** Confirmed: `_begin_bake` installs the working duplicate before the deferred driver runs, and no undo action owns that replacement yet, so the two early returns in `_drive_bake_job` (freed connection, null tree) left the region holding an unbaked duplicate. Both branches now call the ownership-guarded `_bake_restore(job)` before `_release_bake(job)`.

Regressions: the null-tree driver test now also asserts the region's original mesh is back, and a new test drives the driver with a freed connection (via `_bake_job` directly — `bake`'s own null check refuses a freed connection before a job exists, which is why the defensive branch needs the direct call) and asserts the pre-bake resource is restored and the reservation released.

**2. Array form in the `path_get` description.** `[x,y[,z]}` → `[x,y[,z]]` in `src/godot_ai/tools/navigation.py`. (This was the same typo flagged earlier; the scope rewrite carried it forward — it is the only remaining occurrence; the handler comment and `json_values.gd` were already correct.)

Validation on `bf9ea13`: `navigation` suite 20/20; full `test_run` 2352 passed / 0 failed / 31 skipped; `ruff check` clean; parity/description rows pass. CI is running.

dsarno:
Maintainer verification passed on head 45cb5818. The first core version provides bounded mesh-only 3D baking and 2D/3D path queries, with explicit refusal for unsupported baking sources. It does not call Godot's general scene parser.

Local validation passed Ruff, 2,809 Python tests, 2,498 live GDScript tests, and 13 update tests (11 expected skips). Actual authenticated MCP calls verified baking, path queries, refusal cases, and editor Undo/Redo. The final source and bounded-grid revalidation received independent review. All current CI checks passed, including Windows HDR capture.

The contributor history is preserved. Merging this reviewed scope now; broader baking sources remain follow-up work.

Issue #716, opened 2026-07-13
Audit report-only notes: verify-and-fix-if-trivial leftovers (telemetry hygiene, worktree safety, fixture dedup, ack contract, screenshot encode dedup, image manifest)


dsarno:
The remaining report-only items from `docs/audit-tier2-plan.md`, recorded per the plan's "verify; fix if trivial, else note" rule. Several were already fixed in the Phase-2 PRs (`_meta_tool` union-str and `local-self-update-smoke` zip containment moved to other issues/PRs); these are the ones still standing:

- [ ] **`record_failure`/`record_latency` ship raw caller-supplied error text** (`telemetry.py:652`) — sanitize at the API boundary (rename to `error_category` + truncate 64, or require `_safe_exception_c...` routing).
- [ ] **`verify-worktree` silently `rm -rf`s a real (non-symlink) addons dir** (`script/verify-worktree:99`, S3) — move aside (`.stale.<timestamp>`) instead of deleting, with a warning.
- [ ] **Inline Python MCP client duplicated between `ci-game-capture-smoke` and `local-game-capture-diag`** (`local-game-capture-diag:117`) — minimum: wrap the two `_tool_call`s in try/except so a dead server yields a diagnosis, not a traceback.
- [ ] **`script/_dev_env.py free_port` kills any listener with no brand check** (`_dev_env.py:258`) — print each doomed PID's command line before killing; optionally brand-gate.
- [ ] **Integration fixtures drain `handshake_ack` under `except TimeoutError: pass`** (`tests/conftest.py:136` + 3 sites) — make the ack contract mandatory outside the one dedicated negative test.
- [ ] **Screenshot downscale+PNG-encode block maintained as two manually-synced copies across the editor/game boundary** (`editor_handler.gd:832`) — extract `utils/screenshot_encode.gd` shared helper.
- [ ] **`record_milestone` does synchronous file writes on the event loop** (`telemetry.py:418`) — move persistence onto the worker thread or amend the module docstring's "never blocks" claim.
- [ ] **`docs/images/` unreferenced assets** (`blockgame.png` +3) — add a manifest recording external-listing usage (store-hero.png is hotlinked by AssetLib asset 5050), then delete the truly-unreferenced ones.
- [ ] **~13 JSON descriptors carry redundant `detect_paths = PackedStringArray(path_template.values())`** (`clients/cursor.gd:12`) — delete where it exactly mirrors `path_template` and no `config_home_env` is declared.

[...]

PR #1070, opened 2026-09-18
fix(capture): correct HDR 2D game screenshot colors

### LINKED ISSUE #1058: Game screenshots appear too dark with HDR 2D enabled
### Description

With `rendering/viewport/hdr_2d=true`, screenshots returned by `editor_screenshot` (`source="game"`) look significantly darker than the actual game window.

Tested with Godot AI 4.1.0, Godot 4.7.2 Mono, Windows, and Forward+.

### Reproduction

1. Enable HDR 2D.
2. Add a ColorRect with `Color(0.5, 0.5, 0.5, 1.0)`.
3. Run the scene and capture it using `editor_screenshot`, with `source="game"`.
4. Compare the PNG with the game window.

The gray area measured RGB **(54, 54, 54)** in the MCP screenshot, versus **(127, 127, 127)** in a system capture of the window.

- The two squares on the left are from the MCP screenshot, and the two on the right are from the system screenshot.

Image

### Expected behavior

The screenshot should match the colors displayed in the game window.

### Possible cause

The screenshot code reads the viewport image and encodes it as PNG without a linear-to-sRGB conversion.

[[Godot’s documentation](https://docs.godotengine.org/en/stable/classes/class_viewporttexture.html)](https://docs.godotengine.org/en/stable/classes/class_viewporttexture.html) explains that HDR 2D viewport textures use linear encoding and may appear darker without this conversion.

Please handle the appropriate color conversion for HDR 2D captures, while leaving images already encoded as sRGB unchanged.

dsarno:
HDR 2D screenshots encoded linear colors directly as PNG, making midgray too dark. Apply shared linear-to-sRGB conversion to game and editor `viewport_2d` captures on Forward+ and Mobile; Compatibility retains its sRGB readback. Fixes #1058.

The shared encoder converts before downscaling. Its comment documents 8-bit linear quantization and clamping rather than tone mapping (midgray becomes 126 rather than 127). Smoke coverage verifies actual renderer/HDR/raw-gray metadata and native/64-pixel PNG values, preserves SDR controls, and adds separate macOS/Windows HDR rows. The integrated #1019 network deadlines remain intact; 861seconds of phase ceilings fit the 16-minute step allowance.

Validation at `737bb95656de9b49641dd3c2ed88ee797f30e5f4`, merging main `f85d2481bf6f6cef1928f20d0667fc64d454956f`:

- Full CI Ruff and diff checks passed; Python 2,746 passed / 121 skipped.
- Cold Godot 4.7.2 live suite: 2,354 passed / 0 failed / 31 skipped.
- Fresh Compatibility SDR, Forward+ SDR and Forward+ HDR editor/game captures passed at native and 64-pixel limits, plus both encoder tests in each row. HDR gray 126; SDR gray 127.
- Explicit native updater: 13 passed / 11 skipped, no receipt overlay.
- All three owned editors exited normally; final published tree matches the frozen source. Independent source/shipping review passed with notes; new-head hosted CI must finish before merge.

Initial stale-venv failures were retained and superseded by complete reruns after the repository setup script repaired that isolated environment. Three Python warnings and native negative-test/resource-cleanup diagnostics remain recorded; this is not a warning-free console claim. Backends were deliberately terminated after editor exit. Local updater fixtures do not qualify pristine published-release upgrades. The fresh renderer matrix above does not claim a new Mobile run.


plugin/addons/godot_ai/runtime/game_helper.gd
```diff
@@ -343,6 +343,11 @@ func _capture_and_reply(
 		_reply_error(request_id, "Captured an empty image from game viewport")
 		return
 
+	# Compatibility renders HDR 2D in sRGB; its float image needs no transfer conversion.
+	if viewport.use_hdr_2d and RenderingServer.get_current_rendering_method() != "gl_compatibility":
```
dsarno: The editor-side `source="viewport_2d"` capture has the same bug. `EditorInterface.get_editor_viewport_2d()` is the editor's scene-root SubViewport, and EditorNode applies the project's `rendering/viewport/hdr_2d` to it, so on an HDR-2D Forward+ project `editor_handler.gd:506` still encodes linear pixels and returns the too-dark image from #1058 while `source="game"` is now correct.

Since #716 created `McpScreenshotEncode.downscale_and_encode` precisely so the game and editor paths stay in sync, I'd move this conversion there (e.g. a `use_hdr_2d: bool` parameter, or a `normalize_hdr(image, viewport)` static) and call it from both `_capture_and_reply` and `_finalize_image`.

plugin/addons/godot_ai/runtime/game_helper.gd
```diff
@@ -343,6 +343,11 @@ func _capture_and_reply(
...
+		image.convert(Image.FORMAT_RGBA8)
```
dsarno: Worth a comment on the trade-off here: `convert(FORMAT_RGBA8)` runs *before* the gamma conversion, so linear values are quantized to 8 bits first. Consequences: the bottom ~13 sRGB levels collapse into 2 (dark UI/gradients band), anything above 1.0 clamps to white instead of being tone-mapped, and mid-gray truncates 54.6 -> 54 -> 126. That truncation is why the PR needed the +-1 tolerance rather than an exact 127.

`Image.linear_to_srgb()` only accepts RGB8/RGBA8 so the order is forced, but the reason for the 126 should live in this comment rather than be absorbed silently by the test tolerance.

plugin/addons/godot_ai/runtime/game_helper.gd
```diff
@@ -343,6 +343,11 @@ func _capture_and_reply(
 		_reply_error(request_id, "Captured an empty image from game viewport")
 		return
 
+	# Compatibility renders HDR 2D in sRGB; its float image needs no transfer conversion.
```
dsarno: Nit: this comment isn't accurate. Compatibility doesn't render HDR 2D at all; it ignores `use_hdr_2d` and the 2D buffer stays RGBA8 sRGB, so there is no "float image". Suggest: `# Compatibility ignores use_hdr_2d (RGBA8 sRGB buffer); only Forward+/Mobile hand back a linear float image.`

script/ci-game-capture-smoke
```diff
@@ -379,14 +381,18 @@ def _capture_attempt(session_id: str, request_id: int) -> tuple[list[str], bytes
         y = int(img.height * fy)
         actual = img.getpixel((x, y))
         samples.append(actual)
-        match = _within_tolerance(actual, expected)
+        match = (
```
dsarno: Two things here:

1. The `label == "gray"` special case hard-codes a second tolerance by name. A per-sample tolerance in the `QUADRANTS` tuple (or a `tolerance=` parameter on `_within_tolerance`) removes the branch and the next precise sample won't need another one.
2. The generator variables `x, y` shadow the pixel coordinates `x, y` used in the `print`/failure string two lines below. It's not a bug (genexpr scope), but it reads like one.

```suggestion
        match = _within_tolerance(actual, expected, tolerance)
```
(with `tolerance` pulled from a 5th tuple field and `_within_tolerance` taking it as an argument)

test_project/capture_smoke.gd
```diff
@@ -0,0 +1,5 @@
+extends Control
+
+
+func _ready() -> void:
+	get_viewport().use_hdr_2d = true
```
dsarno: Coverage note: forcing HDR on for every row means the default non-HDR Forward+/Mobile capture path (what most users hit) is no longer exercised by any smoke row, and the Linux row (opengl3) can't exercise the fix at all. A future change that applies `linear_to_srgb` unconditionally, or inverts the renderer check, would double-convert non-HDR captures and nothing would catch it.

Cheapest fix: gate this on an env/cmdline flag the smoke passes (e.g. `CAPTURE_HDR_2D=1`) and have the macOS/Windows rows run the capture loop once with it and once without, or add a second scene and have `_capture_attempt` run both.

dsarno (commented):
Reviewed the HDR conversion and the smoke changes. No blocker; one same-bug-elsewhere item and a few test/coverage notes inline.

plugin/addons/godot_ai/runtime/game_helper.gd
dsarno: Addressed in 7df4aa1f0bfbc0c74aa7dc5bd3543d4b32e4a739. Moved the conversion into McpScreenshotEncode.downscale_and_encode(image, max_resolution, use_hdr_2d=false). Both game capture and editor viewport_2d pass their actual viewport flag; other editor sources retain the default. The original editor path reproduced gray 54 with HDR versus 127 without; the corrected public editor_screenshot returns 126 within the documented one-level tolerance, at native and reduced resolution.

plugin/addons/godot_ai/runtime/game_helper.gd
dsarno: Addressed in 7df4aa1f0bfbc0c74aa7dc5bd3543d4b32e4a739. The shared encoder comment now documents why RGBA8 conversion precedes linear_to_srgb, the quantization of linear darks, clamping above 1.0 instead of tone mapping, and the 54.6 -> 54 -> 126 midgray truncation. The one-level gray tolerance follows that measured behavior.

plugin/addons/godot_ai/runtime/game_helper.gd
dsarno: Addressed in 7df4aa1f0bfbc0c74aa7dc5bd3543d4b32e4a739. Reworded the comment to the measured contract: Compatibility readback stays sRGB, including with HDR 2D enabled. I disagree with the proposed always-RGBA8 wording. On official Godot 4.7.2, an actual gl_compatibility run reported use_hdr_2d=true, Image format 11 (FORMAT_RGBAF), raw gray 0.5 and encoded gray 127 for both editor and game. The no-conversion decision is correct; assuming an 8-bit readback is not. The revised comment avoids a buffer-format claim.

script/ci-game-capture-smoke
dsarno: Addressed in 7df4aa1f0bfbc0c74aa7dc5bd3543d4b32e4a739. Each QUADRANTS entry now supplies its tolerance, and _within_tolerance accepts it directly. Primary samples retain 40; gray uses 1. Generator variables are channel/target, separate from sampled pixel coordinates.

test_project/capture_smoke.gd
dsarno: Addressed in 7df4aa1f0bfbc0c74aa7dc5bd3543d4b32e4a739. CAPTURE_HDR_2D now reaches the editor before launch. Existing Linux/macOS/Windows jobs retain SDR coverage, with additional HDR 2D jobs on macOS/Windows. The smoke reads back actual editor/game HDR and renderer state, rejects a Compatibility fallback in an HDR row, and checks public editor/game PNGs at native and 64px resolutions. Local Forward+ HDR/SDR, Mobile HDR and Compatibility SDR controls all passed with normal editor exits.

dsarno:
Addressed the five review threads in 7df4aa1f0bfbc0c74aa7dc5bd3543d4b32e4a739. Editor and game 2D captures now share the HDR conversion. SDR coverage remains in the original CI jobs; macOS and Windows gain explicit HDR rows. Public screenshot smoke verifies the actual renderer, HDR state, raw gray patch, and native/downscaled encoded pixels.

Validation passed:
- CI-scope Ruff.
- Full Python suite: 2,677 passed, 115 skipped.
- Cold live GDScript suite: 2,334 passed, 31 skipped.
- Rendered public-tool smoke on Forward+ HDR and SDR, Mobile HDR, and Compatibility SDR, with normal editor exits.
- Independent source review and visual inspection of corrected editor captures.

Original editor HDR captures encoded gray 54; corrected captures encode 126 within the documented one-level quantization tolerance. Compatibility is left in sRGB. Its actual Godot 4.7.2 HDR readback was floating point, so the corresponding thread explains why the proposed always-RGBA8 comment was not adopted.

A separate new native test fixture was rejected after intermittent shutdown access violations reproduced with both original and changed code. Its cause remains unconfirmed. The existing live MCP capture smoke now carries the regression coverage; all four final smoke editors exited normally. Failed fixture evidence was retained locally, not suppressed or shipped.

Each thread has an individual reply. Leaving the PR unmerged for maintainer review; combined stack verification is still pending.

dsarno:
Follow-up 88f9100871f465104be15a54fac4113d8a8eb369 fixes the capture job's fresh-editor setup. All five failed jobs reached an uninitialized 2x2 editor viewport. Selecting the 2D screen in the scene's _ready could be overwritten by scene activation; the fixture now defers that selection. Failure metadata also includes the actual image and viewport dimensions.

A fresh local project reproduced the CI failure, and the deferred-selection-only fix passed three fresh-project captures with normal exits. The initial small-patch geometry hypothesis was rejected. No product code or pixel tolerance changed.

Repeated gates passed: complete CI Ruff, 2,677 Python tests, and 2,334 live GDScript tests. Hosted CI is running again; this comment does not claim it has passed.

dsarno:
Follow-up in af7045675ad005f434e7315dd82ec4fdc7f70385 fixes the remaining hosted smoke failure. The viewport was now initialized, but CI rendered 649×510 / 649×533 / 850×595 while the gray sample remained at (1126, 651). The editor-only fixture now places that patch inside the visible area; `autosave=False` preserves the game's original on-disk quadrants.

Also addressed the outside-diff Compatibility coverage comment: every capture row now runs `screenshot_encode`, rejects failures/skips/zero passes, and prints failure details. Linux explicitly selects `gl_compatibility`.

Validation: matched 649×510 RED→GREEN with strict raw/color checks and native/64px captures; Compatibility encoder suite 2 passed, 0 skipped; Ruff clean; Python 2,677 passed / 115 skipped; cold live GDScript 2,334 passed / 0 failed / 31 skipped. Owned editors exited normally. Hosted CI is queued for this commit.

dsarno:
Follow-up in 75b5be0d722842bd052449765a55ec7096b727d3 corrects the smoke harness's exact-size assumption: the encoder promises a maximum edge, and 850x595 legitimately becomes 63x44 at cap64 after floating-point scaling/truncation. Both editor/game checks now accept only63..64px, with positive dimensions.

Game rendering diagnostics now retain the full response. Only structured EVAL_GAME_NOT_READY retries within the existing total capture budget; malformed/permanent errors, wrong HDR or renderer, and late responses fail. The previous macOS run discarded its error payload, so its original cause remains unconfirmed. The encoder suite must report exactly two passes without failures, skips or load errors.

Verified actual editor AND game captures at850x595,649x510,649x533, native and capped, plus a disposable helper that deliberately delayed its first liveness response: observed EVAL_GAME_NOT_READY followed by successful rendering proof. All owned editors exited normally.22 focused tests, full Ruff, Python2699 passed/115 skipped, and cold live GDS2334 passed/0 failed/31 skipped passed before commit. Fresh hosted CI remains the cross-platform check.

script/ci-game-capture-smoke
```diff
@@ -408,6 +412,85 @@ def _write_diag_artifacts(attempt_log: list[str], last_png: bytes | None) -> Non
...
+def _check_game_rendering(session_id: str, history: list[str], deadline: float) -> None:
+    while True:
+        if time.monotonic() >= deadline:
+            raise RuntimeError("Game rendering probe exceeded the capture budget")
+        response = _tool_call(session_id, "editor_manage", {
+            "op": "game_eval", "params": {"code": (
+                'return {"hdr": get_tree().root.use_hdr_2d, '
+                '"renderer": RenderingServer.get_current_rendering_method()}'
+            )},
+        }, 12)
```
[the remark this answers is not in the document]
dsarno: The fourth positional argument is request_id, so 12 is the JSON-RPC ID, not a 12-second timeout. This call uses the 30-second default. Reducing it to 12 seconds would also put it below the game's 15-second command timeout.

The narrower observation is valid: an already-running HTTP request can finish beyond the nominal capture deadline. The smoke uses a shared retry deadline plus finite per-call timeouts, as the existing screenshot loop does. The new probe checks the deadline before issuing a request and again after its reply; a late success fails, and the regression test covers that case. It cannot reset the budget or pass after the deadline.

Keeping this as a nonblocking harness timing limitation rather than applying the proposed timeout change. It does not weaken renderer/color checks or allow a false pass. The final hosted macOS SDR/HDR rows now retain the explicit EVAL_GAME_NOT_READY response, then succeed on the next probe; all five capture rows pass at 75b5be0d.

dsarno:
Independent shipping review: **PASS+NOTES**, pending successful new-head hosted CI.

Head `737bb95656de9b49641dd3c2ed88ee797f30e5f4`; base `f85d2481bf6f6cef1928f20d0667fc64d454956f`; stable patch-id `d0a89cad3e1e1cdb4c21026e8374e74732ab5dd4`. Published tree `83daeb5ba1bd18ea553363132b23b7dd405ee25c` exactly matches the frozen tested index; all nine recorded file hashes agree. The two smoke conflict resolutions are the previously reviewed #1019/#1070 combination. Shared conversion, HDR/SDR rows and the 861-second/16-minute budget remain correct; the stale arithmetic comment was fixed.

Reviewed fresh qualification receipts: Ruff, Python 2,746 passes, cold live GDScript 2,354 passes / zero failures, three actual editor/game capture rows (Compatibility SDR, Forward+ SDR/HDR) at native and 64-pixel sizes, and native updater 13 passes. No receipt overlay; normal editor exits and owned-process cleanup recorded. These gates were run by the verification owner; this independent pass reviewed source and evidence rather than duplicating them.

The repaired stale local environment's first failures, three Python warnings and native negative-fixture/resource cleanup diagnostics remain retained. Backend termination was explicit cleanup. Local update fixtures are not pristine published-release qualification; conversion still quantizes linear darks and clamps values above 1. No additional Mobile runtime run is claimed in this refresh.

GitHub currently reports mergeable with hosted checks still running; CodeRabbit succeeds and no new inline findings were present. Require successful checks at this identity before landing. No merge performed by this reviewer.

PR #1071, opened 2026-09-18
fix(windows): detect wildcard listeners before choosing server ports

### LINKED ISSUE #1054: --allow-host on Windows permanently stalls the readiness proof: server is healthy and reachable, but the dock stays amber and the editor never connects
### Summary

On Windows11, enabling **Allow Remote Hosts** (`--allow-host`) makes the managed
server permanently unusable from the plugin's point of view:

- the dock stays amber (`Starting server…`) and then fails with
  `The launched server process exited or changed identity before publishing capabilities.`
  (`Last pending proof: authenticated_status`), or earlier with
  `The managed server proof timed out at capability_record.`
- **the server itself is completely healthy** — it binds the LAN address and answers
  authenticated requests correctly from a remote host the whole time.
- clearing **Allow Remote Hosts** makes the dock go green again immediately
  (same machine, same project, no other change). Adding it back breaks it again, 100% reproducible.

So the LAN opt-in (#421) is currently unusable on Windows for me, even though the
networking part works perfectly.

### Environment

| | |
|---|---|
| Godot | 4.7.2-stable (Steam) |
| Godot AI plugin | 4.1.0 |
| OS | Windows11, x64 |
| Editor Setting `godot_ai/allow_remote_hosts` | `192.168.0.0/24` |
| Effective endpoint ports (`godot_ai/v4_endpoint_ports`) | http `8001`, ws `8002` |

(`HTTP 端口` / `Ws Port` in Editor Settings still show the legacy `8000` / `9500`;
the `v4_endpoint_ports` dictionary is what actually takes effect — the spawn argv
confirms `--port 8001 --ws-port 8002`.)

### Network layout (real setup, the bug is not network-related)
Godot host (Windows) MCP client host 192.168.0.30 Docker host 192.168.0.29 ┌──────────────────────────────┐ ┌───────────────────────────────────┐ │ Godot editor + Godot AI │ │ container 172.18.0.5 │ │ │ ws 127.0.0.1:8002 │ │ MCP client (streamable-http) │ │ ▼ │ │ │ │ godot-ai backend │◄──────────┼── connects to 192.168.0.30:8001 │ │ (plugin-spawned, uvx tier) │ LAN │ │ └──────────────────────────────┘ └───────────────────────────────────┘

The client host and the Godot host are different machines on the same LAN; the
client reaches the Godot host over the LAN with no port forwarding involved.

### Steps to reproduce

1. Godot 4.7.2 + Godot AI 4.1.0 on Windows, project open, plugin enabled.
2. Editor Settings → **Allow Remote Hosts** → leave **empty**.
3. `Reload Plugin` → dock goes **green**, editor session connects, everything works.
4. Editor Settings → **Allow Remote Hosts** → `192.168.0.0/24`.
5. `Reload Plugin` → dock goes **amber** and never recovers.

### Evidence that the server is fine

While the dock is amber, from the other machine on the LAN:
$ curl -i http://192.168.0.30:8001/godot-ai/status HTTP/1.1 401 Unauthorized server: uvicorn www-authenticate: Bearer realm="godot-ai"

$ curl -H "Authorization: Bearer <http capability>" http://192.168.0.30:8001/godot-ai/status {"name":"godot-ai","server_version":"4.1.0","ws_port":8002,"tool_surface":"rollup", "instance_id":"<instance_nonce from http-8001.json>","owner_type":"plugin", ...}

- the status probe returns 200 with the correct bearer token,
- `instance_id` is **equal** to `instance_nonce` in
  `%LOCALAPPDATA%\godot-ai\capabilities\http-8001.json`,
- `name == "godot-ai"`, `ws_port == 8002`, `server_version == 4.1.0`,
- the process chain is alive: `uvx.exe → uv.exe → godot-ai.exe → python.exe`.

i.e. every condition of `_authenticated_status_matches_record()` is satisfied when
checked from outside, yet the plugin reports that proof as pending.

### Root cause

Two pieces of shipped code combine badly:

1. With an IPv4 network in the allowlist, the server binds the **wildcard** address
   (`src/godot_ai/transport/origin_guard.py:123`, `bind_host_for_networks`):

   ```python
   if any(isinstance(net, ipaddress.IPv4Network) for net in networks):
       return "0.0.0.0"
On Windows the readiness probe short-circuits on a loopback bind test (plugin/addons/godot_ai/utils/server_lifecycle.gd:1295):


## Windows can otherwise spend the whole connect deadline on an unbound
## loopback port. This is only an unreachable probe; callers still check
## occupancy and prove any later listener before using it.
if OS.get_name() == "Windows" and PortResolver.can_bind_local_port(port):
    result.error = "port_unbound"
    return result          # the HTTP probe below never runs
and can_bind_local_port tests by binding 127.0.0.1 (plugin/addons/godot_ai/utils/port_resolver.gd:45):


static func can_bind_local_port(port: int) -> bool:
    var server := TCPServer.new()
    var err := server.listen(port, "127.0.0.1")
    if err == OK:
        server.stop()
        return true
    return false
With --allow-host the server holds 0.0.0.0:8001, and a 127.0.0.1:8001 bind probe still succeeds, so the fast path concludes "nobody is listening" and returns port_unbound without ever sending the request. Without --allow-host the server holds 127.0.0.1:8001, the same probe fails, and the probe proceeds correctly — which is exactly why the bug only appears once the LAN opt-in is set.

The authors are clearly aware of this coexistence case; is_port_in_use() handles it (port_resolver.gd:60):


if can_bind_local_port(port):
    ## On POSIX, an IPv6 wildcard listener can coexist with a
    ## successful 127.0.0.1 bind probe. ...
    if OS.get_name() != "Windows":
        return is_port_in_use_via_scrape(port)
    return false
The probe fast path in server_lifecycle.gd:1295 is missing the same fallback.

Knock-on effect (why it ends as "process exited")
authenticated_status can therefore never pass → the plugin never emits transport_ready → the editor never opens the WebSocket → zero sessions → orphan_reaper's boot grace expires and the backend terminates itself:

src/godot_ai/orphan_reaper.py: DEFAULT_IDLE_GRACE_SECONDS = 120.0
plugin/addons/godot_ai/utils/server_lifecycle.gd: DEFAULT_PROVE_TIMEOUT_MS = 180_000
120 s < 180 s, so the backend always dies first and the plugin then reports The launched server process exited or changed identity before publishing capabilities. GODOT_AI_NO_IDLE_EXIT / Keep Server on Exit does not help — the proof is already stuck, it only keeps the corpse alive.

Suggested fix
Give the probe fast path the same occupancy confirmation the rest of the file already uses, so it only short-circuits when the port is definitely free:


if (
    OS.get_name() == "Windows"
    and PortResolver.can_bind_local_port(port)
    and not PortResolver.is_port_in_use_via_scrape(port)
):
    result.error = "port_unbound"
    return result
is_port_in_use_via_scrape() finds the wildcard listener via netstat -ano, so the probe proceeds and the proof completes. The netstat cost is only paid in the ambiguous case (bind probe succeeded), and the loopback case short-circuits on the first condition as before.

Confirmed locally: with exactly this one-line change, --allow-host 192.168.0.0/24 now works — the dock stays green, the editor connects, and remote clients can reach http://<godot-host>:8001/mcp directly.

Workarounds (for other users hitting this)
Keep Allow Remote Hosts empty and expose the loopback-only server some other way (e.g. netsh interface portproxy + a TCP relay that presents a loopback Host header, which the origin guard requires when no allowlist is set), or
apply the patch above.

dsarno:
## Why

On Windows, a successful loopback bind can coexist with an existing wildcard listener. The editor therefore reported an occupied backend port as free and skipped a working authenticated endpoint. Fixes #1054.

## Scope

Use a shared Windows listener snapshot with explicit free, occupied, and unknown states. Startup probes uncertain endpoints; port selection and shutdown remain conservative when discovery fails. Reuse snapshots within a port-selection operation. Authentication and process ownership checks are preserved.

## Verification

- Ruff passed; Python suite: 2,677 passed, 118 skipped, with editor checks run separately.
- Native Windows occupancy and migration regressions: 8 passed, including the real wildcard-bound authenticated backend.
- Live Godot: 2,338 passed, 31 skipped; three reload stress iterations passed.
- 3.2.4 migration and automatic signed-update tests passed.
- Visible local-self-update smoke passed: actual Update/confirmation flow, same editor activation, backend replacement, authenticated new version. Used the local signed fixture, Compatibility renderer, and isolated settings.
- Independent review passed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Windows port detection using listener information to reduce incorrect free-port suggestions.
  - Prevented upgrades and connection probes from proceeding when Windows port discovery is unavailable.
  - Rechecked selected ports before applying configuration changes to avoid conflicts.
  - Preserved existing port selections when no safe replacement is available.
  - Managed server restarts now continue when port status cannot be determined, while still blocking confirmed conflicts.

- **User Feedback**
  - Added clearer retry or manual-selection guidance when discovery fails.
  - Port picker messaging now explains unavailable suggestions and discovery issues.
  - Status displays now distinguish unavailable port discovery from confirmed port conflicts.


plugin/addons/godot_ai/utils/server_lifecycle.gd
```diff
@@ -719,7 +719,16 @@ func _effect_probe(payload: Dictionary) -> Dictionary:
 				"transport": _transport_from(port, ws_port, live, capability),
 			}
 		return _blocked_probe_result("incompatible", port, live, true)
-	if PortResolver.is_port_in_use(port):
+	var occupied: bool
+	if OS.get_name() == "Windows":
+		var occupancy := PortResolver.windows_port_occupancy(port)
+		if occupancy == PortResolver.PortOccupancy.UNKNOWN:
+			return _blocked_probe_result("port_occupancy_unknown", port, live, false,
```
dsarno: `port_occupancy_unknown` isn't in `_dock_state`'s `match reason`, so it falls through to `ServerState.CRASHED`. Combined with `_blocked_probe_result` always prefixing `"Port %d is occupied by another process (%s)."`, the user sees the crash panel saying the port is occupied by another process, when nothing was launched and nothing is known to be on the port.

Suggest: add the reason to `_dock_state` (FOREIGN_PORT keeps the picker + Restart affordances, or a dedicated state), and let `_blocked_probe_result` take the full message for this reason instead of the "occupied (detail)" template. The same wording problem applies to `restart_or_start_managed_server`, which only `push_warning`s and returns false with no dock feedback.

plugin/addons/godot_ai/utils/server_lifecycle.gd
```diff
@@ -719,7 +719,16 @@ func _effect_probe(payload: Dictionary) -> Dictionary:
 				"transport": _transport_from(port, ws_port, live, capability),
 			}
 		return _blocked_probe_result("incompatible", port, live, true)
-	if PortResolver.is_port_in_use(port):
+	var occupied: bool
+	if OS.get_name() == "Windows":
+		var occupancy := PortResolver.windows_port_occupancy(port)
```
dsarno: This probe now takes three separate listener snapshots on Windows: one inside `_probe_with_capability` (line ~1306), this one, and `is_port_in_use(expected_ws_port)` at line ~742. The readiness poll (`_effect_prove`, line ~953/985) repeats the first one on every tick until the server binds. At ~30 ms per `netstat` that's tolerable; when netstat is unusable each of them becomes a ~1.2 s `powershell.exe` run, and on UNKNOWN `_probe_with_capability` additionally falls through to the full connect deadline that the `port_unbound` shortcut was added to avoid.

Taking one `windows_listener_snapshot()` at the top of `_effect_probe` and threading it through (`_probe_with_capability(port, capability, timeout_ms, snapshot)`, `windows_port_occupancy(port, snapshot)`, `windows_port_occupancy(expected_ws_port, snapshot)`) fixes the cost and also makes the three decisions consistent with one listener table instead of three.

plugin/addons/godot_ai/utils/port_resolver.gd
```diff
@@ -51,18 +53,11 @@ static func can_bind_local_port(port: int) -> bool:
...
 static func is_port_in_use(port: int) -> bool:
-	if can_bind_local_port(port):
-		## On POSIX, an IPv6 wildcard listener can coexist with a
-		## successful 127.0.0.1 bind probe. Confirm with lsof so startup
-		## sees the same listener set that shutdown/recovery would see.
-		if OS.get_name() != "Windows":
-			return is_port_in_use_via_scrape(port)
-		return false
+	if OS.get_name() == "Windows":
+		return windows_port_occupancy(port) != PortOccupancy.FREE
```
dsarno, line 60: Returning `true` for UNKNOWN is the right conservative default for a one-shot check, but `wait_for_port_free` (line ~783) loops on this every 100 ms for its 3-5 s timeout, and each iteration now spawns netstat plus, when netstat is unusable, a ~1.2 s PowerShell run. Waiting cannot turn UNKNOWN into FREE, so on a host with broken netstat every stop/replace burns the whole timeout in back-to-back `powershell.exe` spawns and then reports `process_stop_failed` / `ok: false` even though the process was killed.

Suggest `wait_for_port_free` call `windows_port_occupancy` directly and return immediately on UNKNOWN (the callers already treat the result as unproven).

plugin/addons/godot_ai/utils/port_resolver.gd
```diff
@@ -102,6 +83,76 @@ static func is_port_in_use_via_scrape(port: int, trace: Callable = Callable()) -
...
+
+static func windows_port_occupancy(port: int, snapshot: Dictionary = {}) -> PortOccupancy:
+	if snapshot.is_empty():
+		snapshot = windows_listener_snapshot()
+	if not snapshot.known:
+		return PortOccupancy.UNKNOWN
+	return PortOccupancy.OCCUPIED if snapshot.ports.has(port) else PortOccupancy.FREE
+
+
+static func windows_snapshot_from_netstat(exit_code: int, output: Array) -> Dictionary:
```
dsarno, line 110: This is now the third parser over the same `netstat -ano` dump: `parse_windows_netstat_pids` already applies the same `remote ends_with(":0")` listener heuristic (with the locale rationale in its comment), and `windows_netstat_dump_parseable` already does the row-shape check. Occupancy and PID ownership are therefore decided by two different parsers that can disagree on the same row.

If the snapshot recorded `port -> [pids]` instead of a flat port list, `find_all_pids_on_port` could consume it too, the readiness loop could share the snapshot, and `parse_windows_netstat_listening` (line ~344), which has no callers after this PR, could go.

dsarno (commented):
Reviewed the tri-state occupancy change end to end (resolver, lifecycle probe, dock, configurator, tests). The wildcard-listener fix itself looks right. Main items: the new UNKNOWN reason renders as a crash with 'occupied' wording, and UNKNOWN turns several polling paths into multi-second PowerShell loops. Details inline.

plugin/addons/godot_ai/utils/port_resolver.gd
```diff
@@ -102,6 +83,76 @@ static func is_port_in_use_via_scrape(port: int, trace: Callable = Callable()) -
...
+		var port_text := fields[1].get_slice(":", fields[1].get_slice_count(":") - 1)
+		var remote_port_text := fields[2].get_slice(":", fields[2].get_slice_count(":") - 1)
+		var pid_text := fields[fields.size() - 1]
+		if not port_text.is_valid_int() or not remote_port_text.is_valid_int() or not pid_text.is_valid_int():
+			return unknown
+		var port := int(port_text)
+		if port < 1 or port > 65535 or int(pid_text) < 0 or int(remote_port_text) < 0 or int(remote_port_text) > 65535:
+			return unknown
+		if fields[2].ends_with(":0"):
+			if int(pid_text) == 0:
```
dsarno: A listener row with a valid local port and PID 0 invalidates the *whole* snapshot. The port is known here, so the conservative answer is to mark that port OCCUPIED, not to discard the table. As written, one such row on a host (kernel-owned/transitional rows do show PID 0) sends every occupancy query on that machine to the 1.2 s PowerShell fallback, and if that also fails, server start is blocked with `port_occupancy_unknown` and no override.

```suggestion
		if fields[2].ends_with(":0"):
			if not ports.has(port):
				ports.append(port)
```

plugin/addons/godot_ai/dock_panels/port_picker_panel.gd
```diff
@@ -80,17 +80,40 @@ static func _port_spinbox(value: int, tooltip: String) -> SpinBox:
...
+		occupancy = PortResolver.windows_listener_snapshot()
 	var http := ClientConfigurator.http_port()
 	var ws := ClientConfigurator.ws_port()
-	if conflict_port == http or bool(port_in_use_probe.call(http)):
-		http = ClientConfigurator.suggest_free_port(http + 1)
-	if conflict_port == ws or ws == http or bool(port_in_use_probe.call(ws)):
-		ws = ClientConfigurator.suggest_free_port(ws + 1)
-		if ws == http:
-			ws = ClientConfigurator.suggest_free_port(ws + 1)
+	var use_snapshot := OS.get_name() == "Windows" and port_in_use_probe == Callable(PortResolver, "is_port_in_use")
```
dsarno: Comparing `port_in_use_probe` against `Callable(PortResolver, "is_port_in_use")` to decide whether the seam is "real" couples the panel to the default's identity: a test that injects a wrapper around the real probe, or a future default, silently takes the other branch on Windows. Simpler: on Windows always derive `http_occupied`/`ws_occupied` from `occupancy` (the test already injects that), and keep `port_in_use_probe` for the non-Windows path only, or change the seam to return a `PortOccupancy`.

plugin/addons/godot_ai/client_configurator.gd
```diff
@@ -140,17 +143,30 @@ static func prepare_major_upgrade_endpoints(from_version: String, to_version: St
...
+				selected.append(candidate)
+				if selected.size() == 2:
+					break
 		candidate += 1
 		if candidate > MAX_PORT:
 			candidate = MIN_PORT
 	if selected.size() != 2:
 		return {"ok": false, "error": "No independent HTTP/WebSocket port pair was available for this major upgrade."}
+	if OS.get_name() == "Windows":
+		occupancy = PortResolver.windows_listener_snapshot()
```
dsarno, line 162: If this second snapshot comes back UNKNOWN, `available` is false and the user gets "The selected upgrade port %d became unavailable; retry endpoint selection.", which tells them the port was taken. The first check (line ~136) reports the discovery failure correctly; mirror that here so a retry loop on broken netstat says why it keeps failing.

plugin/addons/godot_ai/utils/server_lifecycle.gd
dsarno: Fixed in 07017fad. UNKNOWN now maps to the recoverable port panel with its own message and the dock headline "Windows port discovery unavailable". Restart delegates to the lifecycle owner, preserving pending episodes while making failure visible. Native regression and rendered startup/reload checks exercise the actual dock.

plugin/addons/godot_ai/utils/server_lifecycle.gd
dsarno: Fixed in 07017fad. The initial probe shares one listener snapshot across authenticated preflight, HTTP occupancy, and WebSocket occupancy; readiness proof also shares its initial observation. Final process authority still requires a fresh listener observation after the final authenticated probe. I retained authenticated probing on UNKNOWN: a real wildcard-listener regression confirms a reachable valid server can still authenticate when listener discovery fails, so UNKNOWN cannot safely mean unreachable.

plugin/addons/godot_ai/utils/port_resolver.gd
dsarno: Fixed in 07017fad. wait_for_port_free now returns PortOccupancy and stops immediately on UNKNOWN. Stop/replacement callers preserve an unproven result instead of declaring the port free. The native regression checks that UNKNOWN performs one discovery attempt rather than spinning until the timeout.

plugin/addons/godot_ai/utils/port_resolver.gd
dsarno: Fixed in 07017fad. The canonical snapshot now stores port -> positive PIDs. Occupancy, PID lookup, and parseability use the same parser; readiness shares that snapshot. Removed the dead listening parser and obsolete PowerShell PID-only helpers. Existing compatibility wrappers delegate to the canonical parser, and the published class_name is unchanged.

plugin/addons/godot_ai/utils/port_resolver.gd
dsarno: Fixed in 07017fad. PID 0 records the port as occupied with an empty PID list. It no longer discards a valid listener table and never grants authority to stop a process. Both netstat and PowerShell parser tests cover this distinction.

plugin/addons/godot_ai/dock_panels/port_picker_panel.gd
dsarno: Fixed in 07017fad. Windows always derives occupancy from the supplied snapshot; the Boolean callable remains only for the non-Windows path. Native coverage injects a conflicting callable and verifies that the snapshot controls the Windows suggestion.

plugin/addons/godot_ai/client_configurator.gd
dsarno: Fixed in 07017fad. An UNKNOWN revalidation now reports "Windows could not query listening ports; retry endpoint selection." The native regression checks both that message and unchanged endpoint settings.

plugin/addons/godot_ai/utils/server_lifecycle.gd
```diff
@@ -719,7 +719,16 @@ func _effect_probe(payload: Dictionary) -> Dictionary:
...
+		if occupancy == PortResolver.PortOccupancy.UNKNOWN:
+			return _blocked_probe_result("port_occupancy_unknown", port, live, false,
+				"Windows could not query listening ports; retry when port discovery is available")
+		occupied = occupancy == PortResolver.PortOccupancy.OCCUPIED
```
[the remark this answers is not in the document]
dsarno: Fixed in 07017fad. HTTP and WebSocket occupancy now use the same Windows snapshot. UNKNOWN produces the discovery-unavailable result before the WebSocket conflict check, preventing a later failed scrape from appearing as ws_occupied.

dsarno:
Addressed the seven maintainer comments and the matching WebSocket discovery comment in 07017fad. Each thread now has an individual reply; none was silently resolved.

Verification passed: full CI Ruff scope; 2,677 Python tests; 2,341 live GDScript tests with zero failures; all four native Windows occupancy rows; three isolated plugin reloads; and rendered UNKNOWN startup/retry checks. The signed local update fixture preserved the same editor, dirty scene, undo, client configuration, and original attach bridge.

The update check briefly showed a red launch-unproven diagnostic before automatic recovery. Comparing the original combined code reproduced the same old-attach/new-backend collision. History confirms that failures without proven process identity intentionally remain visible outside the narrow amber handoff case. No authority bypass or new functional regression was demonstrated. This was a locally signed current-source fixture, not untouched published-release qualification.

Separate follow-up found during verification: the existing post-update handoff test uses await despite the synchronous GDScript runner, so its post-timer assertions are not included in its reported result. The new tests in this change do not rely on that coroutine. I am keeping that coverage correction separate from the listener fix.

Leaving this PR unmerged for maintainer review. The combined final validation will include the remaining review fixes and selected dependency updates.

dsarno:
Added c0960bda so the native Windows listener regression runs in the existing Windows editor step. Without this explicit invocation, the ordinary Python jobs skip it because GODOT_BIN is unset.

Verification passed before this CI-only commit: full CI-scope Ruff; full Python suite (2,677 passed, 119 skipped); cold live Godot suite (2,341 passed, 31 skipped); and the exact newly wired native invocation (6 passed). Both owned editors exited normally. Hosted CI is running on the new head.

dsarno:
Independent shipping refresh: **PASS + NOTES** for head `c0960bdaf51dc742f07221608b9c240e8cdd239d`, base `24859943117a9390ec9f924ea7da3adf9fd39866`; stable patch ID `e2b9bab263386f8fb44dade3a8d4e1f93c669abb`. I reviewed the design/source and evidence without implementing this PR's production changes.

The final production patch retains shared Windows listener snapshots, explicit UNKNOWN occupancy, occupied PID-zero rows without fabricated process authority, and fresh ownership checks before authority is granted. The last commit only wires the existing Windows native regression into CI. Retained verification covers Ruff, 2,677 Python passes, 2,341 live GDScript passes, the exact six-test native CI invocation, rendered UNKNOWN startup/retry, reloads and signed local update continuity. All 23 current check/status results succeed; the PR is clean and mergeable.

The outside-diff review note is a valid diagnostic improvement: some low-level PID/ownership paths still express unknown discovery as generic pending/unproven results. They remain fail-closed; no new process authority or false-free result follows. This is nonblocking for the verified wildcard-listener correction. The local update fixture briefly displayed the already-recorded launch-unproven diagnostic before recovery; this evidence is not untouched published-release qualification.

Ready for the coordinator's exact-patch merge check; no merge performed by this review.

PR #1073, opened 2026-09-18
test: stabilize self-update identity receipt handshake

dsarno:
The self-update test harness could fail on Windows when its receipt reader raced the editor's repeated writes. Retry only missing-file and permission/sharing failures within the existing wait, publish the pre-update receipt once, and retain nonce matching. Permanent I/O errors still fail immediately; timeout messages preserve sanitized read-error and nonce-mismatch context.

Check stop/deadline before discovery and receipt reads, and refuse late matching results before opening the client. Native tests cover an exclusive Windows file handle and receipt stability while the editor remains alive at the gate. CI runs the native receipt regression.

Validation: six deadline regressions fail against the previous loop; all 23 focused harness tests and full Ruff pass. Fresh live Godot: 2,352 passed, zero failed, 31 skipped, normal editor exit. Full Python: 2,758 passed/122 skipped; native updater, receipt, client queue and handoff checks: 16 passed/11 skipped. Recorded Python warnings and expected negative-suite/editor teardown diagnostics remain disclosed.

This changes test infrastructure only. Deadline checks do not forcibly interrupt a synchronous filesystem operation already in progress; they prevent subsequent work and late success. Local signed fixtures are not pristine published-release qualification.

Reconciled with main `f85d2481bf6f6cef1928f20d0667fc64d454956f`, retaining all native CI rows. All listed gates were freshly rerun on the reconciled tree.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved self-update receipt handling, including clearer timeout diagnostics and reliable retry behavior.
  - Update failures now report distinct errors when the receipt cannot be opened or written.
  - Prevented stale, mismatched, expired, or late receipts from triggering an agent connection.

- **Tests**
  - Added coverage validating receipt stability during updates and expanded editor-harness scenarios.
  - Added the receipt integration scenario to Linux continuous integration testing.


tests/integration/test_self_update_receipt.py
```diff
@@ -0,0 +1,74 @@
...
+        source, count=1, flags=re.DOTALL,
+    )
+    assert replaced == 1
+    support.write_text(source, encoding="utf-8")
+    (project / "observer.gd").write_text('''@tool
+extends Node
+var observed := 0
+func _process(_delta: float) -> void:
+    if FileAccess.file_exists("res://_test_pre_instance_id.txt"):
+        observed += 1
```
dsarno, line 38: I think this test can pass vacuously. Headless, 120 frames is well under a second; the harness only notices the receipt on its 50 ms poll and then `inspect_receipt` sleeps 0.5 s. So the mtime-stability window can straddle or fall entirely after the observer's `quit()`, and once the process is gone the mtime trivially can't change. With the driver fix reverted the test would still pass whenever the timing lands that way.

The harness writes `probe_done_file` (default `POST_UPDATE_TOOL_PROBE_FILE`) right after `live_probe` returns, so the observer can key its quit on that instead of a frame count:

```suggestion
    if observed >= 120 and FileAccess.file_exists("res://" + "%s"):
```
(with the done-file name formatted in from `fixture.POST_UPDATE_TOOL_PROBE_FILE`). That guarantees the whole 0.5 s window is observed while the driver is still spinning at the gate.

dsarno (commented):
Reviewed the receipt handshake change (fixture, driver, new unit + integration tests). The fix itself is sound; the notes are about the new tests' strength and two hygiene items.

tests/integration/_self_update_fixture.py
```diff
@@ -1055,15 +1061,23 @@ def write_install_update_driver(
...
-\t\tvar pre_file := FileAccess.open(PRE_ID_PATH, FileAccess.WRITE)
-\t\tif pre_file != null:
+\t\tif _pre_instance_id.is_empty():
+\t\t\tvar pre := DriverSupport.fetch_status(HTTP_PORT)
+\t\t\tvar pre_id := str(pre.get("instance_id", ""))
+\t\t\tif pre_id.is_empty():
+\t\t\t\treturn
+\t\t\tvar pre_file := FileAccess.open(PRE_ID_PATH, FileAccess.WRITE)
+\t\t\tif pre_file == null:
+\t\t\t\t_fail(17, "pre-instance receipt could not be opened: "
```
dsarno: Exit code 17 is already taken: line ~1001 uses `_fail(17, "update unexpectedly restarted the editor")`. Both new receipt failures reuse it, so a receipt open/write failure and a restart regression are indistinguishable by exit code (`expected_exit_code`, log triage). Codes 30+ are unused.

tests/unit/test_self_update_editor_harness.py
```diff
@@ -175,3 +175,125 @@ async def __aexit__(self, *_args):
...
+    with pytest.raises(OSError, match="device error"):
+        await agent._poll()
+    assert entered == []
+
+
+@pytest.mark.asyncio
+@pytest.mark.skipif(fixture.sys.platform != "win32", reason="Windows sharing semantics")
+async def test_native_receipt_sharing_violation_retries_until_handle_released(
+    monkeypatch, receipt_agent,
+):
```
dsarno, line 273: `win32con`/`win32file` come from pywin32, which isn't in `pyproject.toml` or `uv.lock`. It passed on the Windows CI row, so something on that runner image provides it, but a clean `uv sync` on a Windows dev box (or a runner image change) will make this test error with `ModuleNotFoundError` rather than skip.

The stdlib already has what's needed: `_winapi.CreateFile(str(receipt), _winapi.GENERIC_READ | _winapi.GENERIC_WRITE, 0, 0, _winapi.OPEN_EXISTING, 0, 0)` opens with share mode 0 (it's what `multiprocessing` uses), and `_winapi.CloseHandle(handle)` releases it. Otherwise declare pywin32 as a `sys_platform == 'win32'` dev dependency.

tests/integration/_self_update_fixture.py
```diff
@@ -625,15 +625,21 @@ async def _poll(self) -> None:
...
+            except (FileNotFoundError, PermissionError) as exc:
+                expected_nonce = ""
+                last_receipt_error = f"{type(exc).__name__} (errno={exc.errno})"
             if capability is not None and expected_nonce == capability.instance_nonce:
                 break
             if self._stop.is_set() or time.monotonic() > deadline:
-                raise AssertionError("server A never published capabilities")
+                detail = (f"; last pre-instance receipt read: {last_receipt_error}"
+                          if last_receipt_error else "")
+                raise AssertionError("server A never published matching capabilities" + detail)
```
dsarno, line 647: Small diagnostic gap: if the capability record *is* present but its nonce never matches the receipt, the failure still reads "never published matching capabilities" with no hint that a record was seen. That case is now more likely to matter because the driver pins the first receipt (previously a server identity change during the gate wait converged as the driver rewrote the file; now it times out here after 120 s). Recording a `capability seen, nonce mismatched` flag beside `last_receipt_error` would make that failure self-explanatory without exposing the nonce.

tests/integration/test_self_update_receipt.py
dsarno: Fixed in 8e17daf1218d0b47a30417ab895964f9dde11ce9. The observer now waits for POST_UPDATE_TOOL_PROBE_FILE as well as the minimum frame count. The harness writes that file only after the 0.5-second receipt probe returns. I also ran a discriminating native control that removed the generated driver's pin-once guard: the permanent test failed at the mtime assertion while the editor was alive and before either completion file existed. The unchanged driver passes.

tests/integration/_self_update_fixture.py
dsarno: Fixed in 8e17daf1218d0b47a30417ab895964f9dde11ce9. Receipt open and write failures now use distinct exit codes 30 and 31. Restart detection retains 17; the generated install driver has no collisions among these codes.

tests/unit/test_self_update_editor_harness.py
dsarno: Fixed in 8e17daf1218d0b47a30417ab895964f9dde11ce9. Replaced pywin32 with stdlib _winapi.CreateFile using share mode zero and _winapi.CloseHandle. The test clears the handle after releasing it, so final cleanup cannot close the same native handle twice. The actual Windows sharing-violation/retry test passes.

tests/integration/_self_update_fixture.py
dsarno: Fixed in 8e17daf1218d0b47a30417ab895964f9dde11ce9. Added a latched diagnostic when a capability record and a nonempty receipt are observed with different nonces. Timeout reports that mismatch without printing either nonce. Tests cover present-but-mismatched and absent records, preserve the original deadline, and check that neither nonce leaks.

dsarno:
Addressed all four review threads in 8e17daf1218d0b47a30417ab895964f9dde11ce9. The receipt observer now remains alive throughout the stability check; Windows coverage uses a stdlib exclusive handle; receipt failures have separate exit codes; timeout diagnostics distinguish nonce mismatch without exposing secrets.

Validation passed:
- CI-scope Ruff.
- Full Python suite: 2,684 passed, 116 skipped.
- Focused receipt harness and actual Windows/editor checks: 19 passed.
- Full live GDScript suite: 2,332 passed, 31 skipped; inspected final viewport, normal editor exit.
- Native updater matrix: 13 passed, 11 platform-specific skips, including same-editor private-HTTPS update, applicable v3 migrations, refusal and rollback.

An independent regression probe restored repeated receipt writes only in a disposable generated driver. The permanent stability test failed at its mtime assertion while that editor was alive, then the evidence harness completed normal cleanup. This rules out the original vacuous pass.

These are local synthetic signed-update fixtures, not qualification of an untouched published release. The combined stack supertest is still pending. Each thread has an individual reply; none was silently resolved. Leaving this PR unmerged for maintainer review.

dsarno:
Added cc091f7c so the native receipt regression runs in the existing Linux updater CI step. The ordinary Python jobs do not set GODOT_BIN and otherwise skip this test.

Verification passed before this CI-only commit: full CI-scope Ruff; full Python suite (2,684 passed, 116 skipped); cold live Godot suite (2,332 passed, 31 skipped); and the native receipt test (1 passed). The owned editor exited normally. Hosted CI is running on the new head.

dsarno:
Addressed the [outside-diff deadline review](https://github.com/hi-godot/godot-ai/pull/1073#pullrequestreview-5252838727) in 9b659263.

Moved the existing stop/deadline guard before capability discovery and receipt reads, with equality treated as expired. Checks after discovery and after the read prevent another I/O step or client startup when that operation consumed the remaining allowance. Timeout text still includes the last sanitized receipt error and any genuine nonce mismatch; a matching receipt returned late is not mislabeled as a mismatch.

Regressions verify exact discovery/read counts at 120 and 122 seconds, retained PermissionError details, a stopped wait doing no I/O, and slow discovery/read returning a matching nonce without entering the client. All six cases fail against the previous loop; the full focused harness suite passes. The pin-once writer is unchanged.

This checks the deadline around synchronous filesystem calls; it does not claim to forcibly interrupt a filesystem operation already in progress. Validation passed: full Ruff; Python 2,758 passed/120 skipped; fresh live Godot 2,341 passed/0 failed/31 skipped with normal editor exit; native updater and receipt observer 14 passed/11 skipped.

dsarno:
Independent shipping review: **PASS+NOTES for the tested patch; not yet landable**.

Head `9b65926384f7243efe576e268302db93134788db`; tested/review base `3eee4a4936b840d227ab45aa3be797a7bb03ea72`; stable base-to-head patch-id `81aa4cd4a552982c447a2574f360c4ed04642352`.

Published correction matches the independently reviewed frozen files. The deadline guard now precedes discovery/read, equality expires the wait, and checks after either operation prevent late client startup. Genuine nonce mismatch and sanitized receipt-error diagnostics remain distinct. The writer still pins once; the observer waits for the harness completion marker, so its stability assertion runs while the editor is alive. The six old-loop regression failures and 23 focused passes are retained.

Reviewed completed owner receipts: full Ruff, Python 2,758 passed / 120 skipped, cold live GDScript 2,341 passed / zero failures / 31 skipped, and native updater plus receipt observer 14 passed / 11 skipped. No test overlay was needed. The owned editor exited normally and no owned native processes remain. This is an independent source/evidence refresh, not a claim that I reran those gates.

Limits: synchronous filesystem calls already in progress are not forcibly cancelled. The native update uses local signing/TLS fixtures, not pristine published releases. Three recorded Python warnings (including the previously observed HTTPResponse finalizer warning), negative-test diagnostics and explicit backend cleanup remain disclosed.

At this refresh GitHub reports a merge conflict; new-head CI jobs are not yet listed and CodeRabbit is pending. Resolve against current trunk, verify the resulting patch identity/integration and require successful new-head checks before merging. No merge performed by this reviewer.

dsarno:
Refreshed independent shipping review: **PASS+NOTES**, subject to successful hosted checks at this new head.

- Head: `387f31e6fc30f32fbfc334217088099bfc8fb57e`
- Base: `f85d2481bf6f6cef1928f20d0667fc64d454956f`
- Stable base-to-head patch-id: `d30d57dbaf3dafca923eb91d38ed958fd81a8319`
- Published tree: `f24fe43d53ff2998c684c3e3deb65027a1443066`, exactly the independently reviewed and freshly tested integration tree.

The prior conflict is resolved: CI retains both receipt and handoff native modules and the existing all-OS client mutation queue step. The three receipt/harness source files remain unchanged from the reviewed deadline correction. No production policy change was introduced.

Reviewed the new qualification receipts, not substituted older results: Ruff and diff checks; Python 2,758 passed / 122 skipped / two deprecation warnings; native updater, receipt, client queue and handoff modules 16 passed / 11 skipped; cold live GDScript 2,352 passed / zero failures / 31 skipped. The isolated editor exited normally, frozen identity remained unchanged and no owned native processes remain. The verification owner ran these gates; this independent reviewer checked their source identity and evidence.

Limits remain explicit: deadline guards do not forcibly interrupt synchronous filesystem work already running; local signed/TLS updater fixtures do not prove pristine published-release upgrades. Expected negative-test and resource-teardown diagnostics are retained, and backend termination was deliberate fixture cleanup after editor exit.

GitHub now reports MERGEABLE. Hosted CI and CodeRabbit are still pending; no new inline/review findings appeared in this refresh. This supersedes the earlier conflict-blocked identity, but successful new-head checks are still required before landing. No merge performed by this reviewer.

dsarno:
Reviewed the new outside-diff finding in review 5262895969 against the fixture and the installed transport implementation. I am keeping this nonblocking.

The 120-second deadline bounds receipt/capability discovery. The accepted matching receipt ends that phase; client initialization has its own 30-second timeout. StdioTransport construction only assigns fields and creates synchronization objects. It does not perform I/O, await, or launch the subprocess.

The new post-read guard rejects a receipt that arrives after the discovery deadline, including slow discovery/read cases covered by the regression tests. An additional check immediately before Client entry would not establish an atomic stop/start guarantee: the thread can still be stopped or descheduled after that check or during Client entry. A strict cancellation contract would require coordinated task ownership, which this proposed check does not provide. We do not claim that the current guards provide that stronger guarantee.

Independent review reached the same conclusion. The tested published tree remains unchanged at 387f31e6; all 23 current-head checks now pass.

PR #1074, opened 2026-09-18
fix(clients): queue Dock mutations behind one worker

### LINKED ISSUE #1047: Configure all races the global client mutation lock and reports misleading recovery errors
## Problem
Configure all can race its own client-configuration actions against the global mutation lock. Ordinary contention is reported as a safety incident: affected rows instruct the user to stop clients or reboot and explicitly delete the lock. The repeated long recovery text also overwhelms the client list. By the time the user investigates, the successful action may already have released the lock.

## Reproduction
Confirmed with Godot 4.7.2 and the unchanged production Config.configure path:
1. Register two disposable JSON client descriptors in an isolated editor environment.
2. Preserve a large unrelated user field in the first file so its real read/write/verify transaction overlaps another request.
3. Start the first configure and observe its global mutation lock.
4. Configure the second client while the first transaction is active.

The first configure succeeds. The second returns the stop-clients/reboot/delete-lock recovery message. The first owner releases its own lock. Retrying the second configure succeeds without reboot, manual lock removal, or any security bypass. Both configuration operations use the actual production strategy and verification path; no lock behavior is mocked.

The user also observed repeated errors after Configure all. That specific occurrence did not capture the transient owner record, so the controlled reproduction establishes the mechanism rather than proving the screenshot's exact timeline.

## Cause and scope
- mcp_dock.gd::_on_configure_all_clients dispatches every eligible client in a loop.
- client_job_owner.gd::_start_action guards only duplicate requests for the same client and starts independent threads for different clients.
- client_configurator.gd::configure acquires one global durable mutation lock across the write and read-back verification.
- client_mutation_lock.gd::acquire returns the same recovery message for an active competing transaction and an unsafe abandoned lock.
- Failed action results remain errors in their client rows after the winning transaction releases the lock.

This behavior already exists in the v4 architecture commit 1a0efe8f. It is not introduced by the PR #1042 endpoint picker or port override changes.

## Proposed bounded follow-up
Serialize automatic client mutation requests through the existing client job owner, preserving the durable cross-editor/process lock and all termination checks. Queue Configure all requests instead of launching conflicting writes concurrently. Do not delete the lock automatically or weaken refusal when a previous owner's termination is unproven.

Distinguish known active work from recovery requiring human action where ownership is established, and avoid repeating oversized recovery instructions on every client row.

## Acceptance checks
- Real Configure all requests complete sequentially without false safety-lock errors.
- Unknown or stranded locks still prevent writes.
- Ambiguous child termination prevents later queued mutations.
- Cancellation and reload handle queued requests without reporting false success.
- Ordinary active work does not instruct the user to delete a safety lock.

## Related client-configuration wording cleanup

The port-conflict guide button tooltip in `mcp_dock.gd` still names only the legacy `godot_ai/http_port` setting. Correct it to match the effective two-port picker and updated guide. This pre-existing hover text does not drive the button action or endpoint selection. CodeRabbit identified it outside the PR #1042 diff; it is tracked here as non-blocking client-configuration UX cleanup.

dsarno:
## Why

Configure all started one mutation worker per client, but those workers shared one durable lock. The first client could succeed while the others reported a recovery error. Queue Dock actions in `client_job_owner.gd` and dispatch one worker at a time.

Closes #1047.

## Scope

- Show queued configure and remove actions in the Dock.
- Capture the launch context when an action starts. Cancel pending actions when the owner stops.
- Block all subsequent mutations if process termination or lock release is unproven.
- Point the port-guide tooltip to Godot AI settings.

## Tradeoffs

Competing MCP mutations return busy immediately. They do not join the UI queue, which could outlive the caller's timeout. An admitted MCP action keeps its existing timeout budget.

## Blast Radius

This changes Dock and MCP mutation admission. The durable lock, worker cancellation, and thread join still determine when another write is safe.

## Verification

- Ruff and the full Python suite passed. Python verification used the independently reviewed test-harness fix in #1073 as a temporary overlay; that fix is not included here.
- The live Godot suite passed with zero failures.
- A native integration invoked the actual Configure all Dock callback and read back both client files, including unrelated values. It checked immediate MCP owner refusal, successful removal, and durable lock refusal.
- Reload smoke and concurrent reload stress passed with no unexpected errors.
- The signed local self-update smoke passed from 4.1.0 to 4.1.1 in the same editor, with client migration and the replacement server verified.

## Summary by CodeRabbit

* **New Features**
  * Client configuration actions now run in order, with additional requests shown as “Queued…”.
  * Queued Configure and Remove actions can now be cancelled directly from the interface.
  * Port-conflict guidance now explains how to change ports in Godot AI settings.

* **Bug Fixes**
  * Prevented competing MCP changes from being deferred behind queued client actions.
  * Improved safeguards after unsuccessful mutation termination or lock-release operations.

* **Tests**
  * Added coverage for ordering, cancellation, safety locks, and multi-client workflows.


plugin/addons/godot_ai/utils/client_job_owner.gd
```diff
@@ -464,16 +474,55 @@ func _start_action(client_id: String, action: String, request_id: String) -> Dic
...
+		_pending_actions.append({"client_id": client_id, "action": action})
+		_start_next_action()
+		_publish_snapshot()
+		return {"ok": true}
+	if not _action_threads.is_empty() or not _pending_actions.is_empty():
+		return {"ok": false, "error": "Another client action is running or queued. Retry when it finishes."}
+	return _begin_action(client_id, action, request_id)
+
+
+func _start_next_action() -> void:
```
dsarno, line 500: Design note rather than a bug: once queued, an action can only leave `_pending_actions` by starting or by `quiesce()`, and the dock disables both buttons for queued rows. So `Configure all` with N clients commits the user to up to N × (75 s + prewarm) of serialized work with no way to drop an entry, and the only escape is a plugin reload. Before this PR the same click at least failed fast.

A cancel for queued entries is cheap and safe because nothing has started: remove the entry from `_pending_actions`, emit the same `cancelled before it started` result, publish. Could hang off the existing Remove/Configure button for a `queued` row instead of disabling it.

plugin/addons/godot_ai/utils/client_job_owner.gd
```diff
@@ -166,6 +175,7 @@ func _process(_delta: float) -> void:
 func quiesce(deadline_msec: int = 0) -> Dictionary:
 	_accepting_work = false
 	_refresh_state = RefreshState.SHUTTING_DOWN
+	_discard_pending_actions("Client action cancelled before it started.")
```
dsarno: Small asymmetry: this emits `action_completed` for every pending entry while `_refresh_state` is already `SHUTTING_DOWN`, but `_finalize_action` deliberately suppresses `action_completed` in that state. During a plugin reload or self-update mid-queue the dock (possibly mid-teardown) gets N red 'cancelled before it started' rows while the running action's result is dropped silently. Either both paths should report or neither; `test_quiesce_discards_queued_actions_without_running_them_after_resume` currently enshrines the asymmetric version, so worth deciding on purpose.

dsarno (commented):
Reviewed the queue against the owner's lifecycle (begin/poll/finalize/timeout/quiesce), the dock plumbing, and both new test layers. The serialization fix is correct: the dock sets rows in-flight before emitting, so a synchronously discarded action still resets its row, and lock release happens in the worker before the join so `_start_next_action`'s `is_locked()` check can't see the previous worker's own claim. Remaining notes are design-level.

tests/integration/test_client_mutation_queue.py
```diff
@@ -0,0 +1,168 @@
...
+    dock.queue_free()
+    owner.queue_free()
+    var file := FileAccess.open("res://result.json", FileAccess.WRITE)
+    file.store_string(JSON.stringify({"checks": checks, "failures": failures, "results": results}))
+    file.close()
+    get_tree().quit(0 if failures.is_empty() else 1)
+'''
+
+
+@pytest.mark.skipif(sys.platform != "win32", reason="isolated Windows client configuration")
```
dsarno: Nothing in the queue behaviour is Windows-specific, but this is the only end-to-end test of Configure-all serialization and the MCP busy path, and it never runs on the Linux/macOS rows. The two Windows-only bits are the `path_template = {"windows": ...}` key and the `OS.get_config_dir() != APPDATA` isolation check. Keying the template by `OS.get_name().to_lower()` and checking against `XDG_CONFIG_HOME` / `HOME` on the other platforms would let it run everywhere.

plugin/addons/godot_ai/utils/client_job_owner.gd
dsarno: Addressed in f5198d99. Queued rows now expose an enabled **Cancel queued** button for the pending operation. Its dedicated signal goes through the plugin to `cancel_pending_action()`, which removes only that queued entry and emits a neutral `cancelled` result. The Dock restores its cached status without showing a repair error; running workers, MCP admission and the durable lock are unchanged. Repeating Configure all remains a configure request rather than cancelling queued rows.

The owner tests cover queued Configure and Remove, survivor ordering, and running/stale/unknown cancellation no-ops. The native regression presses the actual Dock button through the production plugin callback, checks stale-view reconciliation, and proves that the cancelled client's file is unchanged before an intentional requeue. The rendered Clients window was also inspected before and after cancellation; the fixture passed all 41 assertions and exited normally.

plugin/addons/godot_ai/utils/client_job_owner.gd
dsarno: Addressed in f5198d99 by deliberately choosing silent teardown for both pending and running actions. `quiesce()` clears the pending queue and publishes the empty work snapshot without emitting `action_completed`, matching `_finalize_action()` while shutting down. Explicit user cancellation still reports a neutral cancelled result, and unsafe-owner queue discard errors remain intact.

The quiesce regression now requires zero completion callbacks and confirms queued work does not start after resume. The full live Godot suite passed (2,338 passed, 31 skipped, zero failures), and a separate production reload smoke verified scene continuity, a fresh session and working editor/game-log handlers, followed by a normal editor exit.

tests/integration/test_client_mutation_queue.py
dsarno: Addressed in f5198d99. The fixture now supplies both `windows` and `unix` path templates and validates the exact isolated configuration directory: APPDATA on Windows, XDG_CONFIG_HOME on Linux, and HOME/Library/Application Support on macOS. The registry's path resolver uses `unix` for the latter two, so lowercasing `OS.get_name()` would not select the correct template.

Removing the platform skip alone would still leave this test out of the existing CI module lists. I added an explicit headless `GODOT_BIN=godot pytest -v -s tests/integration/test_client_mutation_queue.py` step to every Godot OS row before server startup, with a three-minute timeout. The Windows native regression and full local Python suite passed (2,743 passed, 58 skipped, zero failures); Linux and macOS execution is left to the newly wired CI rows rather than claimed from local testing.

dsarno:
Queued Configure and Remove actions now expose a Cancel queued button. Cancelling removes only that pending action, restores the row's cached status without a repair error, and leaves the running worker and durable safety lock unchanged. Repeating Configure all still queues configuration rather than cancelling rows. Quiesce silently discards pending actions, matching the existing suppression of running completions.

The native regression now exercises the actual Dock button → plugin callback → owner path, checks that cancellation causes no file write before intentional requeue, and verifies stale UI refresh. Its isolated configuration paths support Windows, Linux and macOS; CI invokes it on every Godot OS row.

Validation: full CI-scope Ruff; full Python suite including native editor rows (2743 passed,58 skipped,0 failed); live GDS2338 passed,31 skipped,0 failed; actual rendered queued/cancelled UI,41 native assertions; production reload smoke and normal editor exits. Native updater verification uses the separately reviewed receipt-harness prerequisite from PR1073, excluded from this change.

dsarno:
Independent shipping refresh: **PASS + NOTES** for `f5198d994610d84bed354ec044347ad92ffac855`, base `24859943117a9390ec9f924ea7da3adf9fd39866`, stable patch ID `252c1cc5319bf124a9288eee75c370c85a7e6de3`. I did not implement this PR's production changes. The seven published files match the recorded native-verified candidate, accounting only for line endings.

The queue admits only UI actions; MCP writes still fail promptly when busy. Cancellation removes only a pending entry, routes through the production plugin callback, and restores the row neutrally. Running worker ownership and durable-lock safeguards remain intact. Quiesce deliberately suppresses both pending and running completion notifications. The portable fixture verifies its isolated OS configuration directory and is explicitly wired into every Godot OS CI row.

Retained evidence includes the failing baseline, actual queued-button cancellation with unchanged cancelled-client file, stale-view reconciliation, survivor ordering and quiesce checks, rendered Clients-window inspection, reload continuity, 2,338 live GDScript passes, and 2,743 Python passes. All 23 current check/status results succeed; no new unanswered review finding was returned. The native updater portion of that Python run used the documented test-only #1073 receipt prerequisite, restored and excluded from this patch.

Ready for the coordinator's merge check. This verdict covers this exact patch; any conflict resolution against newer main still needs its own verification. No merge performed here.

Issue #1046, opened 2026-09-10
Windows: IPv6-only HTTP allowlist is unreachable by IPv4 editor lifecycle probes


dsarno:
An exclusively IPv6 `--allow-host` configuration selects an IPv6 HTTP listener, but the editor lifecycle probes the IPv4 loopback endpoint. On Windows, an IPv6-only listener cannot satisfy that probe. This predates the lifecycle/update PR stack and needs its own supported-configuration decision.

Source evidence, checked against both pre-stack commit 9db3e0d0 and PR #1039:
- `src/godot_ai/transport/origin_guard.py::bind_host_for_networks` returns `::` when every allowed network is IPv6. Its existing docstring explicitly describes Windows IPv6-only behavior.
- `plugin/addons/godot_ai/utils/server_lifecycle.gd` connects its HTTP status probes to `127.0.0.1`.
- PR #1039 adds an early IPv4 bind probe, but the endpoint mismatch existed before that change. It is not a newly introduced authentication bypass.

Expected: supported HTTP configurations remain reachable by the local editor, or unsupported combinations produce an actionable refusal.

Next steps:
- Reproduce the full authenticated editor/server flow on Windows with an IPv6-only allowlist; the current finding is source-confirmed, not a retained end-to-end reproduction.
- Choose and document explicit endpoint support or a clear unsupported-configuration error. Do not silently widen allowlists or assume cross-platform dual-stack behavior.
- Preserve bearer authorization, request origin/host checks, and loopback-only WebSocket binding.
- Add a real Windows regression for the selected behavior and retain ordinary IPv4-loopback coverage.

Tracked separately from #1039 to keep its startup/process-proof changes scoped. Review reference: https://github.com/hi-godot/godot-ai/pull/1039#discussion_r3975458117

dsarno:
Confirmed the endpoint mismatch on Windows with current main `2485994`, Godot 4.7.2 and the actual backend. With `--allow-host fd00::/8`, authenticated HTTP on `[::1]` returned 200 while the IPv4 loopback request timed out. The native Godot lifecycle probe still targeted 127.0.0.1 and reported `port_unbound`. The OS table showed HTTP on `[::]` and WebSocket correctly limited to 127.0.0.1.

[...]

Issue #1072, opened 2026-09-18
Support IPv6 endpoints consistently across editor lifecycle and attach clients


dsarno:
## Problem

An IPv6-only HTTP allowlist can select a backend listener on :: while editor lifecycle probes and attach clients still assume IPv4 loopback. On Windows, current main24859943117a9390ec9f924ea7da3adf9fd39866 reproduced authenticated status200 through [::1], an unreachable127.0.0.1 endpoint, and a native editor probe reporting port_unbound. See #1046 for the bounded reproduction.

The maintainer decision for the current bug pass is to reject the unsupported configuration clearly. This issue tracks a future end-to-end IPv6 implementation separately; it is not authorization to widen the editor WebSocket bind or add a new remote credential mode.

## Scope for a future proposal

- Define an explicit endpoint/address-family contract shared by CLI, managed and dev server startup, authenticated lifecycle/status/shutdown probes, capability discovery, and attach clients.
- Handle bracketed IPv6 URLs, loopback identity and actual listener/adoption checks consistently. Do not infer that an IPv6 wildcard accepts IPv4 on every supported OS.
- Define IPv6-only and mixed IPv4/IPv6 allowlist behavior without broadening the requested network admission policy.
- Keep the editor WebSocket loopback-only and preserve per-instance HTTP capability authentication and process-ownership checks.
- Qualify native Windows behavior as well as supported POSIX platforms; exercise editor startup, attach, reload, self-update, backend restart, malformed inputs, unavailable address families, and negative cross-instance/adoption cases.

Acceptance requires an IPv6-only backend to be discoverable and usable by the local editor and attach path throughout its lifecycle, with no ambiguous mutation replay or silent fallback that widens access. Until that contract is implemented and verified, retain the explicit rejection added under #1046.

Related: #1046, #1054 (separate wildcard occupancy issue), #1009 (remote access backlog).

dsarno:
Continuing this deferred proposal in https://github.com/hi-godot/godot-ai/discussions/1102. No implementation is planned at present, so closing this issue for now at the maintainer's request. The original context and comment history remain here, and the discussion links back to them.

PR #1075, opened 2026-09-18
fix(windows): reject unsupported IPv6-only remote access

dsarno:
Windows IPv6-only HTTP allowlists start a backend that the editor and attach bridge cannot reach through IPv4 loopback. Reject that configuration before probing or launching, show an actionable dock/settings error, and emit a terminal startup trace. CLI failures also reach the JSON startup report.

Existing IPv4 and mixed allowlists retain their bind behavior, POSIX IPv6 selection is unchanged, and the editor WebSocket remains loopback-only. Full IPv6 support is tracked separately in #1072.

Depends on #1073; base branch: `fix/1069-updater-receipt`. Its verified Windows test-fixture correction is inherited so native update tests can run reliably.

Validation:
- Ruff and diff whitespace checks pass.
- Full Python suite: 2,759 passed, 58 skipped (three dependency/finalizer warnings).
- Live Godot: 2,335 passed, 0 failed, 32 skipped; 74/74 suites.
- Actual CLI IPv6-only/repeated/mapped IPv6 rejection and supported authenticated IPv4/mixed startup pass.
- Rendered editor with saved IPv6-only settings shows the corrected heading and actionable error before spawning; settings show the inline error with Apply disabled.
- Three isolated reload cycles reconnect successfully; this is not concurrent load coverage. Client SSE teardown emitted ClosedResourceError, with all reload assertions passing.

Closes #1046.


plugin/addons/godot_ai/utils/server_lifecycle.gd
```diff
@@ -103,6 +103,11 @@ func start_server() -> void:
 
 func _begin_start_episode(existing_id := 0, probe := true) -> void:
 	_cancel_effect()
+	var config_error := McpAllowHosts.configuration_error(str(_plan.get("allow_hosts", "")))
+	if not config_error.is_empty():
+		_block_without_effect("unsupported_remote_access", config_error)
```
dsarno, line 110: This early return happens before the two resets that follow (`_replacement_authorization = null`, `_transport = null`). A blocked start therefore keeps the previous episode's transport, which `get_status_dict()` still publishes (line ~1682) and `_probe_current_transport` still probes, plus any pending replacement authorization that the next start could consume. It only bites when the same lifecycle instance is restarted after the setting changed, but the fix is just moving the guard below those two lines.

plugin/addons/godot_ai/mcp_dock.gd
```diff
@@ -827,6 +827,9 @@ func _update_status() -> void:
 	elif bool(server_status.get("handoff_retry_pending", false)):
 		status_text = "Recovering after update…"
 		status_color = COLOR_AMBER
+	elif str(server_status.get("episode_reason", "")) == "unsupported_remote_access":
```
dsarno: `unsupported_remote_access` isn't in `_dock_state`'s `match reason`, so the state is `CRASHED`, and this arm patches only the heading by matching the reason string. The comment a few lines up prescribes the other route: one `ServerStateScript` constant, one arm here, one body in `_crash_body_for_state`. With the string special case, everything else keyed on `CRASHED` (`blocks_client_health`, crash-panel buttons/restart affordances) still treats a configuration error as a crash.

#1071 adds `port_occupancy_unknown` with exactly the same gap, so a `ServerState.UNSUPPORTED_CONFIG`-style constant (or a generic `BLOCKED_CONFIG`) with a `_dock_state` arm would serve both PRs and drop the `episode_reason` plumbing from `get_status_dict`.

tests/unit/test_cli_reload.py
```diff
@@ -486,3 +486,22 @@ def test_run_with_reload_rejects_non_http_transport():
...
         )
+
+
+@pytest.mark.parametrize("transport", ["sse", "streamable-http"])
+@pytest.mark.parametrize("reload", [False, True])
+def test_windows_ipv6_rejection_reaches_startup_report(monkeypatch, transport, reload):
+    from godot_ai import runtime_info
+
+    reported = []
+    monkeypatch.setattr("godot_ai.transport.origin_guard.sys.platform", "win32")
```
dsarno: `origin_guard.sys` *is* the `sys` module, so this sets `sys.platform = "win32"` for the whole process while `godot_ai.main()` runs through `telemetry`, `orphan_reaper` and `transport/capability`, which all branch on it too. It passes today, but on the Linux/macOS rows `main()` is executing Windows-only branches in unrelated modules right up to the `ValueError`, and any new Windows-only import or API in those modules will break this test for reasons that have nothing to do with the allowlist. Same applies to the `test_origin_guard.py` cases.

The GDScript side already does this the clean way (`configuration_error(raw, platform := OS.get_name())`). Mirroring it, `bind_host_for_networks(networks, platform: str = sys.platform)`, lets the tests pass `platform="win32"` without touching global state.

dsarno (commented):
Reviewed the IPv6-only rejection across the lifecycle guard, dock rendering, allow_hosts helper, Python bind check and both test layers. The rejection logic is consistent between GDScript and Python (`::ffff:` mapped addresses classify the same way on both sides) and the Python check runs inside the startup-report try block before the pid file. Three items inline; the dock one overlaps with a comment I left on #1071.

plugin/addons/godot_ai/utils/server_lifecycle.gd
dsarno: Addressed in ba48692e34c979675267524261913ce09aff460d. Moved both resets immediately after cancellation of the old effect and before the configuration guard. The regression test first establishes transport and replacement authority, then switches the same lifecycle instance to an unsupported allowlist. Refusal clears both authority fields, blocks connection, leaves no current transport to probe, and emits no new effect.

plugin/addons/godot_ai/mcp_dock.gd
dsarno: Addressed in ba48692e34c979675267524261913ce09aff460d. Added an explicit UNSUPPORTED_CONFIG state with a stable appended value, terminal diagnosis, blocked client-health interpretation, and state-based heading/body. The actual dock shows the configuration remedy and Reload Plugin, without a port picker or managed-server restart button. I kept port_occupancy_unknown separate: #1071 now handles that transient discovery failure as FOREIGN_PORT with its own message. It is not an unsupported configuration. I also retained episode_reason because the reviewed #1071 diagnostics use it. Startup and an actual Reload Plugin action both verified the new state with empty transport/replacement authority and no backend launch.

tests/unit/test_cli_reload.py
dsarno: Addressed in ba48692e34c979675267524261913ce09aff460d. Added a keyword-only platform parameter defaulting to sys.platform. Direct binding tests pass their platform explicitly. The CLI tests wrap only origin_guard.bind_host_for_networks with functools.partial(platform="win32"), so they exercise the real rejection/reporting path without changing sys.platform for unrelated modules.

dsarno:
Addressed all three review threads in ba48692e34c979675267524261913ce09aff460d, including the verified #1073 prerequisite. Unsupported Windows IPv6-only starts now clear prior authority before refusing, and the dock presents a configuration diagnosis with appropriate controls. Tests no longer patch global sys.platform.

Validation passed:
- Complete CI-scope Ruff.
- Core Python: 2,700 passed, 44 skipped.
- Native editor selection: 61 passed, 14 skipped.
- Cold live GDScript: 2,337 passed, 31 skipped.
- Focused Python: 159 passed.
- Actual rendered startup and Reload Plugin refusal, with both screenshots inspected and normal editor exits. No backend launched; transport and replacement authority remained empty.

Each thread has an individual reply. The proposed shared state for unknown port occupancy was not adopted because transient discovery failure and unsupported configuration need different remedies. Full IPv6 endpoint support remains separate in #1072.

Native updater results are local fixture coverage, not untouched published-release qualification. The combined supertest is pending. Leaving this PR unmerged for maintainer review.

dsarno:
Independent shipping review: **PASS+NOTES** for head `31d8dd366acc4d74d5d4623d36fbaaef612ad460`, base `6652b8e87271ddf5887073fff608b85509500444`, stable patch ID `05e978b9e42260f17e64bd1dd28cc0974f49d13d` (tree `9b662bb86bfe56760de76adc4839ffa71d1deb41`).

I did not author this patch. I reviewed the original change, all three inline review threads, the complete integrated 12-file delta, and the fresh qualification logs/receipts. The published tree matches the frozen tested tree. This is independent source and execution-evidence review; the verification owner ran the fresh native fixtures.

The integration retains the distinct Windows port-discovery diagnosis from #1071 and connection-loss recovery from #1082. The corrected #1073 fixture, unit tests and receipt CI row match main exactly. Unsupported configuration clears prior transport/replacement authority before refusal, has its own dock state and remedies, and does not broaden the loopback WebSocket boundary. Full IPv6 support remains #1072.

Verified evidence:
- Ruff passes; full Python: 2,777 passed, 122 skipped. Explicit native updater/receipt: 14 passed, 11 skipped, without an overlay.
- Cold live GDScript: 2,363 passed, 0 failed, 31 skipped.
- Seven real CLI controls: four IPv6-only rejection variants (including --reload) produce actionable startup reports; empty, IPv4 and mixed allowlists return authenticated HTTP 200, anonymous 401, with a loopback WebSocket listener.
- Actual saved-setting startup and Reload Plugin signal activation retain state 12, empty authority and no backend launch. I inspected the final rendered retry screenshot: correct heading/remedy and Reload Plugin, without misleading port/restart controls.
- Three isolated reloads reconnect. A separate socket-only loss with the backend alive shows amber recovery followed by green. This is not concurrent-load coverage.
- Both native editors exited normally; source hashes and saved scene bytes remained unchanged, and owned processes were cleaned up.

Notes: Windows Compatibility native coverage; local updater fixtures do not establish untouched published-release qualification. Python retained two deprecation warnings and an unattributed HTTPResponse finalizer warning. Expected negative-test editor diagnostics and teardown diagnostics remain recorded; no clean-console claim. Current-head CI still needs to finish before landing.

PR #1078, opened 2026-09-18
test: account for post-update handoff timer assertions

dsarno:
The post-update handoff test awaited real timers inside a synchronous GDScript test runner. The runner returned a passing result before the assertions after the first await executed. A discriminating native-editor check confirmed that a failing assertion before the await failed the test, while the same assertion after it was omitted from the returned result.

Replace that test with four synchronous cases using fresh objects: scheduled capability handoff is amber; an unscheduled handoff, an unproven launch, and missing listener tools remain red. Each case verifies that presentation grants no transport authority. A separate native-editor integration driver awaits the actual retry timers and verifies consumption into a fresh probe, budget exhaustion, and Stop supersession before writing its result. Add that module to the existing Linux real-editor CI invocation.

This changes tests and CI only. Production retry behavior, the conservative red-state policy, and the synchronous test runner are unchanged.

Validation on Windows with Godot 4.7.2:

- Full CI-scope Ruff passed.
- Full pytest: 2,677 passed, 116 skipped; two warnings (Starlette httpx deprecation and an HTTPResponse finalizer warning).
- Cold live GDScript suite: 2,335 passed, 0 failed, 31 skipped across 74 suites.
- Native timer integration: 1 passed; normal editor exit 0.
- Four synchronous cases separately verified through the actual runner, with 6/5/5/5 accounted assertions.

The native timer fixture uses an inert plugin subclass with automatic process effects disabled. It exercises real engine timers and lifecycle state transitions, not a complete signed-update flow. Other asynchronous test findings are outside this PR.

## Summary by CodeRabbit

- **Tests**
  - Expanded integration coverage for post-update handoff behavior, including retry scheduling, consumption, exhaustion, and stopping.
  - Added checks for connection blocking, transport state, and lifecycle episode states during recovery scenarios.
  - Verified that recovery status is shown only for scheduled retries with confirmed capability support; other blocked or unverified cases remain clearly marked as blocked.
  - Included the new handoff integration test in Linux CI validation.


dsarno:
Independent shipping review: **PASS + NOTES** for `ecbcda3dd0091c82db2bc96a079b4e3b34a8471f`, base `24859943117a9390ec9f924ea7da3adf9fd39866`, stable patch ID `5ff27906817eaeca06cf47576e490dac2c532846`. I did not implement this patch.

The original native sentinel experiment proves the synchronous runner returned a passing result before assertions after await executed. This change keeps four presentation cases synchronous and moves actual timer behavior into a native editor driver whose result is written after the awaits. Python checks the stored scheduled/consumed/exhausted/stopped values, including fresh episode, retry consumption, unchanged stopped/exhausted state, and blocked transport authority. The existing harness requires normal editor exit; CI explicitly invokes the new test.

Retained verification: exact native timer test passed; synchronous cases recorded 6/5/5/5 assertions; cold live suite 2,335/0/31; full Python 2,677 passed/116 skipped; Ruff passed. The two Python warnings are disclosed in the retained report. All 23 current check/status results succeed, with no inline or substantive review findings returned.

This repairs test accounting and does not claim a production lifecycle fix. Native timer coverage uses the real lifecycle with automatic process effects disabled; broader lifecycle/update verification remains separate. Code verdict passes; GitHub mergeability was still UNKNOWN in this snapshot, so the coordinator must refresh that before landing. No merge performed here.

Issue #1056, opened 2026-09-13
v4 attach bridge: write tools fail with ConnectionError at ~10x the 3.x rate — steady-state, not release churn


dsarno:
## What telemetry is showing

New in the v4 line and absent in 3.x: ConnectionError on tool calls runs at 2.1-2.9 per 1k calls on every 4.x release (4.0.2: 2.15, 4.0.4: 2.21, 4.1.0: 2.79) versus 0.26 per 1k on 3.2.5 — roughly 10x. It concentrates almost entirely on FILE-WRITING tools; in a 3-day window (2026-09-08..11): script_create 591 events / 269 installs, filesystem_manage(write_text) 548 / 194, script_patch 151 / 58, batch_execute 75 / 48, test_run 59 / 31; reads barely feature. Failures are instant (p50 0.04s), and the caller cannot tell whether the write landed. Two hypotheses were tested and rejected: (a) release churn — five releases in five days cycling backends under live clients — but the hourly rate is flat at 2.5-4.5 per 1k in every busy hour with no spikes at release times; (b) reload timing — 4.1.0's reload stabilization (#1039) moved neither EDITOR_NOT_READY/PLUGIN_DISCONNECTED nor ConnectionError. The fingerprint that remains: a write triggers an editor-side GDScript reload or filesystem rescan, the plugin/bridge cycles, and the attach bridge drops the in-flight connection. #823 (response-stream resumption across same-instance disconnects) is the architectural fix for this class; #1006 (decide GDScript reload before registering the write) touched the same path in 4.0.3 without moving the rate; #1034 (cannot connect at all) is the total-failure cousin, not this. Asks: (1) resume or transparently retry the in-flight request across a same-instance bridge disconnect (#823), or at minimum report in the error payload whether the write was applied so callers can avoid double-writes; (2) keep the bridge connection alive across a plugin reload that the write itself triggered; (3) carry a reason/sub-code on ConnectionError so telemetry can decompose it the way EDITOR_NOT_READY was after #667.

## Scope

- **Window**: last 30 days
- **Affected installs (7d)**: 819
- **Affected installs (14d)**: 821
- **Total matching events (lifetime in 90d retention)**: 3151
[...]

PR #1080, opened 2026-09-20
fix(plugin): set the WebSocket inbound buffer before connecting so >8 KiB commands stop closing the socket with 1009

dsarno:
Commands larger than 8 KiB could close the editor WebSocket with code 1009 before reaching a handler. Godot captures the receive limit during the WebSocket handshake, so increasing `inbound_buffer_size` after authenticating the server did not change the live connection's limit.

Configure the 4 MiB peer buffers before connecting, including on reconnect. The application still rejects handshake messages larger than 8 KiB after receipt. This is an application limit, not a pre-authentication allocation bound: the larger peer buffers are already allocated. The documentation now states that distinction without claiming total memory is bounded by one frame.

Regression coverage checks both initial/reconnect configuration and valid handshake acknowledgements at exactly 8,192 and 8,193 bytes.

Validation on Windows with Godot 4.7.2:

- Baseline and candidate both accepted an 8,034-byte command. Baseline rejected 8,433-, 16,518- and 20,140-byte commands with code 1009 and created no files; candidate accepted them with exact readback, including after a deliberate WebSocket reconnect.
- Live GDScript: 2,339 passed, zero failed, 31 skipped across 75 suites, including five fixture helper tests. Three isolated reloads passed. Final dock was connected; owned editors exited normally.
- Full Ruff passed. Full Python rerun: 2,742 passed, 58 skipped. The first run had 2,733 passed, 58 skipped and one existing Windows receipt-file sharing failure; the successful rerun used the authorized test-only #1073 harness overlay. That overlay was removed before commit and is not included in this PR. The production/GDScript files match the verified source hashes.

Addresses the confirmed large-command subset of #1056. It does not establish the cause of other disconnects or make ambiguous dispatched writes safe to replay. The misleading dock label observed during deliberate reconnect is tracked separately in #1081.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed WebSocket connections closing with “Message too big” errors when sending command frames larger than the handshake limit.
  * Larger command messages are now accepted immediately after connecting, while oversized handshake messages remain rejected.

* **Documentation**
  * Clarified WebSocket message-size limits and buffer behavior during authentication and normal command processing.

* **Tests**
  * Added regression coverage to verify connection buffers are configured correctly before initial connections and reconnections.


dsarno:
Published follow-up a1eaff57: clarified the receive-buffer memory trade-off and added a behavioral handshake boundary regression at 8,192/8,193 bytes. Removed the misleading claim that pre-authentication memory is limited to one frame. The independent source/test review passed.

Native Windows Godot 4.7.2 control: baseline fails 8,433-, 16,518- and 20,140-byte commands with code 1009 and no file creation; candidate succeeds with exact readback, including after reconnect. An 8,034-byte control succeeds on both. Full live GDScript passed 2,339/0/31 (including five fixture helper tests), with three isolated reloads and normal editor exit.

Full Ruff passed. Full Python rerun passed 2,742 tests with 58 skips. The first run exposed the known Windows receipt-file sharing failure (2,733 passed, 58 skipped, one failed); the successful rerun used the explicitly authorized #1073 test-only harness overlay. Removed that overlay before committing and checked the three intended files against their verified hashes.

This closes the confirmed large-command finding only. Other #1056 disconnect causes remain unconfirmed, and ambiguous dispatched writes still must not be replayed. The separate incorrect dock heading is tracked in #1081. Left unmerged for maintainer review; hosted checks are refreshing at the new head.

dsarno:
Independent shipping refresh: **PASS + NOTES** for `a1eaff57920a6c15a6b96fcfea5498a47727e58f`, base `89d12010300ff56bf2b02424e940f7efc99ae284`, stable patch ID `2c7a21e7a29de2f5b46a6b0f690116e8c1dfae53`. I reviewed the follow-up without writing its production changes. The published three-file snapshot matches the retained verified sources after normalizing mixed line endings.

Initial connection and reconnect both configure the peer buffer before dialing. The 8 KiB application-level handshake limit remains enforced: the valid acknowledgement test accepts 8,192 bytes and rejects 8,193. The documented trade-off is a 4 MiB peer receive budget before authentication, not an 8 KiB memory guarantee.

Retained native evidence shows the baseline closing an 8,433-byte command with 1009 and no file creation; the candidate succeeds with exact readback for that and larger commands, including after reconnect. Final live tests passed 2,339/0/31 (five fixture helpers included), plus three isolated reloads and normal editor exit. Ruff and full Python passed; the latter used the disclosed #1073 test-only receipt overlay, removed before publication. All 23 current check/status results succeed; no new review comments were returned.

This fixes the demonstrated large-command defect, not every #1056 disconnect. It does not authorize replay of ambiguous writes or claim exhaustive adversarial buffering qualification. Ready for the coordinator's exact-patch merge check; no merge performed here.

